PR #10, opened 2018-09-27
Expose the cluster's instance role.

### LINKED ISSUE #9: Expose node instance role
Hey! Love the work you are doing on pulumi. Just started experimenting with this and hit a wall when trying to work with [kube2iam](https://github.com/jtblin/kube2iam). Kube2iam requires that you allow your worker nodes to assume the roles of individual pods. This requires the use of the nodeInstanceRoleARN to create a trust relationship on the roles to be assumed. I guess one could jump through some hoops (or maybe I'm missing something) to get the ARN but I don't see why it shouldn't be exposed from the Cluster class. This probably applies to some other properties as well, resulting in increased usability of this component.

pgavlin:
Fixes #9.

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -130,7 +130,7 @@ export class Cluster extends pulumi.ComponentResource {
     /**
      * The service role used by the EKS cluster.
      */
-    public readonly instanceRole: ServiceRole;
+    public readonly instanceRole: pulumi.Output<aws.iam.Role>;
 
     /**
      * The security group for the cluster's nodes.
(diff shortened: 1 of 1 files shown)


nodejs/eks/cluster.ts
```diff
@@ -127,6 +127,11 @@ export class Cluster extends pulumi.ComponentResource {
      */
     public readonly clusterSecurityGroup: aws.ec2.SecurityGroup;
 
+    /**
+     * The service role used by the EKS cluster.
+     */
+    public readonly instanceRole: ServiceRole;
```
lukehoban: Minor - but any reason to expose `ServiceRole` instead of just raw `aws.iam.Role`?

PR #12, opened 2018-10-19
getCluster function

mazamats:


lukehoban:
@mazamats Sorry that we didn't take a look at this sooner!  I do think we should take something like this as an addition here or in `@pulumi/aws`.  cc @pgavlin.

pgavlin:
My apologies as well; this slipped through the cracks.

We definitely want the ability to compute a kubeconfig from an EKS cluster. In an ideal world, we'd make this an instance method on the `aws.eks.Cluster` resource in `@pulumi/aws`. @lukehoban @jen20 that's not possible, right? In lieu of that, we could add a function like `getKubeconfig` that does essentially this and takes a cluster ID as input.

PR #15, opened 2018-10-26
Add support for custom user data for worker nodes.

pgavlin:
These changes add a new argument, `nodeUserData`, that specifies
additional user code to run upon node startup. This code must be
formatted like other user data--e.g. it must start with an interpreter
line (`#!`)--and will be run after EKS bootrstrapping but before the
node signals readiness to CF.

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -58,7 +58,8 @@ export interface ClusterOptions {
 
     /**
      * Extra code to run on node startup. This code will run after the AWS EKS bootstrapping code and before the node
-     * signals its readiness to the managing CloudFormation stack.
+     * signals its readiness to the managing CloudFormation stack. This code must be a typical user data script:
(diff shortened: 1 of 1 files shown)


nodejs/eks/cluster.ts
```diff
@@ -56,6 +56,12 @@ export interface ClusterOptions {
      */
     nodeRootVolumeSize?: pulumi.Input<number>;
 
+    /**
+     * Extra code to run on node startup. This code will run after the AWS EKS bootstrapping code and before the node
+     * signals its readiness to the managing CloudFormation stack.
+     */
+    nodeUserData?: pulumi.Input<string>;
```
lukehoban: Perhaps include some version of this point in the docs - it isn't obvious that this would be a requirement:

> this code must be formatted like other user data--e.g. it must start with an interpreter line (#!)--and will be run after EKS bootrstrapping but before the node signals readiness to CF.

pgavlin: Done :+1:

PR #20, opened 2018-11-07
Expose AWS role and user mappings

pgavlin:
Allow additional role and user mappings to be added to the `aws-auth`
config map.

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -17,12 +17,53 @@ import * as k8s from "@pulumi/kubernetes";
 import * as pulumi from "@pulumi/pulumi";
 import * as crypto from "crypto";
 import * as fs from "fs";
+import * as jsyaml from "js-yaml";
 import * as path from "path";
 import which = require("which");
 
 import { ServiceRole } from "./servicerole";
 import { createStorageClass, EBSVolumeType, StorageClass } from "./storageclass";
 import transform from "./transform";
+
+/**
+ * RoleMapping describes a mapping from an AWS IAM role to a Kubernetes user and groups.
+ */
+export interface RoleMapping {
+    /**
+     * The ARN of the IAM role to add.
+     */
+    roleArn: pulumi.Input<aws.ARN>;
+
+    /**
+     * The user name within Kubernetes to map to the IAM role. By default, the user name is the ARN of the IAM role.
+     */
+    username: pulumi.Input<string>;
+
+    /**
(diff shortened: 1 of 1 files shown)


pgavlin:
cc @eshamay

nodejs/eks/cluster.ts
```diff
@@ -260,15 +311,38 @@ export class Cluster extends pulumi.ComponentResource {
...
         // Enable access to the EKS cluster for worker nodes.
+        const instanceRoleMapping: RoleMapping = {
+            roleArn: instanceRoleARN,
+            username: "system:node:{{EC2PrivateDNSName}}",
+            groups: ["system:bootstrappers", "system:nodes"],
+        };
+        const roleMappings = pulumi.all([pulumi.output(args.roleMappings || []), instanceRoleMapping])
+            .apply(([mappings, instanceMapping]) => {
+                return jsyaml.safeDump([...mappings, instanceMapping].map(m => ({
+                   rolearn: m.roleArn,
```
hausdorff: Assuming the `rolearn` casing is correct, is there a reason to not just make `RoleMapping` use the same casing, and doing something like `jsyaml.safeDump([...mappings, instanceMapping])`?

nodejs/eks/cluster.ts
```diff
@@ -17,13 +17,54 @@ import * as k8s from "@pulumi/kubernetes";
...
+
+    /**
+     * The user name within Kubernetes to map to the IAM role. By default, the user name is the ARN of the IAM role.
+     */
+    username: pulumi.Input<string>;
+
+    /**
+     * A list of groups within Kubernetes to which the role is mapped.
+     */
+    groups: pulumi.Input<pulumi.Input<string>[]>;
```
hausdorff: Actually, one thing, this type looks like it's on purpose, but it's not clear to me why not just make it `pulumi.Input<string[]>`

lukehoban: It's a little clearer if you write `X[]` as `Array<X>` - `pulumi.Input<Array<pulumi.Input<string>>`.  That is, the array itself can be an Output, or the elements of the array can be Outputs.

hausdorff: @lukehoban mmm, yes, that makes sense and seems obvious in retrospect that you'd want to do this.

nodejs/eks/cluster.ts
pgavlin: The `rolearn` casing is correct, yes. I opted to make the user-facing API as idiomatic as possible.

PR #23, opened 2018-11-14
Allow explicit node subnets

pgavlin:
And improve subnet filtering. Explicit subnets for worker nodes can be
specified if the automatic filtering does not work as expected.
The automatic filtering has been improved by using the presence of a
route to an internet gateway in a subnet's route table as the indicator
that the subnet is a public subnet.

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -564,14 +564,30 @@ ${customUserData}
     }
 }
 
+// computeWorkerSubnets attempts to determine the subset of the given subnets to use for worker nodes.
+//
+// As per https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html, an EKS cluster that is attached to public
+// and private subnets will only expose its API service to workers on the private subnets. Any workers attached to the
+// public subnets will be unable to communicate with the API server.
+//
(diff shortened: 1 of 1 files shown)


nodejs/eks/cluster.ts
```diff
@@ -560,8 +565,30 @@ ${customUserData}
...
-    const publicSubnets = subnets.filter(s => s.mapPublicIpOnLaunch);
-    const privateSubnets = subnets.filter(s => !s.mapPublicIpOnLaunch);
-    return (privateSubnets.length === 0 ? publicSubnets : privateSubnets).map(s => s.id);
+    const publicSubnets: string[] = [];
+    const privateSubnets: string[] = [];
+    for (const subnetId of subnetIds) {
+        const routeTable = await (async () => {
+            try  {
+                return await aws.ec2.getRouteTable({ subnetId: subnetId }, { parent: parent });
+            } catch {
```
lukehoban: Under what condition does this throw?

Generally - comments on what this is trying to do and why it needs to do it this way would be helpful.

pgavlin: > Generally - comments on what this is trying to do and why it needs to do it this way would be helpful.

Agreed. I will comment this up.

PR #38, opened 2018-12-13
Fix up links to repo

lukehoban:
We recently renamed the repo - this patches up some links that had not been updated as part of that change.


lukehoban:
> GH should ensure that the old links into this repo continue to work.

Yes - but Travis does not - which is what I was primarily fixing for here (the badges are currently broken).

PR #41, opened 2019-01-08
decompose cluster creation into reusable modules

houqp:
@pgavlin , this PR aims to finish off your previous WIP commit. Once it's in usable state, i am planning to catch up with master. Then lastly my patches from https://github.com/pulumi/pulumi-eks/pull/22.

Let me know if you prefer to approach this in a different direction.

### DIFF
--- a/nodejs/eks/workerpool.ts
+++ b/nodejs/eks/workerpool.ts
@@ -16,13 +16,8 @@ import * as aws from "@pulumi/aws";
 import * as k8s from "@pulumi/kubernetes";
 import * as pulumi from "@pulumi/pulumi";
 import * as crypto from "crypto";
-import * as fs from "fs";
-import * as path from "path";
-import which = require("which");
 
-import { CoreData, CoreOptions, createCore } from "./core";
 import { ServiceRole } from "./servicerole";
-import { createStorageClass, EBSVolumeType, StorageClass } from "./storageclass";
 import transform from "./transform";
 
 /**
@@ -53,6 +48,16 @@ export interface WorkerPoolOptions {
      * The instance type to use for the cluster's nodes. Defaults to "t2.medium".
      */
     instanceType?: pulumi.Input<aws.ec2.InstanceType>;
+
+    /**
+     * The instance role to use for all nodes in this workder pool.
+     */
+    instanceRole?: pulumi.Input<aws.iam.Role>;
+
+    /**
+     * The security group to use for all nodes in this workder pool.
+     */
+    nodeSecurityGroup?: aws.ec2.SecurityGroup;
 
     /**
      * Public key material for SSH access to worker nodes. See allowed formats at:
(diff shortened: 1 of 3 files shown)


nodejs/eks/workerpool.ts
```diff
@@ -114,32 +124,77 @@ export class WorkerPool extends pulumi.ComponentResource {
     constructor(name: string, args: WorkerPoolOptions, opts?: pulumi.ComponentResourceOptions) {
         super("eks:index:WorkerPool", name, args, opts);
 
-        const pool = createWorkerPool(name, args, this, this.getProvider("kubernetes:core:v1/ConfigMap"));
+        const pool = createWorkerPool(name, args, this);
         this.instanceRole = pool.instanceRole;
         this.nodeSecurityGroup = pool.nodeSecurityGroup;
-
-        this.registerOutputs();
```
pgavlin: This call is required in order to inform the CLI's display routine that the component has finished creating. Let's keep it as-is.

nodejs/eks/workerpool.ts
```diff
@@ -49,11 +44,26 @@ export interface WorkerPoolOptions {
      */
     cluster: aws.eks.Cluster;
 
+    /**
+     * k8s provider for this worker pool
+     */
+    provider: k8s.Provider;
```
pgavlin: `provider` should not be a part of `WorkerPoolOptions`: it is already part of `pulumi.ComponentResourceOptions`, which is passed to the `WorkerPool` constructor. I'd rather we leave it as a separate parameter on `createWorkerPool`.

nodejs/eks/workerpool.ts
```diff
@@ -150,60 +205,24 @@ export function createWorkerPool(name: string, args: WorkerPoolOptions, parent:
...
     const eksClusterIngressRule = new aws.ec2.SecurityGroupRule(`${name}-eksClusterIngressRule`, {
         description: "Allow pods to communicate with the cluster API Server",
         type: "ingress",
         fromPort: 443,
         toPort: 443,
         protocol: "tcp",
-        securityGroupId: args.eksClusterSecurityGroup.id,
-        sourceSecurityGroupId: nodeSecurityGroup.id,
+        securityGroupId: args.clusterSecurityGroup.id,
+        sourceSecurityGroupId: this.nodeSecurityGroup.id,
```
pgavlin: I think this should just be `nodeSecurityGroup` (same on line 224).

nodejs/eks/workerpool.ts
houqp: I see, I did not know that. What's the right way to get the k8s provider from within the constructor through ComponentResourceOptions? The original `this.getProvider("kubernetes:core:v1/ConfigMap")` code results in type error.

pgavlin: What was the type error? That method should exist.

pgavlin: If the type error was complaining about the value possibly being `undefined`, you can do something like:

```
const k8sProvider = this.getProvider("kubernetes:core:v1/ConfigMap");
if (k8sProvider === undefined) {
    throw new RunError("a 'kubernetes' provider must be specified for a 'WorkerPool'");
}
```

houqp: This is exactly what I needed, thanks a lot for the tip :)

houqp:
@pgavlin ready for another round of review. Do you prefer me to cherry pick master commits into this PR or use a separate PR for master sync?

pgavlin:
This looks good to me. Let's use a separate PR into this branch to bring it up-to-date against master.

PR #42, opened 2019-01-09
catch up to master

houqp:
This PR syncs decompose branch with master


houqp:
Quick question, currently, eksNodeAccess is created inside workerpool module, which means there will be one such config map created for each pool. Is this the preferred behavior?

houqp:
Yes, that's what I have in mind as well. Will make the change later today.

houqp:
Ok, I have moved eksNodeAccess config map into core.

houqp:
Thanks for the review, what's the process for getting this merged into master? Do we need to work on unit tests?

pgavlin:
You can open a PR from the `pgavlin/decompose` branch into master and we can take it from there :)

PR #43, opened 2019-01-09
Add public IP option to launch configuration

RichardWLaub:
When deploying worker nodes to private subnets I wanted the option to disable public IPs. This is my first code contribution to a pulumi project so any feedback is appreciated!

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -77,6 +77,12 @@ export interface ClusterOptions {
     vpcId?: pulumi.Input<string>;
 
     /**
+     * Whether or not to associate public IP addresses with the worker nodes.
+     * If unset, the nodes will be assigned public IPs.
+     */
+    associatePublicIpAddress?: pulumi.Input<boolean>;
+
+    /**
      * The subnets to attach to the EKS cluster. If either vpcId or subnetIds is unset, the cluster will use the
      * default VPC's subnets. If the list of subnets includes both public and private subnets, the Kubernetes API
      * server and the worker nodes will only be attached to the private subnets. See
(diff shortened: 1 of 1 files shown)


nodejs/eks/cluster.ts
```diff
@@ -479,7 +485,7 @@ ${customUserData}
         }
 
         const nodeLaunchConfiguration = new aws.ec2.LaunchConfiguration(`${name}-nodeLaunchConfiguration`, {
-            associatePublicIpAddress: true,
+            associatePublicIpAddress: args.associatePublicIpAddress && true,
```
lukehoban: This should be `|| true` (like in the `instanceType` case below).

RichardWLaub:
```
import * as awsinfra from "@pulumi/aws-infra";
import * as eks from "@pulumi/eks";

const prefix = "richardtest";
const numAvailabilityZones = 3;

const network = new awsinfra.Network(`${prefix}-net`, {
    numberOfAvailabilityZones: numAvailabilityZones, // Create subnets in many AZs
    usePrivateSubnets: true,                         // Run compute inside private subnets in each AZ
});

const cluster = new eks.Cluster("cluster", {
    deployDashboard: false,
    desiredCapacity: 4,
    maxSize: 4,
    subnetIds: network.subnetIds,
    vpcId: network.vpcId,
});
```
@lukehoban This was the sample code that was not working before my changes. Basically I expected the worker nodes to be deployed to the private subnets without public IPs assigned. The nodes were deployed to the private subnets but they were assigned public IP addresses anyway. 

Like you mentioned there is probably a better way to set a default using `computeWorkerSubnets` instead. I'll look into that when I have cycles. Thanks for the review!

lukehoban:
Great - yes - that code sample is exactly what I was imagining we should add to test coverage.  There is some code specifically to handle private subnets in the current implementation, but sounds like they don't work correctly currently?  

We may want to have an explicit `usePrivateSubnets` option just to be clear about the intention instead of just inferring it from the subnets passed in. @pgavlin probably has thoughts on this as he added the logic here to do the inference.

metral:
https://github.com/pulumi/pulumi-eks/pull/81 resolved this issue, so closing the PR.

PR #44, opened 2019-01-10
decompose cluster.ts into reusable modules

houqp:


houqp:
I kept the full commit history. @pgavlin let me know if you prefer squashing everything into a single commit or any other way to manage the merge.

nodejs/eks/cluster.ts
```diff
@@ -106,24 +62,29 @@ export interface ClusterOptions {
     instanceType?: pulumi.Input<aws.ec2.InstanceType>;
 
     /**
-     * The AMI to use for worker nodes. Defaults to the value of Amazon EKS - Optimized AMI if no value is provided.
+    * The instance role to use for all nodes in this workder pool.
```
pgavlin: s/workder/worker

nodejs/eks/cluster.ts
```diff
@@ -106,24 +62,29 @@ export interface ClusterOptions {
...
+    */
+    instanceRole?: pulumi.Input<aws.iam.Role>;
+
+    /**
+     * The AMI to use for default worker nodes. Defaults to the value of Amazon EKS - Optimized AMI if no value is provided.
```
pgavlin: s/default worker/worker

nodejs/eks/workerpool.ts
```diff
@@ -0,0 +1,364 @@
...
+
+    /**
+     * The maximum number of worker nodes running in the cluster. Defaults to 2.
+     */
+    maxSize?: pulumi.Input<number>;
+
+    /**
+     * The Amazon VPC CNI plugin for this worker pool's cluster.
+     */
+    vpcCni: VpcCni;
```
pgavlin: I'm a bit unsure about passing `vpcCni` and `eksNodeAccess` as arguments; they're really something of an implementation detail. I'm not sure I have a better option, though.

nodejs/eks/workerpool.ts
```diff
@@ -0,0 +1,364 @@
...
+    vpcId: pulumi.Input<string>;
+
+    /**
+     * The IDs of the cluster subnets to filter for worker pool subnet ids.
+     *
+     * If the list of subnets includes both public and private subnets, the Kubernetes API
+     * server and the worker nodes will only be attached to the private subnets. See
+     * https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html for more details.
+     */
+    clusterSubnetIds: pulumi.Input<string[]>;
```
pgavlin: This should be `pulumi.Input<pulumi.Input<string>[]>`

nodejs/eks/workerpool.ts
houqp: how about passing core or cluster as an argument? this is assuming you always have to associate a worker pool with a specific cluster.

pgavlin: > how about passing core or cluster as an argument

I thought about that briefly. I like the idea of allowing these as arguments, but I'd love to leave this type open-ended enough that it could work with with a cluster created manually (i.e. a cluster provisioned with explicit AWS resources or managed outside of Pulumi). We could perhaps use a union type of `aws.eks.Cluster | Core | Cluster` for the `cluster` property and do different things based on that.

houqp: sounds good 👍

houqp:
@pgavlin I have updated cluster to be a union type of `aws.eks.Cluster | Core | CoreData` and removed `vpcCni` argument.

pgavlin:
> I have updated cluster to be a union type of aws.eks.Cluster | Core | CoreData and removed vpcCni argument.

Can we do the same with `eksNodeAccess`?

houqp:
Removed eksNodeAccess from worker pool option

houqp:
@pgavlin I feel like the way type switch on that union argument is a little bit hacky. Let me know if you have better alternatives.

nodejs/eks/workerpool.ts
```diff
@@ -0,0 +1,363 @@
...
+export interface WorkerPoolData {
+    nodeSecurityGroup: aws.ec2.SecurityGroup;
+    cfnStack: aws.cloudformation.Stack;
+}
+
+export function createWorkerPool(name: string, args: WorkerPoolOptions, parent: pulumi.ComponentResource,  k8sProvider: k8s.Provider): WorkerPoolData {
+    let nodeSecurityGroup: aws.ec2.SecurityGroup;
+    let eksCluster: aws.eks.Cluster;
+    let cfnStackDeps: Array<pulumi.Resource> | undefined;
+    if (Core.isInstance(args.cluster) || isCoreData(args.cluster)) {
```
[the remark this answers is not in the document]
houqp: I like your solution, seems easier to maintain. Adding new types to the union is easy as long as it implements the if check "interface". I wish there is an easy way in typescript to check wither a variable implements `CoreData` interface in one statement without tagging :)

nodejs/eks/core.ts
```diff
@@ -0,0 +1,372 @@
...
+}
+
+/**
+ * Core is a component that wraps the core AWS and Kubernetes resources necessary to run an EKS cluster and its optional
+ * StorageClasses.
+ *
+ * Note that the EKS cluster created by this component will not have any worker nodes, so deploying many important
+ * Kubernetes resources will fail. Worker nodes can be created with the WorkerPool component.
+ */
+export class Core extends pulumi.ComponentResource {
```
lukehoban: I'm not completely sure about the name `Core` here.  And more generally, it seems like we may not need a separate component here.  It might be nicer to make `eks.Cluster` not require any of the other stuff (or allow explicitly asking to not include worker nodes, etc.) so that the `eks.Cluster` component and the `eks.Core` component become the same thing (just with different arguments).

That point is related to making sure we have a really nice interface to `eks.Cluster` to take advantage of using existing worker groups.

lukehoban: For this PR in particular though, feels like we may want to not expose this component, and wait for a follow up focused on improving the `eks.Cluster` API and see if we really need an additional component here when we do that API design?

houqp: > It might be nicer to make eks.Cluster not require any of the other stuff (or allow explicitly asking to not include worker nodes, etc.)

My original PR (#22) includes this change and I have rebased that PR on top of this branch: https://github.com/houqp/pulumi-eks/commit/c0c204c7de65a5e9673ab2267ca7146326a1091e. The full rebase is at: https://github.com/houqp/pulumi-eks/commits/improvement. Once this refactoring is merged, I am going to send a PR for that branch.

So in short, setting `deployDashboard` to `true` and `skipDefaultWorkerPool` to `true` will effectively create an `eks.Cluster` that's equivalent to `Core`.

@pgavlin what's your thought on this?

houqp:
> Take a pass over the eks.Cluster API to make sure it works well with multiple node pool scenarios and generall with the additional flexibility we want to offer there.

We are running #22 in production with 8-10 different types of node pools (mix of cpu/gpu/spot) per cluster. So far it's been flexible enough for our own use. Do you have any particular use-case you want to evaluate?

> Add a handful of tests that cover the core ways of composing these building blocks

I can start to work on the following tests:

1. create cluster with default node pool and attach more node pools to it
2. create cluster without default node pool and attach more node pools to it

Anything else you would like to test?

houqp:
another small issue, should we call it `WorkerPool` or `NodePool`?

houqp:
@pgavlin @lukehoban gentle ping :) What's the process for getting this merged into master?

lukehoban:
@houqp I'm taking a look now, and targeting landing this today along with a couple of new examples.

lukehoban:
Thanks @houqp for all of the work on this!  Can't wait to get the user-facing API improvements in next!

houqp:
Most of the work is done by @pgavlin ;) Thank you both for the review. #22 has been rebased on master and sent out as #47

PR #45, opened 2019-01-15
update deps to fix test run

houqp:
This fixes the test build, but I am getting `Failed to upload test command report to S3: AccessDenied: Access Denied` when running the tests on my local machine, is this expected?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 # It may be tempting to add parens around each individual clause in this expression, but Travis then builds pushes anyway
 if: branch = master OR branch =~ ^release/ OR tag IS present
 language: go
-go: 1.9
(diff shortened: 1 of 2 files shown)


houqp:
Looks like the failure is caused by pulumi using an older version of dep, what's the policy for upgrading version in https://github.com/pulumi/scripts/blob/master/ci/install-common-toolchain.sh?

lukehoban:
We're for now sticking with `dep` 0.4.1 for consistency across all the `pulumi` repos (and to avoid some breaking changes introduced in `dep`).  We plan on trying to update everything to `0.5.0` and/or switch everything to  Go Modules soon.

Closing this out for now, as I believe the other changes here are not needed?

PR #47, opened 2019-01-18
Various improvements

houqp:
This PR brings in following features:

* high level method for creating new node group within a cluster
* support spot instances
* support using existing keypair
* add new option to add custom policy to instance role
* add new option to add extra node labels to each node group
* add option to create a cluster without default node group


houqp:
what's the expected way to run tests locally? I tried `make test_all`, but it gives me `Failed to upload test command report to S3` error. `make test_fast` doesn't seem to run what's in examples folder.

joeduffy:
> what's the expected way to run tests locally? I tried make test_all, but it gives me Failed to upload test command report to S3 error. make test_fast doesn't seem to run what's in examples folder. make travis_pull_request errors out because it's trying to create /opt/pulumi/bin, which requires root.

@ellismg @jen20 @pgavlin  Any idea why running tests locally would be trying to upload to S3? @houqp seems to be running the right commands AFAIK ...

nodejs/eks/nodegroup.ts
```diff
@@ -253,7 +263,7 @@ chmod +x /opt/user-data
 
             return `#!/bin/bash
 
-/etc/eks/bootstrap.sh --apiserver-endpoint "${clusterEndpoint}" --b64-cluster-ca "${clusterCa.data}" "${clusterName}"
+/etc/eks/bootstrap.sh --apiserver-endpoint "${clusterEndpoint}" --b64-cluster-ca "${clusterCa.data}" "${clusterName}" ${bootstrapExtraArgs}
```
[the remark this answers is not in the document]
lukehoban: I've addressed this in commits in #49.

nodejs/eks/cluster.ts
```diff
@@ -462,4 +593,16 @@ export class Cluster extends pulumi.ComponentResource {
 
         this.registerOutputs({ kubeconfig: this.kubeconfig });
     }
+
+    createNodeGroup(name: string, args: ClusterNodeGroupOptions): NodeGroupData {
```
lukehoban: I think we should have this return a real `NodeGroup` component - I've done that in #49.

nodejs/eks/index.ts
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-export { Cluster } from "./cluster";
+export { Cluster, ClusterNodeGroupOptions } from "./cluster";
+export { createNodeGroup, NodeGroup } from "./nodegroup";
+export { createDashboard } from "./dashboard";
```
lukehoban: For now at least, I think we can hold off on exposing these functions, and can bring them back publicly using the same pattern as NodeGroup - a component and potentially also an instance function on the cluster.  This is more consistent with our general API style here and in other libraries.  I've limited the exposed surface area a bit here in #49 to give us options for future API design.

lukehoban:
Merged these commits as part of #49.

PR #63, opened 2019-02-06
support node taints

houqp:


houqp:
A noob TS question, right now a lot of the options are duplicated between ClusterNodeGroupOptions and NodeGroupOptions. In fact, ClusterNodeGroupOptions is a subset of NodeGroupOptions. Does typescript support constructs similar to embedding in golang?

nodejs/eks/nodegroup.ts
```diff
@@ -229,16 +234,29 @@ export function createNodeGroup(name: string, args: NodeGroupOptions, parent: pu
...
+        for (const key of Object.keys(args.taints)) {
+            parts.push(key + "=" + args.taints[key]);
+        }
+        if (parts.length > 0) {
+            kubeletExtraArgs.push("--register-with-taints=" + parts.join(","));
+        }
+    }
+    let bootstrapExtraArgs = "";
+    if (kubeletExtraArgs.length > 0) {
+        bootstrapExtraArgs = ` --kubelet-extra-args '${kubeletExtraArgs.join(" ")}'`;
```
lukehoban: One slight potential issue - the addition of `'` around the args will cause any existing node groups to see a change and force all nodes to get replaced as part of an upgrade to the latest version of the `@pulumi/eks` package.

We could avoid this by only adding the surrounding `'` when the length is greater than 1.

lukehoban:
> A noob TS question, right now a lot of the options are duplicated between ClusterNodeGroupOptions and NodeGroupOptions. In fact, ClusterNodeGroupOptions is a subset of NodeGroupOptions. Does typescript support constructs similar to embedding in golang?

Interfaces can "extends" other interfaces, which could allow a set of options to be shared.  That might be a good idea to factor out these shared properties.

lukehoban:
BTW - @houqp - very sorry for the long delayed review on this PR!

nodejs/eks/cluster.ts
```diff
@@ -464,6 +464,11 @@ export interface ClusterNodeGroupOptions {
      * Custom k8s node labels to be attached to each woker node
      */
     labels?: { [key: string]: string };
+
+    /**
+     * Custom k8s node taints to be attached to each worker node
+     */
+    taints?: { [key: string]: string };
```
lukehoban: Looking at this again - I believe this is not sufficient for `taints` - the `register-with-taints` flag expects something of the form:

```
--register-with-taints=node-role.kubernetes.io/master=true:NoSchedule
```

This has three parts:
* `key`
* `value`
* `effect`

arranged as `key=value:effect` and where `effect` is (currently) one of `NoSchedule`, `NoExecute` or `PreferNoSchedule`.

We could keep this as-is - and just document that the `value` here must be of the form `val:effect`.  But probably better to provide an actual data structure for this.

metral:
- [x] Manually tested - PASSED
- [x] Tested against full CI test suite - PASSED

metral:
closes https://github.com/pulumi/pulumi-eks/issues/79

Issue #84, opened 2019-03-28
Latest EKS AMI will get a windows image.


timmyers:
https://github.com/pulumi/pulumi-eks/blob/976c33bb385d5ac1d2447fcf2c3d9e48d7eeca23/nodejs/eks/nodegroup.ts#L315

This will as of sometime recently result in a Windows AMI, likely unexpectedly.

This should theoretically be fixable by filtering for `platform` as `blank`, according to https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-images.html, however it does not seem there is a way to do that in code and get any results.

lukehoban:
Note that a workaround is to set the `nodeAmiId` parameter to the desired EKS Node AMI from https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html explicitly.

PR #85, opened 2019-03-28
fix(getAmi): allow setting master version & explicitly filter Linux AMIs

metral:
fixes https://github.com/pulumi/pulumi-eks/issues/84

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 - feat(workers): add 'nodeAssociatePublicIpAddress' to toggle public IPs
   [#81](https://github.com/pulumi/pulumi-eks/pull/81)
+- fix(getAmi): allow setting master version & explicitly filter Linux AMIs
+  [#85](https://github.com/pulumi/pulumi-eks/pull/85)
+  - Fix a bug where the wrong AMI was being returned due to a loosely defined
+    regex.
+  - Add support for setting the master / control plane version of the cluster.
 
 ## 0.17.2 (Release March 19, 2019)
 
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -131,6 +131,11 @@ export interface NodeGroupBaseOptions {
      * public IPs.
      */
     nodeAssociatePublicIpAddress?: boolean;
+
+    /**
+     * Desired Kubernetes master / control plane version. If you do not specify a value, the latest available version is used.
+     */
+    version?: pulumi.Input<string>;
 }
 
 /**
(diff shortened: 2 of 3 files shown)


nodejs/eks/nodegroup.ts
```diff
@@ -131,6 +131,11 @@ export interface NodeGroupBaseOptions {
      * public IPs.
      */
     nodeAssociatePublicIpAddress?: boolean;
+
+    /**
+     * Desired Kubernetes master / control plane version. If you do not specify a value, the latest available version is used.
+     */
+    version?: pulumi.Input<string>;
```
lukehoban: I’d suggest for now we don’t add the `version` part of this change - and open an issue to follow up with that. 

The primary thing this version should be used for is the master/control plane as the comment notes - but it is not being used for that yet. 

Also - the way platform versions are specified for EKS is different than how this is being used, and we’ll want to be careful to expose the EKS version concept faithfully. 

This is also not strictly necessary to fix the current high-pri break, and adds some risk to the fix for that.

nodejs/eks/nodegroup.ts
```diff
@@ -310,17 +315,50 @@ ${customUserData}
 `;
         });
 
-    let amiId: pulumi.Input<string> = args.amiId!;
+    let amiId: any = (args.amiId) ? args.amiId : undefined;
```
lukehoban: I don’t follow why this needed to change - and generally this could be simpler (in fact even the original code is more complicated than needed - the `!` should have been unnecessary). 

But most importantly, we should not use `: any` unless absolutely required - it throws away valuable type checking.

nodejs/eks/nodegroup.ts
metral: The version is being passed into the `aws.eks` cluster [constructor](https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/eks/cluster.ts#L69) in https://github.com/pulumi/pulumi-eks/pull/85/files#diff-0dfaec1e9f3a8568a8e2516febe83a69R521.

This is also then leveraged to assure the worker node AMI matches the master version set, in the `getAmi` filter changes below, if set. 

Could you elaborate why you think it's not being used to set the master / control plane version?

nodejs/eks/nodegroup.ts
metral: IIRC using the `!` resulted in subsequent errors about the var being undefined or a promise and this not being able to set to the vars type. I'll look into it, but any other suggestions?

nodejs/eks/nodegroup.ts
lukehoban: Ah - sorry - I misread the code.  Looking again and reviewing the upstream provider `version` specification - this looks exactly right.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - feat(workers): add 'nodeAssociatePublicIpAddress' to toggle public IPs
   [#81](https://github.com/pulumi/pulumi-eks/pull/81)
+- fix(getAmi): allow setting master version & explicitly filter Linux AMIs
```
lukehoban: FWIW - allowing setting the master version is a meaningful feature addition in its own right, independent of fixing the AMI computation.  Might be worth two line items here.

nodejs/eks/nodegroup.ts
```diff
@@ -310,17 +315,50 @@ ${customUserData}
...
+                    },
+                    {
+                        name: "name",
+                        values: [ "amazon-eks-node-*" ],
+                    },
+                ],
+                owners: [ "602401143452" ], // Amazon
+                sortAscending: true,
+            }, { parent: parent });
+            amiId = eksWorkerAmi.then(r => r.ids.pop());
```
lukehoban: Given that this code is nearly identical but has finicky details in it - could we share better by just branching on whether to add `+ version` (or even just set version to `""` if not specified)?

CHANGELOG.md
metral: You got it 👍

nodejs/eks/nodegroup.ts
metral: Yup will do! I was primarily focused on getting a fix out ASAP given the P0 status, but I'm certainly in favor of condensing this.

lukehoban: Yeah - no need to block on this if it holds things up.

nodejs/eks/nodegroup.ts
lukehoban: This could be fixed - but happy to do that later.

PR #87, opened 2019-03-28
Reflect change to aws.iam.InstanceProfile::role typing

jen20:
This change is required because of pulumi/pulumi-aws#506 - we can no longer access the role definition through the instance profile, so instead pass it out of createCore explicitly.

The dev package is pinned because the changes are already on `master`, and need to be removed before merging after 0.18.0 of `pulumi-aws` is released.

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -74,6 +74,7 @@ export interface CoreData {
     subnetIds: pulumi.Output<string[]>;
     clusterSecurityGroup: aws.ec2.SecurityGroup;
     provider: k8s.Provider;
+    instanceRole: pulumi.Output<aws.iam.Role>;
     instanceProfile: aws.iam.InstanceProfile;
     eksNodeAccess?: k8s.core.v1.ConfigMap;
     kubeconfig?: pulumi.Output<any>;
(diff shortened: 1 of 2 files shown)


jen20:
We could make that change instead (and it is likely simpler) - however the change as made here fulfils the current claim of populating with a `Role` object, which we have available to use in `createCore`.

nodejs/eks/cluster.ts
```diff
@@ -252,6 +253,7 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
         kubeconfig: kubeconfig,
         provider: provider,
         vpcCni: vpcCni,
+        instanceProfileRole: eksRole.role,
```
lukehoban: The `eksRole` and `instanceRole` are unrelated, right?  I can’t follow why it’s correct to be passing the former here.

lukehoban: Actually - looks like this can just be `instanceRole: instanceRole`?

jen20:
Force pushed here to update the package.json from a dev package to the release of pulumi-aws 0.18.0. I'll wait for CI here and then merge/release

PR #92, opened 2019-03-29
Fix worker node AMI regex to use a version that matches control plane version, or the latest / default AMI published

metral:


nodejs/eks/nodegroup.ts
```diff
@@ -315,32 +315,41 @@ ${customUserData}
...
         ];
 
         if (version !== undefined) {
             filters.push(
+                // Filter to target a specific k8s version
                 {
-                    name: "name",
-                    values: [ "amazon-eks-node-" + version + "*" ],
+                    name: "description",
+                    values: [ "*k8s*" + version + "*" ],
```
lukehoban: Does this approach of having two filters on `description` work?  I assume so since you’ve tried this - but I hadn’t expected it would.

metral: Yup, you can apparently chain them: https://gist.github.com/metral/4ba5cfe72fffe3c30dc0a57cbd08bd97

Issue #82, opened 2019-03-22
Create `instanceRole` per-nodegroup not per-cluster


lukehoban:
Currently we are creating an instance role per cluster which is reused for each NodeGroup.  This is not always sufficient, there are cases where different NodeGroups want different Roles.  The default policies will be the same for each of these roles, but users may want to attach additional policies themselves to only some NodeGroups.

This also more closely aligns with the documentation and CloudFormation template used by EKS:
* https://github.com/awslabs/amazon-eks-ami/blob/master/amazon-eks-nodegroup.yaml#L184
* https://docs.aws.amazon.com/eks/latest/userguide/launch-workers.html

naineel:
We are still blocked on this. What's the timeline on the release of 0.22 ?

PR #98, opened 2019-04-02
feat(nodePools): support per-nodegroup IAM instance roles

metral:
resolves: https://github.com/pulumi/pulumi-eks/issues/82

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -74,8 +74,7 @@ export interface CoreData {
     subnetIds: pulumi.Output<string[]>;
     clusterSecurityGroup: aws.ec2.SecurityGroup;
     provider: k8s.Provider;
-    instanceRole: pulumi.Output<aws.iam.Role>;
-    instanceProfile: aws.iam.InstanceProfile;
+    instanceProfile?: aws.iam.InstanceProfile;
     eksNodeAccess?: k8s.core.v1.ConfigMap;
     kubeconfig?: pulumi.Output<any>;
     vpcCni?: VpcCni;
@@ -88,6 +87,10 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
         which.sync("aws-iam-authenticator");
     } catch (err) {
         throw new Error("Could not find aws-iam-authenticator for EKS. See https://github.com/pulumi/eks#installing for installation instructions.");
+    }
+
+    if (args.instanceRole && args.instanceRoles) {
+        throw new Error("instanceRole and instanceRoles are mutually exclusive, and cannot both be set.");
     }
 
     // If no VPC was specified, use the default VPC.
@@ -184,12 +187,26 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
     // Create the VPC CNI management resource.
     const vpcCni = new VpcCni(`${name}-vpc-cni`, kubeconfig, args.vpcCniOptions, { parent: parent });
 
-    // Create the instance role we'll use for worker nodes.
-    let instanceRole: pulumi.Output<aws.iam.Role>;
-    if (args.instanceRole) {
-        instanceRole = pulumi.output(args.instanceRole);
+    let instanceRoleMappings: pulumi.Output<RoleMapping[]>;
+    let instanceProfile: aws.iam.InstanceProfile | undefined;
+    // Create role mappings of the instance roles specified for aws-auth.
+    if (args.instanceRoles) {
+        instanceRoleMappings = pulumi.output(args.instanceRoles).apply(instanceRoles =>
+            instanceRoles.map(role => createInstanceRoleMapping(role.arn)),
+        );
+    } else if (args.instanceRole) {
+        // Create an instance profile if using a default node group
+        if (!args.skipDefaultNodeGroup) {
+            instanceProfile = new aws.iam.InstanceProfile(`${name}-instanceProfile`, {
+                role: args.instanceRole,
+            }, { parent: parent });
+        }
+
+        instanceRoleMappings = pulumi.output(args.instanceRole).apply(instanceRole =>
+            [createInstanceRoleMapping(instanceRole.arn)],
+        );
     } else {
-        instanceRole = (new ServiceRole(`${name}-instanceRole`, {
+        const instanceRole = (new ServiceRole(`${name}-instanceRole`, {
             service: "ec2.amazonaws.com",
             managedPolicyArns: [
                 "arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy",
@@ -197,25 +214,30 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
                 "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly",
             ],
         }, { parent: parent })).role;
-    }
-    const instanceRoleARN = instanceRole.apply(r => r.arn);
 
-    if (args.customInstanceRolePolicy) {
-        const customRolePolicy = new aws.iam.RolePolicy(`${name}-EKSWorkerCustomPolicy`, {
-            role: instanceRole,
-            policy: args.customInstanceRolePolicy,
-        });
+        // Create a new policy for the role, if specified.
+        if (args.customInstanceRolePolicy) {
+            const customRolePolicy = new aws.iam.RolePolicy(`${name}-EKSWorkerCustomPolicy`, {
+                role: instanceRole,
+                policy: args.customInstanceRolePolicy,
+            });
+        }
+
+        // Create an instance profile if using a default node group
+        if (!args.skipDefaultNodeGroup) {
+            instanceProfile = new aws.iam.InstanceProfile(`${name}-instanceProfile`, {
+                role: instanceRole,
+            }, { parent: parent });
+        }
+
+        instanceRoleMappings = pulumi.output(instanceRole).apply(role =>
+            [createInstanceRoleMapping(role.arn)],
+        );
     }
 
-    // Enable access to the EKS cluster for worker nodes.
-    const instanceRoleMapping: RoleMapping = {
-        roleArn: instanceRoleARN,
-        username: "system:node:{{EC2PrivateDNSName}}",
-        groups: ["system:bootstrappers", "system:nodes"],
-    };
-    const roleMappings = pulumi.all([pulumi.output(args.roleMappings || []), instanceRoleMapping])
-        .apply(([mappings, instanceMapping]) => {
-            return jsyaml.safeDump([...mappings, instanceMapping].map(m => ({
+    const roleMappings = pulumi.all([pulumi.output(args.roleMappings || []), instanceRoleMappings])
+        .apply(([mappings, instanceMappings]) => {
+            return jsyaml.safeDump([...mappings, ...instanceMappings].map(m => ({
                 rolearn: m.roleArn,
                 username: m.username,
                 groups: m.groups,
@@ -242,10 +264,6 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
         data: nodeAccessData,
     }, { parent: parent, provider: provider });
 
-    const instanceProfile = new aws.iam.InstanceProfile(`${name}-instanceProfile`, {
-        role: instanceRole,
-    }, { parent: parent });
-
     return {
         vpcId: pulumi.output(vpcId),
         subnetIds: pulumi.output(subnetIds),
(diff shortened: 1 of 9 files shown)


nodejs/eks/cluster.ts
```diff
@@ -74,13 +74,18 @@ export interface CoreData {
...
     clusterSecurityGroup: aws.ec2.SecurityGroup;
     provider: k8s.Provider;
-    instanceRole: pulumi.Output<aws.iam.Role>;
     instanceProfile: aws.iam.InstanceProfile;
     eksNodeAccess?: k8s.core.v1.ConfigMap;
     kubeconfig?: pulumi.Output<any>;
     vpcCni?: VpcCni;
 }
 
+const eksManagedPolicyArns: string[] = [
```
lukehoban: Nit - but perhaps calling out (in the variable name) that these are the "instance" role policies might make sense.  There are two roles involved, and the sets of policies for each are different.

nodejs/eks/cluster.ts
```diff
@@ -254,12 +270,43 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
+                policyArn: policy,
+                role: role,
+            },
+        );
+    }
+}
+
+// Enable access to the EKS cluster for worker nodes by creating
+// the instance role mapping for aws-auth.
+export function createInstanceRoleMapping(arn: pulumi.Input<aws.ARN>): RoleMapping {
```
lukehoban: For exported function - should use `/** * **/` style docs so they get picked up by doc generators.

nodejs/eks/cluster.ts
```diff
@@ -254,12 +270,43 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
+        // Create RolePolicyAttachment without returning it.
+        const rpa = new aws.iam.RolePolicyAttachment(
+            `${name}-policy-${counter++}`,
+            {
+                policyArn: policy,
+                role: role,
+            },
+        );
+    }
+}
```
lukehoban: Not sure we should include this.  It's very general - not something related to EKS at all.  If we were to add a helper like this - we could do so in the AWS pacakge.

nodejs/eks/cluster.ts
```diff
@@ -184,38 +189,53 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
+    const instanceRoleMappings: RoleMapping[] = [];
+    let instanceProfile: aws.iam.InstanceProfile;
+    if (args.instanceRoleArns && args.instanceRoleArns.length) {
+        // Create role mappings if an array of instanceRoleArn's is specified.
+        for (const [index, arn] of args.instanceRoleArns.entries()) {
+            const roleName = pulumi.output(arn).apply(r => r.split("/")).apply(r => r[1]);
+            const role = aws.iam.Role.get(`${name}-existingRole-${index}`, roleName);
+
+            // Add the EKS IAM Policies to the roles provided to join the cluster.
+            addPoliciesToExistingRole(`${name}-eksServicePolicies-role-${index}`, role, eksManagedPolicyArns);
```
lukehoban: I suggest we don't do this - and just have the documentation indicate that this role should be provided at least a specific set of managed policies.

nodejs/eks/cluster.ts
```diff
@@ -184,38 +189,53 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
     // Create the VPC CNI management resource.
     const vpcCni = new VpcCni(`${name}-vpc-cni`, kubeconfig, args.vpcCniOptions, { parent: parent });
 
-    // Create the instance role we'll use for worker nodes.
-    let instanceRole: pulumi.Output<aws.iam.Role>;
-    if (args.instanceRole) {
-        instanceRole = pulumi.output(args.instanceRole);
+    const instanceRoleMappings: RoleMapping[] = [];
+    let instanceProfile: aws.iam.InstanceProfile;
+    if (args.instanceRoleArns && args.instanceRoleArns.length) {
```
lukehoban: Why `&& length`?

nodejs/eks/cluster.ts
```diff
@@ -184,38 +189,53 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
+        // Create role mappings if an array of instanceRoleArn's is specified.
+        for (const [index, arn] of args.instanceRoleArns.entries()) {
+            const roleName = pulumi.output(arn).apply(r => r.split("/")).apply(r => r[1]);
+            const role = aws.iam.Role.get(`${name}-existingRole-${index}`, roleName);
```
lukehoban: I don't think it's a good idea to use `.get`.  I had assumed we would just ask the user to pass us `Role`s, and they can do the `.get` if that's really what the need.  But more often they'll just construct the `Role` themselves.

nodejs/eks/cluster.ts
```diff
@@ -312,6 +359,12 @@ export interface ClusterOptions {
      */
     instanceRole?: pulumi.Input<aws.iam.Role>;
 
+    /**
+     * The array of instance role ARNs to map into the aws-auth ConfigMap. This
+     * is a required step to specify a NodePool's instanceRoleArn.
+     */
+    instanceRoleArns?: pulumi.Input<aws.ARN>[];
```
lukehoban: Per above - I think a lot of this is simpler if we just do `instanceRoles?`?

Also - should add notes to both this and `instanceRole` that they are mutually exclusive.  As well as checks when reading from these that enforce and throw if both are provided.

nodejs/eks/cluster.ts
```diff
@@ -184,38 +189,53 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
+        } else {
+            instanceRole = (new ServiceRole(`${name}-instanceRole`, {
+                service: "ec2.amazonaws.com",
+                managedPolicyArns: eksManagedPolicyArns,
+            }, { parent: parent })).role;
+        }
+        const instanceRoleArn = instanceRole.apply(r => r.arn);
+
+        // Create a new policy for the role, if specified.
+        if (args.customInstanceRolePolicy) {
```
lukehoban: We could choose if we want to proactively mark this argument as "deprecated" and point to the ability to pass `instanceRoles` instead?

nodejs/eks/cluster.ts
```diff
@@ -184,38 +189,53 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
     // Create the VPC CNI management resource.
     const vpcCni = new VpcCni(`${name}-vpc-cni`, kubeconfig, args.vpcCniOptions, { parent: parent });
 
-    // Create the instance role we'll use for worker nodes.
-    let instanceRole: pulumi.Output<aws.iam.Role>;
-    if (args.instanceRole) {
-        instanceRole = pulumi.output(args.instanceRole);
+    const instanceRoleMappings: RoleMapping[] = [];
+    let instanceProfile: aws.iam.InstanceProfile;
```
lukehoban: Should this be `aws.iam.InstanceProfile | undefined` since there is a branch where it is unset?

I think that also avoids the strange `x! || undefined` later on?

nodejs/eks/nodegroup.ts
```diff
@@ -136,6 +136,13 @@ export interface NodeGroupBaseOptions {
      * Desired Kubernetes master / control plane version. If you do not specify a value, the latest available version is used.
      */
     version?: pulumi.Input<string>;
+
+    /**
+     * The instance role ARN to use for this node group. This instance role ARN
+     * must be supplied along with any other required instance roles to the
+     * cluster's instanceRoleArns'
+     */
+    instanceRoleArn?: pulumi.Input<aws.ARN>;
```
lukehoban: Again - would just accept the `Role` directly here.

lukehoban: Actually - for even more user-flexibility here - any reason not to accept the `InstanceProfile` instead?  That lets users who care about this provide any profile they want - and avoids the need to construct the profile from the role in this library?

naineel:
I have a question: Why is it expected out of the users to create the default Roles as required for the node groups ? Can't pulumi handle this ? 
I can still attach additional policies to the role by accessing the arn and the creating the policy i need to attach to it. Since the default 3 policies which are attached are always going to remain the same.

metral:
@lukehoban Feedback has been addressed. PTAL

metral:
@naineel For us to be able to register and map the IAM roles into `aws-auth`, we need to know the set of roles intended to be used when the cluster is created, so we cannot make creation of the `Role` part of the `NodeGroup`. There's also the scenario where we may not also be able to modify the roles  to attach policies due to possible restrictions on the IAM resources. Therefore, to guarantee expectations users should create the roles in its totality first, and then pass them into the cluster. 

We've added documentation for this requirement in the commit in both the `ClusterOptions` and in the nodegroup example: `nodejs/eks/examples/nodegroup/index.ts`

naineel:
@metral not sure if I understand. I can add and remove roles/users from aws-auth at any time, it doesn't need to be at cluster startup. It's just a config map on the cluster after all.
To your second point, I should be able to attachPolicies to an existing role like:
```
const test_attach = new aws.iam.RolePolicyAttachment("test-attach", {
      policyArn: policy.arn,
      role: role.name,
 });
```
If the nodeGroup can expose the instanceProfile/role object then I should be able to get the role arn/names.
Am I missing something ?

metral:
Specifying the roles isn't particular to cluster startup, apologies if this caused confusion. The path forward is for the cluster to require a list of roles as an option in the cluster to map the roles into aws-auth through `instanceRoles`. If at any time you wish to add new roles after a cluster is up, amending the `instanceRoles` `ClusterOption` will ensure it knows about all of the roles it needs to map into the k8s RBAC.

> To your second point, I should be able to attachPolicies to an existing role like

Correct, that is how you would attach policies to a role. Since you're supplying the role to the cluster you can certainly tweak this as needed, as the role's ARN is already in `aws-auth` via `instanceRoles`, and the `instanceProfile` used with the NodeGroup would be referencing the role, so you wouldn't have to modify either after the fact.

naineel:
@metral So I'm still trying to understand why is it a bad idea for pulumi to handle this and manage it for us ?

nodejs/eks/cluster.ts
```diff
@@ -184,38 +187,54 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
-                "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly",
-            ],
-        }, { parent: parent })).role;
-    }
-    const instanceRoleARN = instanceRole.apply(r => r.arn);
+        // Create the instance role we'll use for worker nodes.
+        let instanceRole: pulumi.Output<aws.iam.Role>;
+        if (args.instanceRole) {
+            instanceRole = pulumi.output(args.instanceRole);
+        } else {
```
lukehoban: The logic here feels more complex than it needs to be.  There are three cases right - `if (args.instanceRoles)` or `if (args.instanceRole)` or `else`.  And all three branches lead to a list of `instanceRoles`.  And then in all cases we create a `roleMappings`.  

I may be missing something - but this looks like there's a fair bit of duplication and missing symmetry which makes it hard for me to be sure we aren't missing something on one of these branches.

nodejs/eks/nodegroup.ts
```diff
@@ -212,6 +218,10 @@ function isCoreData(arg: NodeGroupOptionsCluster): arg is CoreData {
 }
 
 export function createNodeGroup(name: string, args: NodeGroupOptions, parent: pulumi.ComponentResource,  k8sProvider: k8s.Provider): NodeGroupData {
+    if (!args.instanceProfile) {
+        throw new Error(`an instanceProfile is required`);
+    }
```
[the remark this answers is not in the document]
lukehoban: It looks like we still do publish the `instanceProfile` onto `core` in all the same cases as we used to - so seems we can make this not a breaking change with `!args.instanceProfile && !code.instanceProfile` (along with corresponding change to the code below that actually reads from these)?

lukehoban:
>  So I'm still trying to understand why is it a bad idea for pulumi to handle this and manage it for us ?

The approach here gives users who want to control the set of roles being used complete control over those roles and the associated instance profiles.  It looks like the most flexible way to enable this while ensuring that all the related requests to this are now possible.

The examples in the current PR don't make it clear - but this should be very simple and incremental to use - and I believe very "pay-for-play" on complexity for users who want to customize the roles they use for each NodeGroup.

The design here also ensures that things work cleanly with the desired-state model of Pulumi (similar to CloudFormation, Terraform, and Kuberentes) which wants to be able to assert a single description of the state that every resource should be in.  To do that, it is important that for any given deployment, the set of Roles be known in one place so that the desired state of the ConfigMap can be established cleanly.  This ensures repeatable deployments and clean migration from any state of the system to other valid states without the need for operator intervention.

Ultimately though - this approach should be very simple and incremental to apply to your use cases, and should require only a few additional lines of code beyond the approach I think you were originally expecting.  I believe the increased flexibility of the approach in this PR (and lack of need for breaking changes) make this a good solution that will also support the next set of related extensibility points that are likely to be needed.

metral:
@lukehoban Feedback has been addressed. PTAL

nodejs/eks/examples/nodegroup/iam.ts
```diff
@@ -0,0 +1,27 @@
+import * as aws from "@pulumi/aws";
+import * as pulumi from "@pulumi/pulumi";
+
+const managedPolicyArns: string[] = [
+    "arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy",
+    "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy",
+    "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly",
+];
```
lukehoban: Possibly for a follow-up PR - but I wonder if it would be helpful to export this array from the package so this doesn't have to be hard-coded into the user-code?   

We could even add a helper that does the same as the `createRole` below as a convenience thing - like `eks.createDefaultInstanceRole`?

PR #102, opened 2019-04-15
Add support for `assumeRole` in kubeconfig

igorshapiro:
Addresses https://github.com/pulumi/pulumi/issues/2537.

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -80,6 +80,42 @@ export interface CoreData {
     vpcCni?: VpcCni;
 }
 
+export interface EnvVar {
+    name: string;
+    value: string;
+}
+
+export interface KubeConfigAuth {
+    apiVersion: string;
+    command: string;
+    args: string[];
+    env: EnvVar[];
+}
+
+function translateAWSProviderToIAMAuth(clusterName: string, profile?: string, assumeRole?: { roleArn?: string }): KubeConfigAuth {
+    const result: KubeConfigAuth = {
+        apiVersion: "client.authentication.k8s.io/v1alpha1",
+        command: "aws-iam-authenticator",
+        args: ["token", "-i", clusterName],
+        env: [],
+    };
+
+    if (typeof assumeRole === "string") {
+        assumeRole = JSON.parse(<string>assumeRole);
+    }
+
+    if (assumeRole && assumeRole.roleArn) {
+        result.args.push("-r");
+        result.args.push(assumeRole.roleArn);
+    }
+
+    if (profile) {
+        result.env.push({ name: "AWS_PROFILE", value: profile });
+    }
+
+    return result;
+}
+
 export function createCore(name: string, args: ClusterOptions, parent: pulumi.ComponentResource): CoreData {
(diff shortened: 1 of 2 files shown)


nodejs/eks/cluster.ts
```diff
@@ -80,6 +80,29 @@ export interface CoreData {
     vpcCni?: VpcCni;
 }
 
+function translateAWSProviderToIAMAuth(clusterName: string, providerProfile?: string, providerAssumeRole?: { roleArn?: string }): pulumi.Output<any> {
+    const command = "aws-iam-authenticator";
+    const args = ["token", "-i", clusterName];
+    return pulumi.all([providerProfile, providerAssumeRole]).apply(([profile, assumeRole]) => {
```
lukehoban: The `pulumi.all` should not be needed here - because the arguments passed to it are not `Output`s. You should be able to use the values of `profile` and `assumeRole` directly.

nodejs/eks/cluster.ts
```diff
@@ -80,6 +80,29 @@ export interface CoreData {
...
+        const result: {[k: string]: any} = { args, command, apiVersion: "client.authentication.k8s.io/v1alpha1" };
+
+        if (typeof assumeRole === "string") {
```
lukehoban: Can this ever be true?  The TypeScript type doesn't think so, and the typing of the `assumeRole` in the provider doesn't appear to have any case where this would be a string.

nodejs/eks/cluster.ts
```diff
@@ -80,6 +80,29 @@ export interface CoreData {
     vpcCni?: VpcCni;
 }
 
+function translateAWSProviderToIAMAuth(clusterName: string, providerProfile?: string, providerAssumeRole?: { roleArn?: string }): pulumi.Output<any> {
+    const command = "aws-iam-authenticator";
+    const args = ["token", "-i", clusterName];
+    return pulumi.all([providerProfile, providerAssumeRole]).apply(([profile, assumeRole]) => {
+        const result: {[k: string]: any} = { args, command, apiVersion: "client.authentication.k8s.io/v1alpha1" };
```
lukehoban: Perhaps instead of this - just define the interface we expect and then use that here?  A little heavier, but hopefully makes this easier to ensure is correct?

nodejs/eks/cluster.ts
igorshapiro: For some reason, it was a serialized json. I know that in the configuration you should specify the `assumeRole` as json

nodejs/eks/cluster.ts
igorshapiro: Fixed

nodejs/eks/cluster.ts
igorshapiro: Fixed

metral:
@igorshapiro Thanks for the work up to now on this issue. What is the current state of the PR - is it ready for review?

Also, it'd be great if we could have tests to support this - have you looked into this by chance?

metral:
@igorshapiro Is this PR still being worked on?

metral:
Closing this PR as it seems to have gone stale. Please re-open this PR if it gets picked up again.

PR #104, opened 2019-04-19
Move to Go Modules

jen20:
This PR moves `pulumi-eks` to using Go Modules for managing Go dependencies.


CyrusNajmabadi:
I am not qualified to review this :)  Maybe @ellismg .  All i care about is that this works smoothly :)

go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/pulumi/pulumi-eks
+
+go 1.12
+
+require (
+	github.com/pulumi/pulumi v0.17.8-0.20190418223219-0ede30fdb624
```
ellismg: What is the workflow for updating this, @jen20?  I know I need to construct a pseudo version but is there a good way gesture in the toochain today for "pick up HEAD"?

jen20: Yes - it's `GO111MODULE=on go get github.com/pulumi/pulumi@master` - the pseudo version will be determined automatically, and transitive dependencies will be updated. The `GO111MODULE=on` is required if the repository is inside `GOPATH` with Go 1.12.

ellismg: :+1:

Issue #106, opened 2019-04-22
Move from secgroup in-line rules to standalone secgroup rules


metral:
TF has a long, outstanding bug with the use of in-line rules in secgroups and how they're gathered (https://github.com/terraform-providers/terraform-provider-aws/pull/4416), which still exists though the PR suggests otherwise.

TF also states that the conjunction of secgroups with in-line rules and standalone rules [do not play](https://www.terraform.io/docs/providers/aws/r/security_group.html) well together, and we use both in pulumi/eks.

Moving to standalone secgroup rules is a better approach long-term that we should adopt.

ref: https://github.com/pulumi/pulumi-eks/issues/69

Issue #69, opened 2019-02-25
Unable to replace EKS cluster: Error revoking security group ingress rules


colinbankier:
Performing a change that requires an EKS cluster to be replaced fails, and leaves the stack in a state which cannot be destroyed nor updated.

Pulumi version: v0.16.14

I have a stack containing an EKS cluster. Updating the subnetIds causes a "replace" (as expected), however the replace fails.

```
Diagnostics:
  aws:ec2:SecurityGroup (cluster-nodeSecurityGroup):
    error: Plan apply failed: 1 error occurred:
    
    * updating urn:pulumi:test::eks-test::eks:index:Cluster$aws:ec2/securityGroup:SecurityGroup::cluster-nodeSecurityGroup: Error revoking security group ingress rules: InvalidPermission.NotFound: The specified rule does not exist in this security group.
    	status code: 400, request id: 577f76ae-06d1-4030-8f84-a8844b89a6d3
```
The stack is then in a bad state, as the EKS cluster was replaced, but other resources were not updated.
Further updates (e.g. trying to revert to the previous subnetIds) fail with the same issue.
If I have deployments on the cluster, I cannot destroy the stack either, as these cannot be deleted (the cluster they belonged to no longer exists).

colinbankier:
Nothing non-default  - I reproduced the issue with an extremely minimal config, just
```
const cluster = new eks.Cluster("cluster", {
    vpcId: "vpc-xxxx",
    subnetIds: ["subnet-xxxx", "subnet-xxxx"],
    instanceType: "r5.large",
    desiredCapacity: 2,
    minSize: 2,
    maxSize: 2,
    storageClasses: "gp2",
    deployDashboard: true,
});
```
And then added some subnets to the list and the same issue occurred.

ggilmore:
Any updates for this issue? I just ran into the same situation.

metral:
This appears to be related to a long-standing TF bug when using descriptions in the security group rules: https://github.com/terraform-providers/terraform-provider-aws/issues/2879, and a fix supposedly came in terraform-provider-aws 1.19.0 but doesn't appear to be a remedy in this scenario.

Per user comments, removing the descriptions from the security group rules seems to skate around this issue. thoughts @jen20 @lukehoban

metral:
[...]

PR #108, opened 2019-04-23
fix(secgroups): use standalone secgroup rules instead of in-line rules

metral:
fixes: https://github.com/pulumi/pulumi-eks/issues/106
rel: https://github.com/pulumi/pulumi-eks/issues/69


metral:
Replaced with #109

PR #109, opened 2019-04-23
fix(secgroups): use standalone secgroup rules instead of in-line rules

metral:
fixes: https://github.com/pulumi/pulumi-eks/issues/106
rel: https://github.com/pulumi/pulumi-eks/issues/69

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ### Improvements
 
+- fix(secgroups): use standalone secgroup rules instead of in-line rules
+  [#108](https://github.com/pulumi/pulumi-eks/pull/108)
 - fix(nodegroup): filter on x86_64 arch for node AMI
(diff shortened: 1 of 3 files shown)


metral:
Note: this PR currently sets the ingress and egress prop of both secgroups to an empty array `[]`. This is necessary as there seems to be a bug when removing both props all-together. I'll be opening a separate issue to track this

metral:
@lukehoban PTAL

metral:
@lukehoban agree on the fact that we should enable testing prev versions against the new commits.

> Do you know exactly what happens during the update?

I'd imagine that there can be a brief period of unavailability given that the secgroup rules are against the worker instances themselves. That said, this patch update does happen in a matter of seconds (< 10 sec and closer to 5sec) so it isn't a prolonged period.

I'll make sure to mention this ^ as a part of the changelog.

PR #114, opened 2019-04-24
Only use the smart default for AMI at creation time.

### LINKED ISSUE #88: Pulumi shouldn't fetch the latest AMI by default
When using `pulumi update` with an EKS cluster, pulumi's default behavior is to always [fetch the latest version of the EKS AMI](https://github.com/pulumi/pulumi-eks/blob/1145bdaae8f891239cb79f0c2d8ff4f9233e14ef/nodejs/eks/nodegroup.ts#L320-L359) if [nodeAmiId](https://pulumi.io/reference/pkg/nodejs/@pulumi/eks/#ClusterOptions-nodeAmiId) isn't specified.

While this behavior can be convenient, unexpected upgrades to the EKS AMI can cause unexpected downtime or [break the cluster entirely](https://github.com/pulumi/pulumi-eks/issues/84). 

For production situations, it's [strongly recommended that you explicitly pass the nodeAmiId parameter so that you have full control over _when_ your nodes upgrade](https://pulumi-community.slack.com/archives/C84L4E3N1/p1553793187012100?thread_ts=1553791785.005000&cid=C84L4E3N1). 

pulumi-eks's default behavior should enforce best practices. I propose the following: 

1. Make `nodeAmiId` a _required_ parameter (force users to make a conscious decision about the AMI that they want to use).
1. Provide [the existing fetch behavior](https://github.com/pulumi/pulumi-eks/blob/1145bdaae8f891239cb79f0c2d8ff4f9233e14ef/nodejs/eks/nodegroup.ts#L320-L359) only as a convenience function for users that don't need fine control over when their cluster upgrades. 
1. Warn people in the documentation for the above helper function about not using this in production.

_See https://pulumi-community.slack.com/archives/C84L4E3N1/p1553791785005000 for more discussion._

lukehoban:
When we use the "smart default" to pick the most recent EKS-optimized AMI, we now also mark the image as `ignoreChanges` to ensure that we do not pick up new "smart defaults" later.  Users can still explicitly specify an `amiID`, and if they do this will also remove the `ignoreChanges` annotation so that the change will be applied.

Fixes #88.

### DIFF
--- a/nodejs/eks/package.json
+++ b/nodejs/eks/package.json
@@ -11,9 +11,9 @@
     "homepage": "https://pulumi.io",
     "repository": "https://github.com/pulumi/pulumi-eks",
     "dependencies": {
-        "@pulumi/aws": "^0.18.0",
+        "@pulumi/aws": "dev",
         "@pulumi/kubernetes": "^0.21.0",
-        "@pulumi/pulumi": "^0.17.4",
+        "@pulumi/pulumi": "^0.17.8",
         "which": "^1.3.1"
     },
     "devDependencies": {
--- a/nodejs/eks/package.json
+++ b/nodejs/eks/package.json
@@ -11,7 +11,7 @@
     "homepage": "https://pulumi.io",
     "repository": "https://github.com/pulumi/pulumi-eks",
     "dependencies": {
-        "@pulumi/aws": "dev",
+        "@pulumi/aws": "^0.18.3",
         "@pulumi/kubernetes": "^0.21.0",
         "@pulumi/pulumi": "^0.17.8",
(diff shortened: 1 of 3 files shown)


nodejs/eks/package.json
```diff
@@ -11,9 +11,9 @@
     "homepage": "https://pulumi.io",
     "repository": "https://github.com/pulumi/pulumi-eks",
     "dependencies": {
-        "@pulumi/aws": "^0.18.0",
+        "@pulumi/aws": "dev",
```
lukehoban: I will update this to a released build before merging.

PR #121, opened 2019-05-02
chore(cluster): de-dupe creation of nodeSecurityGroup & ingress rule

metral:
In general, the `nodeSecurityGroup` and `eksClusterIngressRule` resources are
better suited to live and be managed in the `NodeGroup` class instead of
also in the `Cluster` class, and minimizing the areas to configure these
resources is better in the long-run.

The `nodeSecurityGroup` and its related `eksClusterIngressRule` are **being
removed** from the `Cluster` class, as these steps are [duplicated and handled](https://github.com/pulumi/pulumi-eks/blob/master/nodejs/eks/nodegroup.ts#L245-L260)
in the `NodeGroup` class already, if a user does not supply them.

Further examples:

  - The user could expect to provide the tags in either the `Cluster` or
  `NodeGroup` class, and this could cause confusion.
  - If the user sets `skipDefaultNodeGroup: true`, uses the `Cluster` API method
  `eks.Cluster.createNodeGroup()`, and supplies the `nodeSecurityGroup` tags
  through the API call, these tags will not be applied.
  This is because the cluster's `nodeSecurityGroup` is being created in the
  `Cluster` class, and passed as an arg to the `NodeGroup` using the API call in
  `eks.Cluster.createNodeGroup()`. And we can't tag the
  `nodeSecurityGroup` as its a ref with `.tags` set to `readonly`.

By removing both of these resources from the `Cluster` class, we'll simplify
working with the `nodeSecurityGroup` and its related rules, and make the
`NodeGroup` path and its specific resources more straight-forward.

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -501,16 +501,6 @@ export class Cluster extends pulumi.ComponentResource {
     public readonly instanceRole: pulumi.Output<aws.iam.Role>;
 
     /**
-     * The security group for the cluster's nodes.
-     */
-    public readonly nodeSecurityGroup: aws.ec2.SecurityGroup;
-
-    /**
-     * The ingress rule that gives node group access to cluster API server
-     */
-    public readonly eksClusterIngressRule: aws.ec2.SecurityGroupRule;
-
-    /**
      * The default Node Group configuration, or undefined if `skipDefaultNodeGroup` was specified.
      */
     public readonly defaultNodeGroup: NodeGroupData | undefined;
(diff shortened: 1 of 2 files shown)


metral:
This was tested on an existing cluster using the 0.18.3 and PASSED

nodejs/eks/cluster.ts
```diff
@@ -500,16 +500,6 @@ export class Cluster extends pulumi.ComponentResource {
...
 
-    /**
-     * The security group for the cluster's nodes.
-     */
-    public readonly nodeSecurityGroup: aws.ec2.SecurityGroup;
-
-    /**
-     * The ingress rule that gives node group access to cluster API server
-     */
-    public readonly eksClusterIngressRule: aws.ec2.SecurityGroupRule;
```
lukehoban: This is an API breaking change.  It's somewhat unlikely this is being used though - and it should in practice be easy enough to work around by referring to the properties on the nodegroup.  But may be worth calling out as breaking change in the `CHANGELOG`.

lukehoban: It looks like the `NodeGroupData` exposed as `defaultNodeGroup` does not include the ingress rule. Should it?  I can't think of any specific scenario for this - but its a resource we create that we used to make available but no longer will after these changes.

metral: Good point - I'll have to think on this one some more

metral:
Closing due to being too disruptive of a change for existing clusters.

PR #126, opened 2019-05-03
fix(nodeSecurityGroupTags): only expose option through Cluster class

### LINKED ISSUE #125: nodeSecurityGroupTags is not being applied when provided in a NodeGroup
Due to the assumed (but forgotten) dep on #121, which is turning out to not be a good path forward due to its disruptiveness, PR #122 intro'd the shift over to specifying the `nodeSecurityGroupTags` in the `NodeGroup` class only, instead of also the `Cluster` class for ease-of-use, and compartmentalization of code & config.

Given that #122 required the `nodeSecurityTags` be provided in the `NodeGroup` class, in practice the security group is being created in the `Cluster` class before the `NodeGroup` class, therefore any tags the user provides in `NodeGroup` aren't even being considered.

The fix should be that `nodeSecurityTags` is an option in only the `Cluster` class and not in the `NodeGroup` class, where if the `Cluster` has `nodeSecurityGroupTags` provided, it will apply this to the `nodeSecurityGroup` it creates, and then passes along to the `NodeGroup` class.

If a user wants to set tags for the `nodeSecurityGroup` they're providing in the `NodeGroup`, they must do so at creation time of the secgroup, and before being used in the `NodeGroup` creation.

metral:
Fixes #125.
Related #122.

Update for the new tag opts (based on orig in #122):
- New options per class:
  - `Cluster` class:
    - `.tags` - A map of tags that is applied to all AWS resources under management by a `pulumi/eks` cluster. Currently applied to: `eksClusterSecurityGroup`, `nodeSecurityGroup`, the Worker AutoScalingGroup (ASG), and cloudFormation stack.
    - `.clusterSecurityGroupTags` - A map of tags applied to the EKS Cluster security group
    - `.nodeSecurityGroupTags` - A map of tags applied to the Worker Nodes security group
  - `NodeGroup` class:
    - `.cloudFormationTags` - A map of tags applied to the CloudFormation Stack of Worker Nodes
    - `.autoScalingGroupTags` - A map of tags applied to the Worker NodeGroup ASG & Instances
    - Note:
      - If you create & pass in an external `nodeSecurityGroup`, you can tag it by manually setting the tags during its external creation, and before it is used in the NodeGroup.

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -121,7 +121,7 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
         vpcId: vpcId,
         revokeRulesOnDelete: true,
         tags: <aws.Tags>{
-            "Name": `${name}`,
+            "Name": `${name}-eksClusterSecurityGroup`,
             ...args.tags,
             ...args.clusterSecurityGroupTags,
         },
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -79,6 +79,8 @@ export interface CoreData {
     kubeconfig?: pulumi.Output<any>;
     vpcCni?: VpcCni;
     tags?: { [key: string]: string };
+    nodeSecurityGroup?: aws.ec2.SecurityGroup;
+    nodeSecurityGroupTags?: { [key: string]: string };
 }
 
 export function createCore(name: string, args: ClusterOptions, parent: pulumi.ComponentResource): CoreData {
@@ -285,6 +287,7 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
         instanceProfile: instanceProfile,
         eksNodeAccess: eksNodeAccess,
         tags: args.tags,
+        nodeSecurityGroupTags: args.nodeSecurityGroupTags,
     };
 }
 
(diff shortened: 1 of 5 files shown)


nodejs/eks/cluster.ts
```diff
@@ -122,7 +122,7 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
         ingress: [],
         egress: [],
         tags: <aws.Tags>{
-            "Name": `${name}`,
+            "Name": `${name}-eksClusterSecurityGroup`,
```
metral: If we omit the postfix from the Name tag, both secgroups get tagged with the same name (clusterName), and it is the Name tag that is used for the display Name in the AWS Console; thus, requiring the user to look at the secgroup `Group Name` to distinguish the two, since their names are identical.

nodejs/eks/securitygroup.ts
```diff
@@ -44,7 +44,7 @@ export function createNodeGroupSecurityGroup(name: string, args: NodeGroupSecuri
         ingress: [],
         egress: [],
         tags: args.eksCluster.name.apply(n => <aws.Tags>{
-            "Name": `${name}`,
+            "Name": `${name}-nodeSecurityGroup`,
```
metral: If we omit the postfix from the Name tag, both secgroups get tagged with the same name (clusterName), and it is the Name tag that is used for the display Name in the AWS Console; thus, requiring the user to look at the secgroup `Group Name` to distinguish the two, since their names are identical.

nodejs/eks/nodegroup.ts
```diff
@@ -60,6 +60,10 @@ export interface NodeGroupBaseOptions {
 
     /**
      * The security group to use for all nodes in this worker node group.
+     *
+     * Note: If you want to tag this `nodeSecurityGroup`, you must manually
+     * tag it during its external creation, before it is used in this node
+     * group.
```
lukehoban: Perhaps phrase this as `nodeSecurityGroup` and `nodeSecurityGroupTags` are mutually exclusive?  We could also test that at most one is set and error.

metral: @lukehoban both the option comment, and the mutual exclusion check have been addressed. PTAL

PR #128, opened 2019-05-04
fix(secgroups): do not null out ingress & egress

### LINKED ISSUE #127: Defaulting the secgroups in-line rules to null creates empty secgroups on a cluster replacement
When repro'ing https://github.com/pulumi/pulumi-eks/issues/69#issuecomment-485060221 and successfully completing cluster replacement after adding more subnets, destroying the cluster results in the same errors as #69 for **all** of the rules belonging to and including the `nodeSecurityGroup`.

The cause is tied to defaulting the ingress and egress rules to `[]` during the update that replaces the cluster, which never ends up recreating the secgroup rules. Then on the destroy, when the stand-alone secgroup rules resources are attempted to be deleted, because the rules were already removed / null'd out from the secgroup, we're returned the `rule does not exist in the security group` errors.

The null default for ingress & egress was intro'd in #109 as a requirement to be able to successfully update the cluster from using in-line secgroup rules to standalone rules.

metral:
Fixes #127 

- Note: This PR reverses the default null values used for the
  ingress and egress in-line rules of the secgroups, introduced in `v0.18.3`.
  The null default was required to move to standalone secgroup rules.
- Upgrade Path - This is a breaking change **unless** you do the following steps:
  - If using >= `v0.18.3`: update using the typical package update path.
  - If using <= `v0.18.2`:
    1. First, update your cluster from using your current version to `v0.18.4`.
    1. Next, update your cluster from `v0.18.4` to `v0.18.5` (or higher) using the typical package update path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 
 ### Improvements
 - fix(secgroups): use standalone secgroup rules instead of in-line rules
-  [#108](https://github.com/pulumi/pulumi-eks/pull/108). Note, because we are
+  [#109](https://github.com/pulumi/pulumi-eks/pull/109). Note, because we are
   replacing existing in-line secgroup rules with standalone rules,
   there may be a brief period of outage where the security group rules are
   removed before they get added back. This update happens in a matter of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ### Improvements
 
+- fix(secgroups): do not null out ingress & egress
+  [#128](https://github.com/pulumi/pulumi-eks/pull/128)
+    - Note: This PR reverses the default null values used for the
+      ingress and egress in-line rules of the secgroups, introduced in `v0.18.3`.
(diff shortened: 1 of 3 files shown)


naineel:
Just hit this myself

CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ### Improvements
 
+- fix(secgroups): do not null out ingress & egress
+  [#128](https://github.com/pulumi/pulumi-eks/pull/128)
+    - Note: This PR reverses the default null values used for the
+      ingress and egress in-line rules of the secgroups, introduced in `v0.18.3`.
+      The null default was required to move to standalone secgroup rules, but it
+      has introduced [issues](https://github.com/pulumi/pulumi-eks/issues/127), and thus is being removed in this PR.
+    - Upgrade Path - This is a **breaking change** depending on the version used:
```
lukehoban: Perhaps note that this is a breaking change **unless** you follow these steps?

metral: Done.

Issue #124, opened 2019-05-03
tests: Add better test coverage for nodejs SDK


metral:
Currently we have the following nodejs [examples/tests](https://github.com/pulumi/pulumi-eks/tree/master/nodejs/eks/examples):

We should also be testing commits against the following use-cases:
- [x] **replace-secgroup:**
  - Stand up a cluster with an external `nodeSecurityGroup` secgroup to a new `NodeGroup`, on an update remove the external secgroup to fall back on the cluster's default `nodeSecurityGroup`, and verify if the patch works as expected with the cluster remaining in a running state.
  - Update: Being added in #130 (M23)
- [x] **add-more-subnets:**
  - Stand up a cluster, and then swap out its config (e.g. go from using 2 subnets -> 3 subnets), and issue another update to see if the patch works as expected with the cluster remaining in a running state.
  - Update: Being added in #130, but not enabled yet due to dep on #107 
  - Update (06/06/2019): Being enabled in https://github.com/pulumi/pulumi-eks/pull/150 (M24)
- [x] **upgrade-from-latest:**
  - Stand up a cluster using the latest, known release available. Then apply and test the patch on another update to the existing cluster to see if the patch works as expected with the cluster remaining in a running state.
  - Being tracked in https://github.com/pulumi/pulumi-eks/pull/241 (M27)

Note: To verify the k8s cluster "remains in a running state," a small scoped smoke-test testing the following would suffice:
  - The `aws-auth` k8s ConfigMap can be retrieved, is correct, and can be correctly used by Nodes to join the cluster with no interruptions.
  - All existing k8s nodes remain RUNNING
  - All existing k8s pods remain RUNNING

metral:
Closing as this has been complete for a while, and our smoke tests validate all pods scheduled are in the RUNNING or SUCCEEDED state.

PR #130, opened 2019-05-14
Expand nodejs SDK tests coverage, and add Kubernetes Smoke Tests for examples & tests

metral:
**Note: It is suggested to review this PR by commits instead of files, as they were structured to ease reading.**

This PR adds better test coverage for the nodejs SDK, and a set of general utilities to perform Kubernetes Smoke Tests to validate that clusters are in fact up, running, and ready for use - coverage we were lacking to date, and at times creating false positives of fixes in the `master` branch.

Ref: https://github.com/pulumi/pulumi-eks/issues/124.
- `replace-secgroup` has been added.
- `add-more-subnets` is [pending](https://github.com/pulumi/pulumi-eks/issues/107) a fix, but its test is included in this PR for review, even though it is not being currently exec'd on test runs. This will be enabled in a future PR once the issue is resolved.
- `upgrade-from-latest` is being left for another PR.

Check out the [sample test output](https://gist.github.com/metral/fb64128761aa9781392b9817248dcebf) or the [Travis build output](https://travis-ci.com/pulumi/pulumi-eks/builds/111833608#L1725-L1745) showcasing the new k8s smoke tests against `examples` and the new `tests` dir.


utils/utils.go
```diff
@@ -0,0 +1,356 @@
...
+			fmt.Printf("Waiting for '%s' ConfigMap to be returned. Retrying...\n", configMapName)
+			retries--
+			time.Sleep(RetryInterval * time.Second)
+			continue
+		}
+		awsAuthReady = true
+		break
+	}
+
+	require.Equal(t, true, awsAuthReady, "EKS ConfigMap '%s' does not exist in namespace '%s'", configMapName, namespace)
```
lblackstone: nit: You can use `%q` to get a quoted string in the output.

utils/utils.go
```diff
@@ -0,0 +1,356 @@
...
+	require.NotNil(t, awsAuth.Data, "%s ConfigMap should not be nil", configMapName)
+	require.NotEmpty(t, awsAuth.Data, "'%s' ConfigMap should not be empty", configMapName)
+	fmt.Printf("EKS ConfigMap '%s' exists and has data\n", configMapName)
```
lblackstone: Why do you need to `Printf` and `Logf` the same message?

utils/utils.go
```diff
@@ -0,0 +1,356 @@
...
+		node := nodes.Items[i]
+		n, err := clientset.CoreV1().Nodes().Get(node.Name, metav1.GetOptions{})
+		if err != nil {
+			fmt.Printf("Waiting for Node '%s' to be returned. Retrying...\n", node.Name)
+			retries--
+			time.Sleep(RetryInterval * time.Second)
+			continue
+		}
+		for _, condition := range n.Status.Conditions {
+			if condition.Type == "Ready" {
```
lblackstone: Prefer using the defined condition consts here: https://github.com/kubernetes/api/blob/db5a9d1c40ebe3cdff47aab1b7ecdcd22f841ef8/core/v1/types.go#L2412-L2424

utils/utils.go
```diff
@@ -0,0 +1,356 @@
...
+				nodeReady = condition.Status == "True"
+			}
+		}
+		if nodeReady {
+			readyCount++
+			t.Logf("Node: %s | Ready Status: %t\n", node.Name, nodeReady)
+			i++ // reset vars for next run
```
lblackstone: Could you split this logic into a checker function? This loops is pretty dense and hard to follow.

utils/utils.go
```diff
@@ -0,0 +1,356 @@
...
+			retries = MaxRetries
+		} else {
+			fmt.Printf("Waiting for Node '%s' to be ready. Retrying...\n", node.Name)
+			time.Sleep(RetryInterval * time.Second)
+		}
+	}
+	require.Equal(t, readyCount, len(nodes.Items),
+		"%d out of %d Nodes are ready", readyCount, len(nodes.Items))
+
+	t.Logf("%d out of %d Nodes are ready", readyCount, len(nodes.Items))
```
lblackstone: Why a log and print here?

utils/utils.go
```diff
@@ -0,0 +1,356 @@
...
+	for i < len(pods.Items) && retries > 0 {
+		pod := pods.Items[i]
+		p, err := clientset.CoreV1().Pods(pod.Namespace).Get(pod.Name, metav1.GetOptions{})
+		if err != nil {
+			retries--
+			fmt.Printf("Waiting for Pod '%s' to be returned. Retrying...\n", pod.Name)
+			time.Sleep(RetryInterval * time.Second)
+			continue
+		}
+		if p.Status.Phase == "Running" || p.Status.Phase == "Succeeded" {
```
lblackstone: You can use consts from the same file I linked earlier.

utils/utils.go
```diff
@@ -0,0 +1,356 @@
...
+			for _, condition := range p.Status.Conditions {
+				if condition.Type == "Ready" {
+					podReady = condition.Status == "True"
+				}
+			}
+		}
+		if podReady {
+			readyCount++
+			t.Logf("Pod: %s | Ready Status: %t\n", pod.Name, podReady)
+			i++ // reset vars for next run
```
lblackstone: It would be nice to split the checking logic out of this loop if possible.

nodejs/eks/examples/examples_test.go
```diff
@@ -47,6 +49,30 @@ func Test_Examples(t *testing.T) {
 				"@pulumi/eks",
 			},
 			ExpectRefreshChanges: true,
+			ExtraRuntimeValidation: func(t *testing.T, info integration.RuntimeValidationStackInfo) {
```
lblackstone: This looks like the same copy-pasted function several times. Can you define the function once and just reference it here?

utils/utils.go
metral: Done.

utils/utils.go
metral: tl;dr: `t.Logf` in Go [only outputs](https://github.com/golang/go/issues/24929#issuecomment-384484654) success/failure after tests have completed, or failed - not live.

The `Printf's` are a means of showing transient output as the tests are running live, given that we're spinning up many different types of clusters, and sometimes its better to know more detail upfront about what's happening right or wrong, versus waiting 45+ min for builds to finish or crash to get the output.

This is more of a personal "quality of life" change, than anything else.

utils/utils.go
metral: See above: https://github.com/pulumi/pulumi-eks/pull/130#discussion_r284517932

utils/utils.go
```diff
@@ -0,0 +1,394 @@
...
+// EKSSmokeTest runs a checklist of operational successes required to deem the
+// EKS cluster as successfully running and ready for use.
+func EKSSmokeTest(t *testing.T, clientset *kubernetes.Clientset, desiredNodeCount int) {
+	// Print API Server info.
+	APIServerVersionInfo(t, clientset)
+
+	// Run all tests.
+	AssertEKSConfigMapReady(t, clientset)
+	AssertAllNodesReady(t, clientset, desiredNodeCount)
+	AssertAllPodsReady(t, clientset)
```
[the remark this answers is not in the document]
hausdorff: Ah regrouped with @metral offline -- I was suggesting that these validations should go inside the `@pulumi/eks` package itself, vs. only in the tests. This would give the EKS library a sustained competitive advantage over alternatives, since ~no other systems allow this kind of post-validation.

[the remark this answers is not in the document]
metral: I'll plan to incorporate these new EKS validations within `pulumi/eks` for post-deployment validation in a future PR.

nodejs/eks/examples/examples_test.go
metral: Done

utils/utils.go
metral: Done

utils/utils.go
metral: Done

utils/utils.go
metral: Done

utils/utils.go
metral: Done

metral:
@lblackstone Feedback has been addressed. PTAL

nodejs/eks/examples/tests/replace-cluster-add-subnets/index.ts
```diff
@@ -0,0 +1,90 @@
...
+import * as eks from "@pulumi/eks";
+
+const projectName = pulumi.getProject();
+const zones = aws.getAvailabilityZones();
+const tags = { "Name": `${projectName}` };
+
+/**
+ * VPC
+ */
+const vpcCidr = "10.0.0.0/16";
```
lukehoban: FWIW - this test could probably be massively simplified by using awsx.ec2.Vpc. That might help make it clearer what’s changing between runs, and what’s the essential point of the test.

lukehoban: Even if you don’t love to that, it might be worth splitting this into two files, so that the file you replace in the second stendoesnt have to copy/paste quite so much.

nodejs/eks/examples/tests/replace-secgroup/index.ts
```diff
@@ -0,0 +1,56 @@
...
+}, { parent: testCluster });
+
+// Create the second node group using a spot price instance.
+const spot = new eks.NodeGroup(`${projectName}-ng-tags-spot`, {
+    cluster: testCluster,
+    instanceType: "t2.medium",
+    desiredCapacity: 2,
+    nodeSecurityGroup: nodeSecurityGroup,
+    clusterIngressRule: eksClusterIngressRule,
+    minSize: 1,
```
lukehoban: There seems to be a lot of incidental complexity in this test. Is it worth pairing down a bit to focus cleanly on the test case?

utils/utils.go
```diff
@@ -0,0 +1,397 @@
...
+// MaxRetries is the maximum number of retries that a resource will be
+// attempted to be fetched from the Kubernetes API Server.
+const MaxRetries = 20
+
+// RetryInterval is the number of seconds to sleep in between requests
+// to the Kubernetes API Server.
+const RetryInterval = 15
+
+// RunEKSSmokeTest instantiates the EKS Smoke Test.
+func RunEKSSmokeTest(t *testing.T, resources []apitype.ResourceV3, kubeconfigs ...interface{}) {
```
lukehoban: Somewhat of a nit - but seems like nearly everything here should be private?  Would make clearer what the api of this module is.

utils/utils.go
```diff
@@ -0,0 +1,397 @@
...
+		clientset := kubeAccess[clusterName].Clientset
+		EKSSmokeTest(t, clientset, clusterNodeCount[clusterName])
+	}
+}
+
+// EKSSmokeTest runs a checklist of operational successes required to deem the
+// EKS cluster as successfully running and ready for use.
+func EKSSmokeTest(t *testing.T, clientset *kubernetes.Clientset, desiredNodeCount int) {
+	// Print API Server info.
+	APIServerVersionInfo(t, clientset)
```
lukehoban: If the function was naked `printXYZ`, I think we could do without the comment.

utils/utils.go
```diff
@@ -0,0 +1,397 @@
...
+
+	// Run all tests.
```
lukehoban: I’ll admit I had assumed we would do something a little more course grained - like just deploying a workload into the cluster and validating that it works (including can allocate LB, can allocate storage, can respond correctly to network traffic, etc). 

That may be an additional thing on top of what’s here.

metral: You're assumption is valid, and doing workload testing post cluster readiness is definitely something I too wanted to add - this PR just took longer than expected to test the readiness piece, so I did not get around to it. I've gone ahead and added it to the test coverage todo being tracked in https://github.com/pulumi/pulumi-eks/issues/124.

utils/utils.go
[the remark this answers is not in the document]
metral: Integration of the validations post-deployment is being tracked in https://github.com/pulumi/pulumi-eks/issues/124.

utils/utils.go
metral: Great points - the idea was to make this as generic and modular as possible, to be able to leverage the validation code here for other managed k8s providers in the future, hence the many public / exported methods.

I'll clean this up and only expose the proper methods suitable to this pkg API

nodejs/eks/examples/tests/replace-cluster-add-subnets/index.ts
metral: Great points. I'll make sure to simplify this by using `awsx` instead of raw `aws`

metral: Done

nodejs/eks/examples/tests/replace-secgroup/index.ts
metral: Done

utils/utils.go
metral: Done

PR #131, opened 2019-05-15
Update link to aws-iam-authenticator install documents

ellismg:
The old link was pointing at a page AWS had updated which did not have
a section on this step. This link goes directly to the page about how
to install `aws-iam-authenticator`

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,7 @@ or `yarn`:
 
 
 This package requires `aws-iam-authenticator` for Amazon EKS in order to deploy resources to EKS clusters. Install this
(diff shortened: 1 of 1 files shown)


lukehoban:
Aside: Interestingly, AWS just released a replacement for `aws-iam-authenticator` yesterday.  Not clear whether a dependency on `aws` CLI is "better", but directionally I suspect we may want to move to that to reduce dependence on a somewhat awkward `aws-iam-authenticator` dependency.

https://aws.amazon.com/about-aws/whats-new/2019/05/amazon-eks-simplifies-kubernetes-cluster-authentication/

Issue #135, opened 2019-05-17
Bump aws-k8s-cni to v1.5.0 when available


metral:
A [fix](https://github.com/aws/amazon-vpc-cni-k8s/pull/461) that aims to improve ENI Pod IP + CNI management, and failure + retry logic is slated for v1.5.0. Both seem like welcomed additions when up against general service timeouts.

naveensrinivasan:
We would be interested in this. When is it likely to go in. Thanks

metral:
We're currently waiting for it to be released: https://github.com/aws/amazon-vpc-cni-k8s/releases

ellismg:
The 1.5.0 release happened earlier today.

PR #148, opened 2019-06-06
update(aws-k8s-cni): move from 1.4.1 -> 1.5.0

metral:
Based on: https://git.io/fjztn

Closes https://github.com/pulumi/pulumi-eks/issues/135


metral:
This was tested on an existing cluster using v0.18.6 and PASSED.

PR #154, opened 2019-06-07
Add support for endpoint access control and custom cluster securitygroup

### LINKED ISSUE #86: Add support for cluster endpoint access control
I know the feature is recent, but I would like to set the cluster endpoint access control as shown here: https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html

I may be wrong, but I don't see the option currently

lukehoban:
Expose these options through to the `aws.eks.Cluster`.

Fixes #86.

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -119,28 +119,33 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
     }, { parent: parent });
 
     // Create the EKS cluster security group
-    const eksClusterSecurityGroup = new aws.ec2.SecurityGroup(`${name}-eksClusterSecurityGroup`, {
-        vpcId: vpcId,
-        revokeRulesOnDelete: true,
-        tags: pulumi.all([
-            args.tags,
-            args.clusterSecurityGroupTags,
-        ]).apply(([tags, clusterSecurityGroupTags]) => (<aws.Tags>{
-            "Name": `${name}-eksClusterSecurityGroup`,
-            ...tags,
-            ...clusterSecurityGroupTags,
-        })),
-    }, { parent: parent });
+    let eksClusterSecurityGroup: aws.ec2.SecurityGroup;
+    if (args.clusterSecurityGroup) {
+        eksClusterSecurityGroup = args.clusterSecurityGroup;
+    } else {
+        eksClusterSecurityGroup = new aws.ec2.SecurityGroup(`${name}-eksClusterSecurityGroup`, {
+            vpcId: vpcId,
+            revokeRulesOnDelete: true,
+            tags: pulumi.all([
+                args.tags,
+                args.clusterSecurityGroupTags,
+            ]).apply(([tags, clusterSecurityGroupTags]) => (<aws.Tags>{
+                "Name": `${name}-eksClusterSecurityGroup`,
+                ...tags,
+                ...clusterSecurityGroupTags,
+            })),
+        }, { parent: parent });
 
-    const eksClusterInternetEgressRule = new aws.ec2.SecurityGroupRule(`${name}-eksClusterInternetEgressRule`, {
-        description: "Allow internet access.",
-        type: "egress",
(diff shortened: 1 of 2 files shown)


nodejs/eks/cluster.ts
```diff
@@ -495,6 +500,16 @@ export interface ClusterOptions {
      * By default it is off.
      */
     enabledClusterLogTypes?: pulumi.Input<pulumi.Input<string>[]>;
+
+    /**
+     * Indicates whether or not the Amazon EKS public API server endpoint is enabled. Default is `true`.
+     */
+    endpointPublicAccess?: boolean;
```
metral: This default is `false` not `true

metral: Both vars are being passed in as `false` if not set

metral:
Per https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html#modify-endpoint-access, when `endpointPrivateAccess` = `true` and `endpointPublicAccess` = `false` we have no internet access to the cluster. This means that to test this in tests requires a VM in the same VPC issuing the commands to the APIserver - either via `kubectl` or `client-go`

lukehoban:
> This means that to test this in tests requires a VM in the same VPC issuing the commands to the APIserver - either via kubectl or client-go

It's actually even worse than that - we won't even be able to successfully deploy resources into the Kubernetes cluster from the machine running `pulumi up` as needed to complete initialization - so the deployment will fail.  I was just about to reply with some options on this on https://github.com/pulumi/pulumi-eks/issues/86#issuecomment-499637790.

lukehoban:
> Also per the defaults here of undefined for both if not provided - this does not match the state of what EKS defaults to (public: true, private: false) - do you foresee any issues with this mismatch?

It passes through the underlying defaults - we'll pass undefined (not present) which will trigger the underlying default value.  Right?

metral:
> It passes through the underlying defaults - we'll pass undefined (not present) which will trigger the underlying default value. Right?

Ya you're correct - I initially skipped over the `undefined` more than I should've and that's what caused me confusion. Ignore this comment 👍

metral:
From https://github.com/pulumi/pulumi-eks/issues/86#issuecomment-500077791:

> Ensure that users can provide a custom Cluster security group so that they can configure ingress with the CIDR blocks, IPs or other SGs they want to have access (and document that deployments will only be possible if done from an IP within one of these).

It'd be really great to see new examples added to the `examples` directory of privateAccess clusters with an accompanying VM in the same the VPC, for each of the config scenarios of the custom secgroup you list out (CIDR, IP, and other SGs).

lukehoban:
> It'd be really great to see new examples added to the examples directory of privateAccess clusters with an accompanying VM in the same the VPC

Yeah - I'll add a test before merging this.  It's more involved that I initially expected unfortunately - it'll require setting up VPN endpoints inside the AWS VPC and creating a VPN connection from the test machine (or Travis).  That's going to add non-trivial complexity to the test environment.

lukehoban:
@metral After spending some time thinking about this - it's actually going to be exceedingly hard to test private-only endpoint access - as it will require VPN'ing the test machine into the VPC - or running the `pulumi` deployment from a bastion host in the VPC - neither of which is going to be easy to do in our current test/CI setup.

Given that this is mostly just passing flags through to EKS, I'm inclined to merge this to unblock all the usage scenarios here (including private+public) - and to open a follow-up issue to add test coverage (and examples) of private-only endpoint access.

Thoughts?

metral:
> Given that this is mostly just passing flags through to EKS, I'm inclined to merge this to unblock all the usage scenarios here (including private+public) - and to open a follow-up issue to add test coverage (and examples) of private-only endpoint access.

That SGTM @lukehoban. I feel like the follow-up issue for test coverage & examples applies to not only `pulumi/eks`, but generally speaking, `@pulumi/pulumi` also for testing updates & flow from a bastion + VPC + private subnets client or CI setup.

PR #159, opened 2019-06-12
Add serviceRole and instanceProfile to ClusterOptions

VRanga000:
Addresses issue #159

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -85,6 +85,19 @@ export interface CoreData {
     nodeSecurityGroupTags?: InputTags;
 }
 
+function createOrGetInstanceProfile(parent: pulumi.ComponentResource, instanceRoleName?: pulumi.Input<aws.iam.Role>, instanceProfileName?: pulumi.Input<string>): aws.iam.InstanceProfile {
+  let instanceProfile: aws.iam.InstanceProfile;
+  if (instanceProfileName) {
+    instanceProfile = aws.iam.InstanceProfile.get(`${name}-instanceProfile`, instanceProfileName, undefined, { parent: parent });
+  } else {
+    instanceProfile = new aws.iam.InstanceProfile(`${name}-instanceProfile`, {
+        role: instanceRoleName,
+    }, { parent: parent });
+  }
+
+  return instanceProfile;
+}
+
 export function createCore(name: string, args: ClusterOptions, parent: pulumi.ComponentResource): CoreData {
     // Check to ensure that aws-iam-authenticator is installed, as we'll need it in order to deploy k8s resources
     // to the EKS cluster.
@@ -231,13 +244,7 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
     } else if (args.instanceRole) {
         // Create an instance profile if using a default node group
         if (!args.skipDefaultNodeGroup) {
-            if (args.instanceProfileName) {
-              instanceProfile = aws.iam.InstanceProfile.get(`${name}-instanceProfile`, args.instanceProfileName, undefined, { parent: parent });
-            } else {
-              instanceProfile = new aws.iam.InstanceProfile(`${name}-instanceProfile`, {
-                  role: args.instanceRole,
-              }, { parent: parent });
-            }
+            instanceProfile = createOrGetInstanceProfile(parent, args.instanceRole, args.instanceProfileName);
         }
 
         instanceRoleMappings = pulumi.output(args.instanceRole).apply(instanceRole =>
@@ -253,6 +260,7 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
                 "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly",
             ],
         }, { parent: parent })).role;
+
         instanceRoles = pulumi.output([instanceRole]);
 
         // Create a new policy for the role, if specified.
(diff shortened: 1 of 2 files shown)


nodejs/eks/cluster.ts
```diff
@@ -371,6 +380,17 @@ export interface ClusterOptions {
...
     instanceRole?: pulumi.Input<aws.iam.Role>;
 
+    /**
+     * Node instance profile - for worker nodes. If not supplied, it will be created.
+     */
+    instanceProfileName?: string;
+
+    /**
+     * Service role for eks cluster
+     *
```
metral:
```suggestion
     *IAM Service Role for EKS to use to manage the cluster.
```

nodejs/eks/cluster.ts
```diff
@@ -371,6 +380,17 @@ export interface ClusterOptions {
      */
     instanceRole?: pulumi.Input<aws.iam.Role>;
 
+    /**
+     * Node instance profile - for worker nodes. If not supplied, it will be created.
+     */
```
metral:
```suggestion
     */The default IAM InstanceProfile to use on the Worker NodeGroups, if one is not already set in the NodeGroup.
```

nodejs/eks/cluster.ts
```diff
@@ -109,14 +109,19 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
+    } else {
+      eksRole = (new ServiceRole(`${name}-eksRole`, {
         service: "eks.amazonaws.com",
         description: "Allows EKS to manage clusters on your behalf.",
         managedPolicyArns: [
             "arn:aws:iam::aws:policy/AmazonEKSClusterPolicy",
             "arn:aws:iam::aws:policy/AmazonEKSServicePolicy",
         ],
-    }, { parent: parent });
+      }, { parent: parent })).role;
```
metral: I appreciate the changes for brevity in accessing and only using the `.role` here and below in the lambda for getting its ARN, but I wonder if for future cases there may be a need to export the details of the entire `ServiceRole` versus just extracting the role here and tossing out the rest of the object props.

nodejs/eks/cluster.ts
metral: replace both lines w/ the suggestion - I guess you can only make suggested changes to a single line instead of two.

nodejs/eks/cluster.ts
VRanga000: Thank you for all the feedback, I've accepted your other suggestions. 

Regarding this one: the reason for doing this was because the changes would pervade deeper if I were to create a ServiceRole in both logical paths (role is injected / role is created). We would need to modify the servicerole class to allow the role to be injected. I can proceed along this route if you think this is preferable.

nodejs/eks/cluster.ts
```diff
@@ -216,9 +221,13 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
     } else if (args.instanceRole) {
         // Create an instance profile if using a default node group
         if (!args.skipDefaultNodeGroup) {
-            instanceProfile = new aws.iam.InstanceProfile(`${name}-instanceProfile`, {
-                role: args.instanceRole,
-            }, { parent: parent });
+            if (args.instanceProfileName) {
+              instanceProfile = aws.iam.InstanceProfile.get(args.instanceProfileName, args.instanceProfileName);
```
pgavlin: I would use `${name}-instanceProfile` for the name here to allow `args.instanceProfileName` to be a `pulumi.Input<string>`. i.e.

```suggestion
              instanceProfile = aws.iam.InstanceProfile.get(`${name}-instanceProfile`, args.instanceProfileName, undefined, { parent: parent });
```

nodejs/eks/cluster.ts
```diff
@@ -371,6 +380,16 @@ export interface ClusterOptions {
      */
     instanceRole?: pulumi.Input<aws.iam.Role>;
 
+    /**
+     * The default IAM InstanceProfile to use on the Worker NodeGroups, if one is not already set in the NodeGroup.
+     */
+     instanceProfileName?: string;
```
pgavlin:
```suggestion
     instanceProfileName?: pulumi.Input<string>;
```

nodejs/eks/cluster.ts
```diff
@@ -216,9 +221,13 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
     } else if (args.instanceRole) {
```
lukehoban: There is a third branch of this `if/else if/else if` that also creates an `InstanceProfile`.  I assume we need to update that one as well?

The branching here is a little wild, and probably eventually should get simplified down or refactored a bit - but that can be part of a a follow-up PR.

VRanga000: You are right, currently the instanceprofile only seems to get created if a single iam role is passed in among the args. Presumably an instanceprofile (or profiles?) need to get get created in the "if" branch as well? In that branch, a list of roles are passed in. The only use case we currently have is when a single role is passed in, so I didn't pay attention to the other logical branches :)

VRanga000:
Thank you guys for the review / comments. Is anything else needed from me at this point to merge this PR?

nodejs/eks/cluster.ts
lukehoban: I think we do need to make the change in the other branch as well to be correct here.  We are allowing the user to pass an explicit `InstanceRole` they want to use, but then in some cases will ignore it and create a new one anyway.

VRanga000:
> LGTM. Resolve the conflicts and you should be all set. Thanks for the PR!
> 
> @lukehoban any thoughts before merging?

Thank you, conflicts resolved!

lukehoban:
Just bubbling this up - since I think my note got hidden in a resolved thread:

https://github.com/pulumi/pulumi-eks/pull/159#discussion_r298192371

> I think we do need to make the change in the other branch as well to be correct here. We are allowing the user to pass an explicit InstanceRole they want to use, but then in some cases will ignore it and create a new one anyway.

nodejs/eks/cluster.ts
VRanga000: Ok.. Ill address this shortly. Thanks.

VRanga000: just pushed a commit for this.

Issue #170, opened 2019-06-18
userMappings and roleMappings don't allow omitting optional data


MerelyAPseudonym:
In particular:
* The default `username` is the ARN of the IAM user or role, but the interface declarations don’t allow omitting the `username` to take advantage of the default behavior:
  https://github.com/pulumi/pulumi-eks/blob/f5fb36c53c504f3f05fb5ba260d76225a7e9a033/nodejs/eks/cluster.ts#L38-L41
  https://github.com/pulumi/pulumi-eks/blob/f5fb36c53c504f3f05fb5ba260d76225a7e9a033/nodejs/eks/cluster.ts#L58-L61
* As far as AWS is concerned, it's not necessary to include _any_ groups in a mapping, but the type declarations require a group list regardless:
  https://github.com/pulumi/pulumi-eks/blob/f5fb36c53c504f3f05fb5ba260d76225a7e9a033/nodejs/eks/cluster.ts#L43-L46
  https://github.com/pulumi/pulumi-eks/blob/f5fb36c53c504f3f05fb5ba260d76225a7e9a033/nodejs/eks/cluster.ts#L63-L66

MerelyAPseudonym:
> As far as AWS is concerned, it's not necessary to include any groups in a mapping
I'm actually not 100% sure it's acceptable to omit `groups` entirely, although `groups: []` seems to be fine.

MerelyAPseudonym:
Well, I was wrong about what was causing the problem in our cluster. Even though the AWS documentation [_clearly_ states](https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html) that “by default, the user name is the ARN of the IAM user”, removing the `username` entry from an `aws-auth` user-mapping subsequently causes `kubectl get pods` to fail with:

    Error from server (Forbidden): pods is forbidden: User "" cannot list resource "pods" in API group "" in the namespace "default"

To be honest, now I’m not sure if `@pulumi/eks` should be doing anything differently.

lukehoban:
@MerelyAPseudonym Based on the latest reply here - it sounds like there is no action that is needed here for this package.  I'm going to close out, but if there is something you still think this package should be doing differently here - let us know.

PR #171, opened 2019-06-18
Allow omitting optional {user,role}Mapping data

MerelyAPseudonym:
Fixes https://github.com/pulumi/pulumi-eks/issues/170


MerelyAPseudonym:
I see (and appreciate!) the requirement to update the changelog. I'll handle that shortly.

MerelyAPseudonym:
Actually, I've run into some confusing results in my end. I'm going to close this PR while I'm still testing.

Issue #141, opened 2019-05-31
Multiple default storageClasses when specifying storageClass in EKS cluster


Brianbrifri:
When creating an EKS cluster and specifying a custom storage class (e.g. `st1`), the default `gp2` storageClass is created along side the custom `st1` storageClass which is also marked as default. 

Where `storageClass = "st1"`:
```// Create a VPC for our cluster.
const vpc = new awsx.ec2.Vpc("Kafka-VPC", {numberOfAvailabilityZones: 3});

// Create an EKS cluster with the given configuration.
const cluster = new eks.Cluster("kafka-cluster", {
    vpcId: vpc.id,
    subnetIds: vpc.privateSubnetIds,
    instanceType: instanceType,
    desiredCapacity: desiredCapacity,
    minSize: minSize,
    maxSize: maxSize,
    storageClasses: storageClass,
    deployDashboard: deployDashboard,
});
```

This results in:
```> kubectl get storageclasses
NAME                                   PROVISIONER             AGE
gp2 (default)                          kubernetes.io/aws-ebs   23m
kafka-cluster-st1-evr2qvmh (default)   kubernetes.io/aws-ebs   21m
```

This can cause issues when helm charts rely on one default storage class since the storageClass name cannot be passed directly to the charts. 

Expected behavior: One storageClass is marked as default after EKS creation (preferably the user specified class)
Actual behavior: Both the EKS created `gp2` **AND** the user specified `st1` storage class are marked as default.

Other thoughts:
Ability to add/remove default status to storageClasses

metral:
By default, the `gp2` storage class is created for us [automatically](https://docs.aws.amazon.com/eks/latest/userguide/storage-classes.html) by the EKS service for clusters with version >= 1.11. This is not created by Pulumi. 

To get the desired result, you'll have to either patch the `gp2` storage class to not be default after the cluster is up using kubectl, but preferably your helm values.yaml should take a storage class param to explicitly use the intended class as to get around EKS' defaults - for example, [here](https://github.com/helm/charts/tree/master/stable/efs-provisioner#configuring) is how the AWS EFS helm chart specifies the desired class it needs.

Brianbrifri:
Is there a way of querying the storage class names from the pulumi/kubernetes api? That's the only way I know of to get the actual name of the user provided storage class to be able to pass it to the helm charts.

Brianbrifri:
I know #139 has just been opened but until then, I'm hoping there is a workaround.

[...]

Issue #139, opened 2019-05-29
Expose storage class(es) from cluster


clstokes:
Storage classes may be used by other resources and the eks package doesn't currently allow for predictable naming (it uses the auto-naming functionality) nor does it expose the storage classes that are created.

The EKS package should store the resulting storage class on the `Cluster` so that downstream dependencies can use it.

PR #172, opened 2019-06-19
feat(storage-classes): export all user created storage classes

metral:
- Closes https://github.com/pulumi/pulumi-eks/issues/139 feat request.
- Fixes https://github.com/pulumi/pulumi-eks/issues/141 by exporting `StorageClasses` names that can be used, for example, to pass into Helm charts.

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -26,7 +26,7 @@ import { createNodeGroup, NodeGroup, NodeGroupBaseOptions, NodeGroupData } from 
 import { createNodeGroupSecurityGroup } from "./securitygroup";
 import { ServiceRole } from "./servicerole";
 import { createStorageClass, EBSVolumeType, StorageClass } from "./storageclass";
-import { InputTags } from "./utils";
+import { InputTags, UserStorageClasses } from "./utils";
 
 /**
  * RoleMapping describes a mapping from an AWS IAM role to a Kubernetes user and groups.
@@ -80,6 +80,7 @@ export interface CoreData {
     instanceRoles: pulumi.Output<aws.iam.Role[]>;
     instanceProfile?: aws.iam.InstanceProfile;
     eksNodeAccess?: k8s.core.v1.ConfigMap;
+    storageClasses?: UserStorageClasses;
     kubeconfig?: pulumi.Output<any>;
     vpcCni?: VpcCni;
     tags?: InputTags;
@@ -228,12 +229,14 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
 
     // Add any requested StorageClasses.
     const storageClasses = args.storageClasses || {};
+    const userStorageClasses: UserStorageClasses = pulumi.output({});
     if (typeof storageClasses === "string") {
         const storageClass = { type: storageClasses, default: true };
-        createStorageClass(`${name.toLowerCase()}-${storageClasses}`, storageClass, { parent: parent, provider: provider });
+        userStorageClasses[storageClasses] = pulumi.output(
+            createStorageClass(`${name.toLowerCase()}-${storageClasses}`, storageClass, { parent: parent, provider: provider }));
     } else {
         for (const key of Object.keys(storageClasses)) {
-            createStorageClass(`${name.toLowerCase()}-${key}`, storageClasses[key], { parent: parent, provider: provider });
+            userStorageClasses[key] = pulumi.output(createStorageClass(`${name.toLowerCase()}-${key}`, storageClasses[key], { parent: parent, provider: provider }));
         }
     }
 
@@ -334,6 +337,7 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
         eksNodeAccess: eksNodeAccess,
         tags: args.tags,
         nodeSecurityGroupTags: args.nodeSecurityGroupTags,
+        storageClasses: userStorageClasses,
     };
 }
 
@@ -489,9 +493,11 @@ export interface ClusterOptions {
 
     /**
      * An optional set of StorageClasses to enable for the cluster. If this is a single volume type rather than a map,
-     * a single StorageClass will be created for that volume type and made the cluster's default StorageClass.
+     * a single StorageClass will be created for that volume type.
      *
-     * Defaults to "gp2".
+     * Note: As of Kubernetes v1.11+ on EKS, a default `gp2` storage class will
+     * always be created automatically for the cluster by the EKS service. See
+     * https://docs.aws.amazon.com/eks/latest/userguide/storage-classes.html
      */
     storageClasses?: { [name: string]: StorageClass } | EBSVolumeType;
 
--- a/nodejs/eks/examples/examples_test.go
+++ b/nodejs/eks/examples/examples_test.go
@@ -106,6 +106,23 @@ func Test_Examples(t *testing.T) {
 				)
 			},
 		},
+		{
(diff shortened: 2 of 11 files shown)


nodejs/eks/examples/storage-classes/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+export function checkStorageClass(storageClasses: eks.UserStorageClasses) {
+    if (storageClasses) {
+        pulumi.output(storageClasses).apply(scs => {
+            for (const key of Object.keys(scs)) {
+                const sc = scs[key];
+                sc.metadata.apply(m => {
+                    // 'defaultName' used per suggestion in https://git.io/fjVYi due to
+                    // https://github.com/pulumi/pulumi/issues/2433
+                    const name = m.name || "defaultName";
+                    if (!name) {
```
lblackstone: Wouldn't this always evaluate to `true`? If `m.name` is falsy, I think `defaultName` would be assigned.

metral: This was done per the suggestion in #2 of https://git.io/fjVYi due to the [issue](https://github.com/pulumi/pulumi/issues/2433) linked in the comment above it that returns an `undefined` on the first `pulumi preview`. During a preview this will show up as `defaultName`, but after `pulumi update --skip-preview` we get the right storageClass names:

ex. on `preview`:
image

but then on the `update`:
image

lblackstone: I realized I had that backwards. I don't think the conditional block will ever execute since the `||` guarantees that a truthy value will be set.

metral:
@lukehoban @pgavlin PTAL

nodejs/eks/cluster.ts
```diff
@@ -192,12 +193,13 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
 
     // Add any requested StorageClasses.
     const storageClasses = args.storageClasses || {};
+    const userStorageClasses: UserStorageClasses = {};
     if (typeof storageClasses === "string") {
         const storageClass = { type: storageClasses, default: true };
-        createStorageClass(`${name.toLowerCase()}-${storageClasses}`, storageClass, { parent: parent, provider: provider });
+        userStorageClasses[`${storageClasses}`] = createStorageClass(`${name.toLowerCase()}-${storageClasses}`, storageClass, { parent: parent, provider: provider });
```
lukehoban: This is equivalent to `[storageClasses]` - can we use that simpler form?

nodejs/eks/examples/examples_test.go
```diff
@@ -107,6 +107,28 @@ func Test_Examples(t *testing.T) {
...
+		{
+			Dir: path.Join(cwd, "storage-classes"),
+			Config: map[string]string{
+				"aws:region": region,
+			},
+			Dependencies: []string{
+				"@pulumi/eks",
+			},
+			// Per suggestion in https://git.io/fjVYi due to
+			// https://github.com/pulumi/pulumi/issues/2433
```
lukehoban: Perhaps use our standard `TODO` format to make this follow up more greppable?

nodejs/eks/examples/storage-classes/utils.ts
lukehoban: Yes - @lblackstone is right - this conditional looks like it can never be true and this test will never fail.

nodejs/eks/utils.ts
```diff
@@ -21,3 +22,14 @@ import * as pulumi from "@pulumi/pulumi";
...
+/**
+ * UserStorageClasses represents a map of an Input type to a
+ * k8s StorageClass to capture all of the *user-created* storage classes. This
+ * map does not include the default `gp2` automatically added by EKS.
+ *
+ * Note: As of Kubernetes v1.11+ on EKS, a default `gp2` storage class will
+ * always be created automatically for the cluster by the EKS service. See
+ * https://docs.aws.amazon.com/eks/latest/userguide/storage-classes.html
+ */
+export type UserStorageClasses = pulumi.Input<{ [key: string]: k8s.storage.v1.StorageClass }>;
```
lukehoban: Since this only appears in output positions, I assume it would be more useful to make it an `Output`?

nodejs/eks/cluster.ts

nodejs/eks/examples/examples_test.go
metral: Done

nodejs/eks/utils.ts
metral: Done

metral:

metral:

nodejs/eks/examples/storage-classes/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+    storageClasses.apply(scs => {
+        for (const key of Object.keys(scs)) {
+            const sc = scs[key];
+            sc.metadata.apply(m => {
+                if (m.name) {
+                    // TODO: @metral remove this when issue is fixed.
+                    // 'defaultName' is used per suggestion in https://git.io/fjVYi due to
+                    // https://github.com/pulumi/pulumi/issues/2433
+                    const name = m.name || "defaultName";
+                    if (!name || name === "defaultName") {
```
lukehoban: The logic here looks very odd. We’re inside an `if m.name`, then we default a value in (which can never happen). Then we check if it’s still undefined which doubly can’t ever happen. 

It feels like this is not accomplishing what you intend, or at least is an awkward way to accomplish it if it is?

nodejs/eks/cluster.ts
```diff
@@ -192,12 +193,17 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
 
     // Add any requested StorageClasses.
     const storageClasses = args.storageClasses || {};
+    const userStorageClasses: UserStorageClasses = pulumi.output({});
     if (typeof storageClasses === "string") {
         const storageClass = { type: storageClasses, default: true };
-        createStorageClass(`${name.toLowerCase()}-${storageClasses}`, storageClass, { parent: parent, provider: provider });
+        userStorageClasses.apply(usc => {
+            usc[storageClasses] = createStorageClass(`${name.toLowerCase()}-${storageClasses}`, storageClass, { parent: parent, provider: provider });
+        });
```
lukehoban: I would not populate these asynchronously. There could be very subtle races here. 

The suggestions to make the output an Output was to wrap in `Pulumi.output` before storing into the return value to this function.

nodejs/eks/examples/storage-classes/utils.ts
metral: Done

metral:
Feedback has been addressed, and code has been simplified. PTAL @lukehoban

nodejs/eks/examples/storage-classes/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    },
+});
+
+if (cluster2.core.storageClasses) {
+    // Use many storage classes.
+    if ("mygp2" in cluster2.core.storageClasses) {
+        cluster2.core.storageClasses["mygp2"].apply(sc => {
+            sc.metadata.name.apply(n => {
+                if (n) {
+                    console.log(n);
```
lukehoban: Curious - what is the goal of these `console.log`s?  They aren't verified are they?

metral: It's simply to use for output of the generated storage class names. Figured printing them to the console worked well for an example. It wasn't meant as a verification other than the name is not `undefined`.

PR #175, opened 2019-06-19
Default to a node AMI that matches the cluster version

### LINKED ISSUE #173: Identify a better means of selecting smart defaults for versions & AMIs used for both control plane & workers to create supported, working clusters.
See [thread](https://github.com/aws/containers-roadmap/issues/231#issuecomment-503643262).

Current [filtering](https://github.com/pulumi/pulumi-eks/blob/master/nodejs/eks/nodegroup.ts#L371-L411) done in the `NodeGroup` class for smart defaults.

Current control plane versions supported on EKS as of June 6, 2019:

Current worker node AMIs returned, sorted in ascending order by `CreationDate`: https://gist.github.com/metral/0c6c7c436d53a056d726fc79fbf988b3

lukehoban:
When the user provides neither a specific Node AMI or a specific Node version, default to using the same version as the cluster control plane.

Fixes #173.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Improvements
 
+- Default to a node AMI that matches the cluster version
+  [#175](https://github.com/pulumi/pulumi-eks/pull/175)
 - fix(tags): rm ASG tag dupes, and consider tag inheritance for all tags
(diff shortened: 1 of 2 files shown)


metral:
Update the changelog, and this should be good to go!

PR #181, opened 2019-06-25
Allow more granular configuration of kubelet

### LINKED ISSUE #180: Expose kubelet extra args for every node group
This is required by us to enable some sysctl commands which are considered unsafe.
https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#enabling-unsafe-sysctls
The only way to do this right now is through kubelet-extra-args flag on EKS which is already used to add labels and taints to the nodes in this module. 
This option should be exposed for the ability to add more commands as required by users.
https://github.com/pulumi/pulumi-eks/blob/master/nodejs/eks/nodegroup.ts#L321

lukehoban:
Allow passing through additional arguments to `/etc/eks/bootstrap.sh` and `--kubelet-extra-args`.

Fixes #180.

### DIFF
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -130,6 +130,22 @@ export interface NodeGroupBaseOptions {
     taints?: { [key: string]: Taint };
 
     /**
+     * Extra args to pass to the Kubelet.  Corresponds to the options passed in the `--kubeletExtraArgs` flag to
+     * `/etc/eks/bootstrap.sh`.  An array of flags should be provided, such as: ['--port=10251', '--address=0.0.0.0'].
+     * Note that the `labels` and `taints` properties will be applied to this list after to the expicit
+     * `kubeletExtraArgs`.
+     */
+    kubeletExtraArgs?: string[];
+
+    /**
+     * Additional args to pass directly to `/etc/eks/bootstrap.sh`.  Fror details on available options, see:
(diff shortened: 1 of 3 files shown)


nodejs/eks/nodegroup.ts
```diff
@@ -129,6 +129,22 @@ export interface NodeGroupBaseOptions {
...
+     */
+    kubeletExtraArgs?: string[];
+
+    /**
+     * Additional args to pass directly to `/etc/eks/bootstrap.sh`.  Fror details on available options, see:
+     * https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh.  Note that the `--apiserver-endpoint`,
+     * `--b64-cluster-ca` and `--kubelet-extra-args` flags are included automatically based on other configuration
+     * parameters.
+     */
+    bootstrapExtraArgs?: string;
```
metral: Is it worth making this an array instead of a string to collect the flags? Seems like an array would also make the string generation simpler below using a `.join`

lukehoban: I think I'll actually do the opposite and make both of these just `string`.  They are meant to just pass through the raw additional args to the command line, and any interpretation of the args as an array runs some risk of being both less clear what exactly it will do, as well as being subtly incompatible with some argument scheme the underlying command expects.

metral:
The cleanup looks great!

Issue #140, opened 2019-03-10
EKS cluster creation does not handle assumed roles


igorshapiro:
Hi, trying to create an EKS cluster using a VPC (awsx) from the same stack. Here's the code:

```js
module.exports = function(vpc) {
  const allSubnetIds = pulumi.all([vpc.getSubnetIds('private'), vpc.getSubnetIds('public')])
    .apply(([private, public]) => private.concat(public))
  return new eks.Cluster('eks_cluster', {
    deployDashboard: true,
    desiredCapacity: 1,
    instanceType: aws.ec2.M5InstanceLarge,
    maxSize: 2,
    nodeRootVolumeSize: 100,
    subnetIds: allSubnetIds,
    vpcId: vpc.vpc.id
  })
}
```

Getting the following errors:

```
     Type                                  Name                 Status                  Info
     pulumi:pulumi:Stack                   account-dev          **failed**              2 errors; 16 messages
     └─ eks:index:Cluster                  eks_cluster                                  
 +      └─ pulumi-nodejs:dynamic:Resource  eks_cluster-vpc-cni  **creating failed**     1 error
 
Diagnostics:
  pulumi:pulumi:Stack (account-dev):
    Method handler checkConfig for /pulumirpc.ResourceProvider/CheckConfig expected but not provided
    Method handler diffConfig for /pulumirpc.ResourceProvider/DiffConfig expected but not provided
    unable to recognize "/var/folders/28/z072xy4d5970vkd6l3vb_r780000gn/T/tmp-32396RlquTqkLiKdX.tmp": Unauthorized
    unable to recognize "/var/folders/28/z072xy4d5970vkd6l3vb_r780000gn/T/tmp-32396RlquTqkLiKdX.tmp": Unauthorized
    unable to recognize "/var/folders/28/z072xy4d5970vkd6l3vb_r780000gn/T/tmp-32396RlquTqkLiKdX.tmp": Unauthorized
    unable to recognize "/var/folders/28/z072xy4d5970vkd6l3vb_r780000gn/T/tmp-32396RlquTqkLiKdX.tmp": Unauthorized
    unable to recognize "/var/folders/28/z072xy4d5970vkd6l3vb_r780000gn/T/tmp-32396RlquTqkLiKdX.tmp": Unauthorized
 
    (node:32400) ExperimentalWarning: queueMicrotask() is experimental.
    Error: invocation of aws:ec2/getSubnet:getSubnet returned an error: grpc: the client connection is closing
[...]

PR #184, opened 2019-06-27
add optional arg to create EKS cluster with IAM role entity

mazamats:
Addresses issue: https://github.com/pulumi/pulumi-eks/issues/140

This fixes the issue of multiple AWS IAM users not being able to manage an EKS cluster within the same stack, this is due to the way EKS initially allows authentication only to the IAM entity which created the cluster (User or Role).

Creating a role and using that for the AWS provider which creates the EKS cluster makes the initial entity an IAM role that others can assume to mange the Kube resources that the Cluster deploys (RBAC ConfigMap, Dashboard, CNI Daemonset)

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -170,38 +170,40 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
                 }`,
             ),
             description: `Admin access to eks-${name}`,
-          },
+          }, { parent: this },
         );
+
         eksCreatorEntityARN = eksCreatorEntity.arn.apply(arn => arn);
+
         eksCreatorEntityPolicy = new aws.iam.Policy(`${name}-eksClusterCreatorPolicy`, {
           description: "EKS Cluster Creator Policy",
-          policy: eksRole.role.apply(role => {
-                return {
-                    Version: "2012-10-17",
-                    Statement: [
-                      {
-                        Effect: "Allow",
-                        Action: "eks:*",
-                        Resource: "*",
-                      },
-                      {
-                        Effect: "Allow",
-                        Action: "iam:PassRole",
-                        Resource: role.arn,
-                      },
-                    ],
-              };
-          }).apply(JSON.stringify),
-        });
+          policy: {
+            Version: "2012-10-17",
+            Statement: [
+              {
+                Effect: "Allow",
+                Action: "eks:*",
+                Resource: "*",
+              },
+              {
+                Effect: "Allow",
+                Action: "iam:PassRole",
+                Resource: "*",
+              },
+            ],
+          },
+        }, { parent: eksCreatorEntity });
+
         eksCreatorEntityRolePolicyAttachment = new aws.iam.RolePolicyAttachment(`${name}`, {
             policyArn: eksCreatorEntityPolicy.arn.apply(arn => arn),
             role: eksCreatorEntity,
-        });
+        }, { parent: this });
+
         eksCreatorEntityProvider = new aws.Provider(`${name}-eksClusterCreatorEntity`, {
             assumeRole: {
                 roleArn: eksCreatorEntity.arn,
             },
-        }, { dependsOn: eksCreatorEntityRolePolicyAttachment });
+        }, { parent: this, dependsOn: eksCreatorEntityRolePolicyAttachment });
     }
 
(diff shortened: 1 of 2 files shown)


nodejs/eks/cluster.ts
```diff
@@ -323,6 +382,16 @@ function createInstanceRoleMapping(arn: pulumi.Input<string>): RoleMapping {
...
 }
 
+function generateAuthenticatorArgs(clusterName: string, roleArn?: string): string[] {
+    let args = ["token", "-i", clusterName];
+
+    if (roleArn) {
+        args = [ ...args, "-r", roleArn];
+    }
+
+    return args;
```
metral: We can simplify this further with:

```
if (roleArn) {
  return [ ...args, "-r", roleArn];
}
return ["token", "-i", clusterName];
```

nodejs/eks/cluster.ts
```diff
@@ -147,6 +147,65 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
     }
 
+    // Create an AWS IAM Entity that will be the IAM entity which is automatically granted system:masters permissions in the cluster's RBAC configuration
+    let eksCreatorEntity: aws.iam.Role | undefined;
+    let eksCreatorEntityARN: pulumi.Input<aws.ARN> | undefined;
+    let eksCreatorEntityPolicy: aws.iam.Policy;
+    let eksCreatorEntityRolePolicyAttachment: aws.iam.RolePolicyAttachment;
+    let eksCreatorEntityProvider: aws.Provider | undefined;
+
+    if (args.useRoleForClusterCreation) {
```
metral: This looks good - can we capture this block in a new func like `createIAMProvider` or some such to simplify legibility of the class?

nodejs/eks/cluster.ts
```diff
@@ -147,6 +147,65 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
+        eksCreatorEntity = new aws.iam.Role(`${name}-eksClusterCreatorRole`, {
```
lukehoban: I assume we should be storing this `Role` as an output property of the component so that users can get it and make changes if needed.

Should the be an option to just provide this `Role` directly instead of asking to have it be created?

metral: Agreed, this feels like something that should be provided in an option. We tend to default to letting users create these themselves and give them to us in other areas of the cluster config.

mazamats: I think making it default and allowing users to pass in the provider or the provider-role would be the simplest way to go about it.

- Passing in only the role would create an aws provider
- Passing in both the provider and role would cause the class to create neither
- Passing in neither role nor provider would create both role and provider
- Passing in only a provider and no role would throw an error since we need the ARN of the Role to create the kubeconfig

Thoughts?

mazamats:
> Is this the right fix here? It somehow doesn't feel like what I would have expected to need to do.
> 
> A couple questions:
> 
> 1. If this works, why can't the user create their own Role and provider for that role and then pass that provider through to `new eks.Cluster`?  Is it really "right" to create just this one resource with a unique provider/role?
> 2. If this is right, should it be the default?

I agree with you, it feels dirty to create a single provider for the sole purpose of creating a single EKS cluster resource. Ideally Amazon would allow us to override the ARN of the initial IAM entity that receives `system:master` permission when creating an EKS cluster instead of inferring it.

metral:
Hi @mazamats - do you have any plans to update this PR with the [feedback](https://github.com/pulumi/pulumi-eks/pull/184#pullrequestreview-256591710) provided?

mazamats:
> Hi @mazamats - do you have any plans to update this PR with the [feedback](https://github.com/pulumi/pulumi-eks/pull/184#pullrequestreview-256591710) provided?

Yes, I'm hoping to get some info on how to work around [this](https://github.com/pulumi/pulumi-aws/issues/673) issue first since it's needed for this PR.

metral:
> Yes, I'm hoping to get some info on how to work around [this](https://github.com/pulumi/pulumi-aws/issues/673) issue first since it's needed for this PR.

@lukehoban @pgavlin any ideas? ^

mazamats:
Unable to commit to this PR, continuing here: https://github.com/pulumi/pulumi-eks/pull/205

PR #189, opened 2019-07-01
fix(cluster): fix name ref error when using instanceProfile

metral:
Fixes #188

PR #190, opened 2019-07-01
Revert master to last working commit

metral:


metral:
That's right. We currently don't run the full CI test suite on community PRs and rely on pulumi staff to test them but we forgot, and an [error](https://github.com/pulumi/pulumi-eks/pull/189) was not caught.

More so, we're actively trying to fix our CI that's been acting up so we wanted to revert to a known working state to help unblock the pipeline as we fix it. Once we do, we'll get your commit back in along with any changes needed.

Thanks for your understanding

VRanga000:
Ok, makes sense. Thanks for the explanation. Let me know if there's anything I could do to help.

PR #191, opened 2019-07-03
Issue # 159: Add serviceRole and instanceProfile to ClusterOptions

### LINKED ISSUE #158: Allow service role and instance profile to be injected during cluster creation
Currently, the iam instance profile (attached to the instance role)  and the service role (used by master nodes / control plane) are created during cluster creation. We have a use case for the roles to be passed in (injected) rather than created - because the user performing the cluster creation does not have permissions to create roles (corporate policy), and roles are managed externally. 

I can create a PR with the implementation for the above changes.

VRanga000:
A redo of PR #158 - with a bug fix in the last commit.

Fixes #158

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -85,6 +85,19 @@ export interface CoreData {
     nodeSecurityGroupTags?: InputTags;
 }
 
+function createOrGetInstanceProfile(parent: pulumi.ComponentResource, instanceRoleName?: pulumi.Input<aws.iam.Role>, instanceProfileName?: pulumi.Input<string>): aws.iam.InstanceProfile {
+  let instanceProfile: aws.iam.InstanceProfile;
+  if (instanceProfileName) {
+    instanceProfile = aws.iam.InstanceProfile.get(`${name}-instanceProfile`, instanceProfileName, undefined, { parent: parent });
+  } else {
+    instanceProfile = new aws.iam.InstanceProfile(`${name}-instanceProfile`, {
+        role: instanceRoleName,
+    }, { parent: parent });
+  }
+
+  return instanceProfile;
+}
+
 export function createCore(name: string, args: ClusterOptions, parent: pulumi.ComponentResource): CoreData {
     // Check to ensure that aws-iam-authenticator is installed, as we'll need it in order to deploy k8s resources
     // to the EKS cluster.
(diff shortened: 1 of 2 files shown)


lukehoban:
@VRanga000 Were you able to run tests yourself on this?  It also hit this error when it was merged to master previously - and I don't see any commits that would address this?

https://travis-ci.com/pulumi/pulumi-eks/builds/117576992#L1423

```
[ i-eks/nodejs/eks/examples/tags ]   aws:iam:InstanceProfile (example-tags-cluster1-instanceProfile):
[ i-eks/nodejs/eks/examples/tags ]     error: Plan apply failed: Either `role` or `roles` (deprecated) must be specified when creating an IAM Instance Profile
```

VRanga000:
My commit fixed a different (glaring) bug .. 
and I was skipping tests apparently: 
`--- SKIP: Test_Examples (0.00s)
    examples_test.go:33: Skipping test due to missing AWS_REGION environment variable
--- SKIP: Test_AllTests (0.00s)
    all_test.go:33: Skipping test due to missing AWS_REGION environment variable
`

When I run with my aws credentials, I see failures / warnings like these:
`
Failed to upload test command report to S3: InvalidAccessKeyId: The AWS Access Key Id you provided does not exist in our records.
        status code: 403, request id: 8AC62CC8B1316F61, host id: 79MQDOTRXalVcd2R0TnDpe5IcbLeit7yTgM45828SP9kSUckqzHzLpJ7Cy+s8SD9f1cT20tYXWI=
[ /nodejs/eks/examples/nodegroup ] Wrote output to /tmp/p-it-vikas-dell-nodegroup-321b0e36-229095234/command-output/yarn-install.20190702-193303.d42f9.log
`
I will try and address this bug by inspection..

VRanga000:
Hm as far as I can tell, I am passing in the role into InstanceProfile creation : 
`
 93     instanceProfile = new aws.iam.InstanceProfile(`${name}-instanceProfile`, {
 94         role: instanceRoleName,
 95     }, { parent: parent });
`

The other path, where its merely looking up an existing instance profile, should not require a role (and there is no role param in your library as expected, either.
`
     instanceProfile = aws.iam.InstanceProfile.get(`${name}-instanceProfile`, instanceProfileName, undefined, { parent: parent });

`
So Im confused as to why / where its expecting a role to be passed in. Could the "get" method be erroneously doing a "create" downstream? Ill take a look at that angle as well.

nodejs/eks/cluster.ts
```diff
@@ -85,10 +85,11 @@ export interface CoreData {
     nodeSecurityGroupTags?: InputTags;
 }
 
-function createOrGetInstanceProfile(name: string, parent: pulumi.ComponentResource, instanceRoleName?: pulumi.Input<aws.iam.Role>, instanceProfileName?: pulumi.Input<string>): aws.iam.InstanceProfile {
+export function createOrGetInstanceProfile(name: string, parent: pulumi.ComponentResource, instanceRoleName?: pulumi.Input<aws.iam.Role>, instanceProfileName?: pulumi.Input<string>): aws.iam.InstanceProfile {
```
lukehoban: I don’t think we want to export this, it’s just an internal helper - right?

VRanga000: maybe not.. which is why I did not do that at first. But it did not show up in the generated js file - I guess its optimized away.. which was kinda annoying. Your call.

VRanga000:
The last commit should have fixed the regression (build/tests passed in ci). It was caused by the instanceRole param to the InstanceProfile creation being erroneously passed the wrong value.

nodejs/eks/cluster.ts
VRanga000: removed export

metral:
@lukehoban any further thoughts on this PR?

@VRanga000 we'll need to resolve the conflicts before moving forward - can you please rebase?

lukehoban:
Sorry for the delay - this looks good to me.  @metral I don't see any merge conflicts, but if anything is needed do you want to take care of that and get this merged in?

Issue #199, opened 2019-07-23
Empty preview step on tests is proposing changes that are not expected


metral:
After the initial update of the [`replace-cluster-add-subnets`](https://github.com/pulumi/pulumi-eks/blob/master/nodejs/eks/examples/tests/replace-cluster-add-subnets/index.ts) test, the empty preview step is being shown changes even though it should not:

- See the `templateBody` of the ASG being [planned](https://gist.github.com/metral/8c756d45f90ceebeffde23860dc7c53d#file-log-txt-L1730) for an update.
- See the empty preview step erroring as changes are being [proposed](https://gist.github.com/metral/8c756d45f90ceebeffde23860dc7c53d#file-log-txt-L2208) when not expected.

e.g. here is a sample of the changes being proposed from a run of another test, the [`migrate-nodegroups`](https://github.com/pulumi/pulumi-eks/tree/master/nodejs/eks/examples/tests/migrate-nodegroups/index.ts), also hitting this change proposal for `templateBody`:

![image](https://user-images.githubusercontent.com/1112768/62639251-05c77700-b8f4-11e9-8e61-7e593867cdbc.png)

metral:
https://s3-us-west-2.amazonaws.com/eng.pulumi.com/travis-logs/pulumi/pulumi-eks/834.1/failed-tests.tar.gz

metral:
Another run of CI hit this same issue again, showing a proposed change for the `templateBody` of the CFStack, when the code was not touched: https://gist.github.com/metral/60aff5b74d7cafa980cee56403853151#file-log-txt-L1502

metral:
Seems related:
- https://github.com/pulumi/pulumi-aws/issues/684#issuecomment-519199905
- https://github.com/pulumi/pulumi-terraform/pull/444

metral:
Full output logs `-v=10`: [command-output.tar.gz](https://github.com/pulumi/pulumi-eks/files/3479183/command-output.tar.gz).

Note the changes in `VPCZoneIdentifier` from 4 subnets to 2 in the planned Diff of ` command-output/pulumi-preview-empty.20190807-145341.daef9.log`:

```
[...]

PR #200, opened 2019-07-23
fix(test): sort subnets before using them in replace-cluster-add-subnets

metral:
Fixes https://github.com/pulumi/pulumi-eks/issues/199


metral:
This turned out to be a dud. Need to revert the commit to remove it, as it proved unnecessary.

Issue #149, opened 2019-06-06
VpcCni state includes absolute path to YAML, leading to unnecessary brittleness and opaque changes


MerelyAPseudonym:
I can run `pulumi preview` on one machine and see that there will be no changes, but if I move the project to a different directory and run `pulumi preview` again, I see that Pulumi wants to update a `pulumi-nodejs:dynamic:Resource` but it can’t explain any more about that change. By inspecting the output of `pulumi preview --json`, I can see that the only changes are in the (serialized) content of `__provider`, and the only changes there are in the paths to the CNI YAML:

```diff
--- provider1.js	2019-06-06 16:22:34.771169181 -0400
+++ provider2.js	2019-06-06 16:22:32.715455152 -0400
@@ -80,7 +80,7 @@
 
 function __f3() {
     return (function() {
-        with({ applyVpcCniYaml: __applyVpcCniYaml, yamlPath: "/tmp/demo1/node_modules/@pulumi/eks/cni/aws-k8s-cni.yaml", crypto: require("crypto") }) {
+        with({ applyVpcCniYaml: __applyVpcCniYaml, yamlPath: "/tmp/demo2/node_modules/@pulumi/eks/cni/aws-k8s-cni.yaml", crypto: require("crypto") }) {
 
             return (inputs) => {
                 applyVpcCniYaml(yamlPath, inputs);
@@ -93,7 +93,7 @@
 
 function __f4() {
     return (function() {
-        with({ applyVpcCniYaml: __applyVpcCniYaml, yamlPath: "/tmp/demo1/node_modules/@pulumi/eks/cni/aws-k8s-cni.yaml" }) {
+        with({ applyVpcCniYaml: __applyVpcCniYaml, yamlPath: "/tmp/demo2/node_modules/@pulumi/eks/cni/aws-k8s-cni.yaml" }) {
 
             return (id, state, inputs) => {
                 applyVpcCniYaml(yamlPath, inputs);
```

Tying the Pulumi state to file paths on a deployment machine does not seem desirable or necessary.

MerelyAPseudonym:
Also, note that the explanation from Pulumi is very sparse:

```sh-session
$ pulumi preview
```

I wouldn’t be surprised if it

metral:
I believe I've run into something similar to this. I'll try to repro on my end.

> I wouldn’t be surprised if it’s difficult to be more informative with dynamic resources, though.

[...]

PR #201, opened 2019-07-24
fix(cni): modify CNI filepath to store the relative path only in state

metral:
Fixes https://github.com/pulumi/pulumi-eks/issues/149

This fix gets around the absolute path being inserted into the state for
the dynamic provider, by only storing the relative path for the CNI YAML
file.

The absolute path has now been pushed out to be formed and resolved in
`computeVpcCniYaml()` before the file is opened, instead of the
depending on the full path from the `applyVpcCniYaml()` caller.

### DIFF
--- a/nodejs/eks/cni.ts
+++ b/nodejs/eks/cni.ts
@@ -78,6 +78,9 @@ interface VpcCniInputs {
 function computeVpcCniYaml(yamlPath: string, args: VpcCniInputs): string {
     // Read the CNI YAML. The original source for this YAML is
     // https://github.com/aws/amazon-vpc-cni-k8s/tree/master/config.
+
+    // Form the absolute CNI YAML file path.
+    yamlPath = path.resolve("./", yamlPath);
     const cniYamlText = fs.readFileSync(yamlPath).toString();
     const cniYaml = jsyaml.safeLoadAll(cniYamlText);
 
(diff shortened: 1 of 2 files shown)


nodejs/eks/cni.ts
```diff
@@ -131,7 +134,10 @@ export class VpcCni extends pulumi.dynamic.Resource {
             throw new Error("Could not set VPC CNI options: kubectl is missing. See https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-kubectl for installation instructions.");
         }
 
-        const yamlPath = path.join(__dirname, "cni", "aws-k8s-cni.yaml");
+        // Form the relative CNI YAML file path.
+        const projectPath = __dirname.substring(0, __dirname.indexOf("node_modules"));
+        const relativeClassDir = __dirname.replace(projectPath, "");
+        const yamlPath = path.join(relativeClassDir, "cni/aws-k8s-cni.yaml");
```
lukehoban: Would it have worked instead to just move `const yamlPath = path.join(__dirname, "cni", "aws-k8s-cni.yaml");` inside each of the functions below?

I'll admit I don't quite follow the three lines above, and hardcoding `node_modules` here feels "wrong" though I'm sure it works okay for normal cases.

metral: Moving the `yamlPath` definition into each function did not work. It does not find the file:
`error: Plan apply failed: ENOENT: no such file or directory, open 'cni/aws-k8s-cni.yaml`

> I'll admit I don't quite follow the three lines above, and hardcoding node_modules here feels "wrong"

This was the only way I found to get around embedding an absolute path into the state of the dynamic provider, which seems to be the root issue. Do you have any other suggestions?

lukehoban: Fair enough - if this solves the problem then let’s go ahead with it!

Issue #204, opened 2019-07-25
IAM role times out after 30sec which fails CI jobs


metral:
Every now and then we get IAM eventual consistency issues that seem to show up in different ways - this one is related to IAM roles timing out after 30sec have passed for it to be created.

I do not have any logs at the moment, but I'll will add on its next reoccurrence. 

See:
- https://pulumi.slack.com/archives/CDNTZPPFT/p1560987586069100
- https://pulumi.slack.com/archives/CDNTZPPFT/p1560988423070400

Related:
- https://github.com/pulumi/pulumi-eks/issues/203

metral:
Closing as this has not occurred in months

Issue #203, opened 2019-07-25
creating error: Plan apply failed: error creating EKS Cluster: InvalidParameterException: Error in role params


metral:
```
aws:eks:Cluster test-replace-cluster-add-subnets-eksCluster creating error: Plan apply failed: error creating EKS Cluster
(test-replace-cluster-add-subnets-eksCluster-1480f64): InvalidParameterException: Error in role params
```

Every now and then we get IAM eventual consistency issues that seem to show up in different ways - this one is related to IAM roles erroring during creation.

Tracking this given that Travis CI hits this every now and then: [failed-tests.tar.gz](https://s3-us-west-2.amazonaws.com/eng.pulumi.com/travis-logs/pulumi/pulumi-eks/835.1/failed-tests.tar.gz).

Related:
- https://github.com/pulumi/pulumi-eks/issues/204

metral:
Recently hit this today, and here are the command-output logs with pulumi test log level set to `-v=10`:

[command-output.tar.gz](https://github.com/pulumi/pulumi-eks/files/3478876/command-output.tar.gz)

metral:
Potentially fixed https://github.com/pulumi/pulumi-aws/pull/817

metral:
Reopening, as the fix in https://github.com/terraform-providers/terraform-provider-aws/pull/11127 is for the eks provider in TF, which is not used here for `pulumi/eks`

lukehoban:
> for the eks provider in TF, which is not used here for pulumi/eks

I don’t quite follow - we do indeed use `aws.eks.Cluster` here which is the resource that was fixed.

metral:
@lukehoban You're correct, disregard my comment - I thought it had resurfaced per Cyrus' thread, but I was crossing wires here. Closing this issue.

PR #205, opened 2019-07-26
create EKS cluster & resources with IAM Role Provider

mazamats:
Addresses issue: https://github.com/pulumi/pulumi-eks/issues/140

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -69,6 +69,15 @@ export interface UserMapping {
 }
 
 /**
+ * CreationRoleProvider is a component containing the AWS Role and Provider necessary to override the `[system:master]`
+ * entity ARN. This is an optional argument used in `ClusterOptions`. Read more: https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html
+ */
+export interface CreationRoleProvider {
+    role: aws.iam.Role;
+    provider: pulumi.ProviderResource;
+}
+
+/**
  * CoreData defines the core set of data associated with an EKS cluster, including the network in which it runs.
  */
 export interface CoreData {
@@ -100,6 +109,105 @@ function createOrGetInstanceProfile(name: string, parent: pulumi.ComponentResour
   }
 
   return instanceProfile;
+}
+
+function generateKubeconfig(clusterName: string, clusterEndpoint: string, certData: string, roleArn?: string) {
+    let args = ["token", "-i", clusterName];
+
+    if (roleArn) {
+        args = [...args, "-r", roleArn];
+    }
+
+    return {
+        apiVersion: "v1",
+        clusters: [{
+            cluster: {
+                server: clusterEndpoint,
+                "certificate-authority-data": certData,
+            },
+            name: "kubernetes",
+        }],
+        contexts: [{
+            context: {
+                cluster: "kubernetes",
+                user: "aws",
+            },
+            name: "aws",
+        }],
+        "current-context": "aws",
+        kind: "Config",
+        users: [{
+            name: "aws",
+            user: {
+                exec: {
+                    apiVersion: "client.authentication.k8s.io/v1alpha1",
+                    command: "aws-iam-authenticator",
+                    args: args,
+                },
+            },
+        }],
+    };
+}
+
+export function getRoleProvider(name: string, region?: aws.Region, profile?: string): CreationRoleProvider {
+    const iamRole = new aws.iam.Role(`${name}-eksClusterCreatorRole`, {
+        assumeRolePolicy: aws.getCallerIdentity().then(id => `{
+            "Version": "2012-10-17",
+            "Statement": [
+                {
+                "Effect": "Allow",
+                "Principal": {
+                    "AWS": "arn:aws:iam::${id.accountId}:root"
+                },
+                "Action": "sts:AssumeRole"
+                }
+            ]
+            }`,
+        ),
+        description: `Admin access to eks-${name}`,
+    });
+
+    // `eks:*` is needed to create/read/update/delete the EKS cluster, `iam:PassRole` is needed to pass the EKS service role to the cluster
+    // https://docs.aws.amazon.com/eks/latest/userguide/service_IAM_role.html
+    const rolePolicy = new aws.iam.RolePolicy(`${name}-eksClusterCreatorPolicy`, {
+        role: iamRole,
+        policy: {
+          Version: "2012-10-17",
+          Statement: [
+            {
+              Effect: "Allow",
+              Action: "eks:*",
+              Resource: "*",
+            },
+            {
+              Effect: "Allow",
+              Action: "iam:PassRole",
+              Resource: "*",
+            },
+          ],
+        },
+      },
+      { parent: iamRole },
+    );
+
+    const provider = new aws.Provider(`${name}-eksClusterCreatorEntity`, {
+        region: region,
+        profile: profile,
+        assumeRole: {
+            roleArn: iamRole.arn.apply(async(arn) => {
+                // wait 30 seconds to assume the IAM Role https://github.com/pulumi/pulumi-aws/issues/673
+                if (!pulumi.runtime.isDryRun()) {
+                    await new Promise(resolve => setTimeout(resolve, 30 * 1000));
+                }
+                return arn;
+            }),
+        },
+    }, { parent: iamRole });
+
+    return {
+        role: iamRole,
+        provider: provider,
+    };
 }
 
 export function createCore(name: string, args: ClusterOptions, parent: pulumi.ComponentResource): CoreData {
@@ -180,7 +288,10 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
         },
         version: args.version,
         enabledClusterLogTypes: args.enabledClusterLogTypes,
-    }, { parent: parent });
+    }, {
+        parent: parent,
+        provider: args.creationRoleProvider ? args.creationRoleProvider.provider : undefined,
+    });
 
     // Instead of using the kubeconfig directly, we also add a wait of up to 5 minutes or until we
     // can reach the API server for the Output that provides access to the kubeconfig string so that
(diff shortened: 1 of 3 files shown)


nodejs/eks/cluster.ts
```diff
@@ -140,6 +184,26 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
       }, { parent: parent })).role;
     }
 
+    // Create the EKS cluster IAM Role Provider
+    let creationRole: pulumi.Output<aws.iam.Role>;
+    if (args.creationRole !== undefined) {
```
metral:
```suggestion
    if (args.creationRole) {
```

nodejs/eks/cluster.ts
```diff
@@ -140,6 +184,26 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
+        creationRole = pulumi.output(args.creationRole);
+    } else {
+        creationRole = createEntityRole(name, parent);
+    }
+    const eksCreationRoleArn = creationRole.apply(role => role.arn);
```
metral:
```suggestion
    const eksCreationRoleArn = creationRole.arn;
```

This should work

nodejs/eks/cluster.ts
```diff
@@ -435,6 +500,11 @@ export interface ClusterOptions {
      */
     serviceRole?: pulumi.Input<aws.iam.Role>;
 
+    /**
+     * The IAM Role used to create & authenticate against the EKS cluster.
```
metral: Let's add a note about how this role is granted `system:masters` permissions in k8s, and possibly link to https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html

nodejs/eks/cluster.ts
```diff
@@ -620,6 +690,11 @@ export class Cluster extends pulumi.ComponentResource {
      */
     public readonly instanceRoles: pulumi.Output<aws.iam.Role[]>;
 
+    /**
+     * The IAM Role used to create & authenticate against the EKS Cluster
```
metral: Same note as above.

nodejs/eks/cluster.ts
```diff
@@ -140,6 +184,26 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
+    } else {
+        creationRole = createEntityRole(name, parent);
+    }
+    const eksCreationRoleArn = creationRole.apply(role => role.arn);
+    const eksClusterProvider = new aws.Provider(`${name}-eksClusterCreatorEntity`, {
+        assumeRole: {
+            roleArn: eksCreationRoleArn.apply(async(arn) => {
+                // wait to assume the IAM Role https://github.com/pulumi/pulumi-aws/issues/673
+                if (!pulumi.runtime.isDryRun()) {
+                    await new Promise(resolve => setTimeout(resolve, 15 * 1000));
```
metral: I'd consider bumping this up to 30sec given that we've seen IAM timeout [issues](https://github.com/pulumi/pulumi-eks/issues/204) before.

metral:
Given that this PR is introducing a `aws.Provider` for the cluster it [waits](https://github.com/pulumi/pulumi-eks/pull/205/files#diff-0dfaec1e9f3a8568a8e2516febe83a69R200) to assume the role, I wonder if this will temper down the IAM related issues we're seeing around role creation & assuming roles in CI: https://github.com/pulumi/pulumi-eks/issues/203 & https://github.com/pulumi/pulumi-eks/issues/204.

Thoughts @lukehoban ?

nodejs/eks/cluster.ts
mazamats: That gives me `Property 'arn' does not exist on type 'OutputInstance<Role>'.`

nodejs/eks/cluster.ts
```diff
@@ -140,6 +184,26 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
       }, { parent: parent })).role;
     }
 
+    // Create the EKS cluster IAM Role Provider
+    let creationRole: pulumi.Output<aws.iam.Role>;
+    if (args.creationRole) {
+        creationRole = pulumi.output(args.creationRole);
+    } else {
+        creationRole = createEntityRole(name, parent);
+    }
```
lukehoban: Do we want to always create a custom role for creating the cluster?  Or do we want to keep defaulting to using the outer provider?

As written, I suspect this will cause all existing clusters to be replaced - which is probably not an option.

nodejs/eks/cluster.ts
```diff
@@ -102,6 +103,49 @@ function createOrGetInstanceProfile(name: string, parent: pulumi.ComponentResour
...
+            {
+              Effect: "Allow",
+              Action: "eks:*",
+              Resource: "*",
+            },
+            {
+              Effect: "Allow",
+              Action: "iam:PassRole",
+              Resource: "*",
+            },
```
lukehoban: Are these required permissions documented anywhere in AWS docs we could link to here?

nodejs/eks/cluster.ts
```diff
@@ -140,6 +184,26 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
+        assumeRole: {
+            roleArn: eksCreationRoleArn.apply(async(arn) => {
+                // wait 30 seconds to assume the IAM Role https://github.com/pulumi/pulumi-aws/issues/673
+                if (!pulumi.runtime.isDryRun()) {
+                    await new Promise(resolve => setTimeout(resolve, 30 * 1000));
+                }
+                return arn;
+            }),
+        },
+    }, { parent: parent });
```
lukehoban: I’ll admit I still don’t exactly follow why we have to create a new provider. Does EKS implicitly take the role that was used to create it and bake it in somewhere?  That would be really weird. Is there not a way to just pass it the role you want it to associate with IAM access?

lukehoban: How does this handle region and other provide configuration?  I’m not sure there is any automatic propagation of such things from the parents provider down to a newly constructed provider instance. (Perhaps there should be?)

mazamats: > Does EKS implicitly take the role that was used to create it and bake it in somewhere? That would be really weird. Is there not a way to just pass it the role you want it to associate with IAM access?

Yes, the role is inferred and there isn't a way to specify the it as a a param when creating the EKS cluster.

From the AWS [docs](https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html): `When you create an Amazon EKS cluster, the IAM entity user or role, such as a federated user that creates the cluster, is automatically granted system:masters permissions in the cluster's RBAC configuration. To grant additional AWS users or roles the ability to interact with your cluster, you must edit the aws-auth ConfigMap within Kubernetes.`

nodejs/eks/cluster.ts
mazamats: > As written, I suspect this will cause all existing clusters to be replaced - which is probably not an option.

I can see how that would cause issues for others, would adding a bool flag to toggle the custom role suffice? How would the outputs respond to that? (i.e. `this.registerOutputs` using undefined vars)

I think this should eventually be default, until AWS allows the creation of EKS clusters with the `system:masters` entity as a param instead of inferring it.

nodejs/eks/cluster.ts
mazamats: The `eks:*` is needed to create, read, update, delete the EKS cluster. The `iam:Passrole` is to give the EKS cluster the `serviceRole` that gets created in the class. https://github.com/pulumi/pulumi-eks/pull/205/files#diff-0dfaec1e9f3a8568a8e2516febe83a69R173

https://docs.aws.amazon.com/eks/latest/userguide/service_IAM_role.html

nodejs/eks/cluster.ts
mazamats: We could access region with `aws.getRegion()` and define it explicitly, assuming that will pick up the parent provider for the api call.

metral: > That would be really weird. Is there not a way to just pass it the role you want it to associate with IAM access?

I too have not seen any explicit association between the role used at cluster creation time and the EKS cluster. Per the AWS quote, this is implicitly inferred.

nodejs/eks/cluster.ts
metral: I lean towards making this role & provider optional for those who require it, and to continue allowing backwards compatibility.

mazamats:
Thanks for the feedback @metral  and @lukehoban

If nothing else comes up, I'll push an update this week to address the remaining concerns.

- Make the Role & Provider optional
- Add documentation for the Role Policy permissions

mazamats:
I found this issue on one of AWS's repos: https://github.com/aws/containers-roadmap/issues/378

If anyone with some sway could chime in to get the `[system:masters]` role defined explicitly that would awesome.

For the time being I refactored the PR to be backwards compatible and added a `getRoleProvider` and `CreationRoleProvider` in case someone wanted to use a different role/provider.

Example usage:

```node
  const creationRoleProvider = eks.getRoleProvider("test");

  const cluster = new eks.Cluster("test", {
    creationRoleProvider: creationRoleProvider,
    subnetIds: [...privateSubnets.ids, ...publicSubnets.ids],
    nodeSubnetIds: privateSubnets.ids,
    vpcId: vpc.id
  });```

File: nodejs/eks/cluster.ts [hunk 9]
```diff
@@ -68,6 +68,11 @@ export interface UserMapping {
     groups: pulumi.Input<pulumi.Input<string>[]>;
 }
 
+export interface CreationRoleProvider {
```
metral: Could we please add comments given that this is publicly exported?

nodejs/eks/cluster.ts
```diff
@@ -102,6 +107,109 @@ function createOrGetInstanceProfile(name: string, parent: pulumi.ComponentResour
...
+            }`,
+        ),
+        description: `Admin access to eks-${name}`,
+    });
+
+    /**
+     * `eks:*` is needed to create/read/update/delete the EKS cluster
+     * `iam:PassRole` is needed to pass the EKS service role to the cluster
+     *
+     * https://docs.aws.amazon.com/eks/latest/userguide/service_IAM_role.html
```
metral: Lets remove these 2 empty lines

nodejs/eks/cluster.ts
```diff
@@ -620,6 +717,14 @@ export class Cluster extends pulumi.ComponentResource {
      */
     public readonly instanceRoles: pulumi.Output<aws.iam.Role[]>;
 
+    /**
+     * The IAM Role used to create & authenticate against the EKS cluster. This role is given `[system:masters]`
+     * permission in K8S, See: https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html
+     *
+     * Note: When supplying a role, make sure that it has `eks:*` and `iam:Passrole` permissions.
```
metral:
```suggestion
     * Note: When supplying a role, make sure that it has permissions for each action, `eks:*` and `iam:Passrole`,  on `Resource: *`.
```

Consider adding a bit more detail on the requirements.

metral:
Recent changes LGTM. @lukehoban PTAL

Could we also squash down to a single commit?

Issue #194, opened 2019-07-10
DependencyViolation: resource sg has a dependent object


metral:
We're hitting an issue on EKS cluster tear downs: the security group fails (http 400) on a deletion request due to a lingering ENI created by AWS for the worker instance that does not get deleted first.

A manual work around is going into the console and deleting the ENI first, and then the sg can be deleted as expected.

The only diff between this setup and a vanilla EKS cluster is that there is a workload deployed to k8s with a classic ELB that gets all torn down together on `pulumi destroy`, but the ENI shows it belongs to the instance, so doubt its related to the LB.

Semi-related / same err message: https://github.com/terraform-providers/terraform-provider-aws/issues/1671#issuecomment-390375295.

```
[ /aws-ts-eks-migrate-nodegroups ] Diagnostics:
[ /aws-ts-eks-migrate-nodegroups ]   aws:ec2:SecurityGroup (migrate-nodegroups-nodeSecurityGroup):
[ /aws-ts-eks-migrate-nodegroups ]     error: Plan apply failed: deleting urn:pulumi:p-it-argon-aws-ts-eks-d389e1a1::migrate-nodegroups::eks:index:Cluster$aws:ec2/securityGroup:SecurityGroup::migrate-nodegroups-nodeSecurityGroup: DependencyViolation: resource sg-07d3e15caa6895678 has a dependent object
[ /aws-ts-eks-migrate-nodegroups ]      status code: 400, request id: 5c77f9c1-6b9c-4691-9556-1776c831d86
```

lukehoban:
> a lingering ENI created by AWS

Do you have details on what this ENI is?

> The only diff between this setup and a vanilla EKS cluster is that there is a workload deployed to k8s with a classic ELB that gets all torn down together on pulumi destroy, but the ENI shows it belongs to the instance, so doubt its related to the LB.

This feels suggestive that the ENI is created by Kubernetes. Are you sure this is not something that the Kubernetes load balancer support is doing?

lukehoban:
Ahh - so CNI allocates ENIs.  Likely related to that?  

https://github.com/aws/amazon-vpc-cni-k8s#eni-allocation

metral:
> Ahh - so CNI allocates ENIs on instances dynamically as pods are scheduled. Likely related to that?

[...]

PR #214, opened 2019-08-07
Fix and improve migrate-nodegroup test

metral:
Fixes https://github.com/pulumi/pulumi-eks/issues/194.

The changes in this fix resolve a leaked ENI that lead to a dependency violation on tear down. It was resolved by bumping up `aws-cni` to v1.5.2 that was recently [released](https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.5.2) [1]. We now also:
 - Wait 5 min after the migration step to ensure it completes,
 - Delete the workload Namespace, and `aws-cni` DaemonSet to ensure these get explicitly torn down.

This fix was tested across 70+ runs of the `migrate-nodegroup` test, and no run hit issue #194.

Additionally, `logLevel`, `logFile`, and `image` have been added as options to `VpcCniOptions`, and defaults for logging have been adjusted to be verbose, and be produced for the `aws-cni` Pods.

---

[1] - The specific fixes in `aws-cni` v1.5.2 were:
  - [Detach ENI before deleting.](https://github.com/aws/amazon-vpc-cni-k8s/pull/538)
  - The addition of a `healthz` endpoint: https://github.com/aws/amazon-vpc-cni-k8s/pull/548 and https://github.com/aws/amazon-vpc-cni-k8s/pull/553. The `healthz` endpoint is used in the readiness & liveness probes of the DaemonSet.

### DIFF
--- a/nodejs/eks/cni.ts
+++ b/nodejs/eks/cni.ts
@@ -64,6 +64,28 @@ export interface VpcCniOptions {
      * assignment on the node.
      */
     warmIpTarget?: pulumi.Input<number>;
+
+    /**
+     * Specifies the log level used for logs.
+     *
+     * Defaults to "DEBUG".
+     * See more options: https://git.io/fj92K
+     */
+    logLevel?: pulumi.Input<string>;
+
+    /**
+     * Specifies the file path used for logs.
+     *
+     * Defaults to "stdout" to emit Pod logs for `kubectl logs`.
+     */
+    logFile?: pulumi.Input<string>;
+
+    /**
+     * Specifies the container image to use in the AWS CNI cluster DaemonSet.
+     *
+     * Defaults to the official AWS CNI image in ECR.
+     */
+    image?: pulumi.Input<string>;
 }
 
 interface VpcCniInputs {
@@ -73,6 +95,9 @@ interface VpcCniInputs {
     externalSnat?: boolean;
     warmEniTarget?: number;
     warmIpTarget?: number;
+    logLevel?: string;
+    logFile?: string;
+    image?: string;
 }
 
 function computeVpcCniYaml(yamlPath: string, args: VpcCniInputs): string {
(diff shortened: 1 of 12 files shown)


nodejs/eks/cni.ts
```diff
@@ -87,21 +117,24 @@ function computeVpcCniYaml(yamlPath: string, args: VpcCniInputs): string {
...
-    if (args.warmEniTarget !== undefined) {
+    if (args.warmEniTarget) {
         env.push({name: "WARM_ENI_TARGET", value: args.warmEniTarget.toString()});
     }
-    if (args.warmIpTarget !== undefined) {
+    if (args.warmIpTarget) {
         env.push({name: "WARM_IP_TARGET", value: args.warmIpTarget.toString()});
     }
+    env.push({name: "AWS_VPC_K8S_CNI_LOGLEVEL", value: args.logLevel ? args.logLevel.toString() : "DEBUG"});
+    env.push({name: "AWS_VPC_K8S_CNI_LOG_FILE", value: args.logFile ? args.logFile.toString() : "stdout"});
```
lukehoban: Are these not the existing defaults?  I expected these to be similar to the above (using `if`) but perhaps this is intentional?

nodejs/eks/cni.ts
```diff
@@ -87,21 +117,24 @@ function computeVpcCniYaml(yamlPath: string, args: VpcCniInputs): string {
...
+    daemonSet.spec.template.spec.containers[0].image = args.image ? args.image.toString() : CNI_IMAGE;
```
lukehoban: Shouldn’t we just update the image in the yaml to the right default and use similar overrides as above?

nodejs/eks/cni/aws-k8s-cni.yaml
```diff
@@ -81,15 +81,20 @@ spec:
       tolerations:
         - operator: Exists
       containers:
-        - image: 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.5.0
```
lukehoban: Does this now march any released version of this yaml from the source?  Would love to keep this identical to some released version if possible.

nodejs/eks/cni.ts
metral: Yes, `DEBUG` is a default that was being set in the YAML manifest, but I moved it into `cni.ts` and removed it from the `YAML` manifest to make it user-configurable.

I did not originally use `if` statements, as this would then only set `logLevel` and `logFile` if the user provided them.

But if we adopt the override approach then we'd address your feedback.

nodejs/eks/cni/aws-k8s-cni.yaml
metral: The upstream YAML manifest and ours match. Below is the diff between upstream and ours.
The main differences lie in:

- Removing the image, and `AWS_VPC_K8S_CNI_LOGLEVEL` as we set those above in `cni.ts`
- Enabling the use of the readiness & liveness probes since these are omitted upstream as this healthz endpoint is only available on CNI images >= `v1.5.2` which we use now.

```
$ diff -u upstream-aws-k8s-cni.yaml aws-k8s-cni.yaml
--- upstream-aws-k8s-cni.yaml   2019-08-08 11:28:15.838079451 -0700
+++ aws-k8s-cni.yaml    2019-08-07 19:12:21.888614038 -0700
@@ -81,23 +81,20 @@
       tolerations:
         - operator: Exists
       containers:
-        - image: 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.5.2
-          imagePullPolicy: Always
+        - imagePullPolicy: Always
           ports:
             - containerPort: 61678
               name: metrics
           name: aws-node
-          #readinessProbe:
-          #  exec:
-          #    command: ["/app/grpc_health_probe", "-addr=:50051"]
-          #  initialDelaySeconds: 5
-          #livenessProbe:
-          #  exec:
-          #    command: ["/app/grpc_health_probe", "-addr=:50051"]
-          #  initialDelaySeconds: 5
+          readinessProbe:
+            exec:
+              command: ["/app/grpc_health_probe", "-addr=:50051"]
+            initialDelaySeconds: 5
+          livenessProbe:
+            exec:
+              command: ["/app/grpc_health_probe", "-addr=:50051"]
+            initialDelaySeconds: 5
           env:
-            - name: AWS_VPC_K8S_CNI_LOGLEVEL
-              value: DEBUG
             - name: MY_NODE_NAME
               valueFrom:
                 fieldRef:
```

metral:
> Shouldn’t we just update the image in the yaml to the right default and use similar overrides as above?

I'll make changes to adopt the override approach instead of the current one, and not remove bits from the YAML manifest directly to maintain that with upstream as best we can.

metral:

lukehoban:
Great that this is ultimately addressed by fixes in 1.5.2 CNI!

metral:
> Great that this is ultimately addressed by fixes in 1.5.2 CNI!

Agreed!

Also, the explicit removal of the NGINX workload and `aws-cni` also helped to achieve successful cluster tear downs. https://github.com/pulumi/pulumi-eks/issues/215 could help here, and would remove our need to do these tear downs in the `migrate-nodegroup` test.

Issue #219, opened 2019-08-12
Using I/O to form the file path of the CNI YAML creates errors in the dynamic provider since it can't locate the file


metral:
A user hit this error per a recent PR [change](https://github.com/pulumi/pulumi-eks/pull/201) that attempted to fix how the file path of the CNI YAML file is computed.

```
Diagnostics:
  pulumi-nodejs:dynamic:Resource (arcus-alpha-aws-vpc-cni):
    error: Plan apply failed: ENOENT: no such file or directory, open '/PROJECT_PATH/src/node_modules/@pulumi/eks/cni/aws-k8s-cni.yaml'
```

A temporary work around seems to be to export the state, manually edit the CNI file path stored in the dynamic provider to the correct path, and then re-import the state.

Related: https://github.com/pulumi/pulumi-eks/issues/149.

We should consider ingesting the CNI YAML content and then passing this value to the dynamic provider, rather than trying to form the file's relative path on the client and read it through I/O.

PR #223, opened 2019-08-13
fix(cni): read CNI YAML outside of the dynamic provider and update to v1.5.3

metral:
Fixes https://github.com/pulumi/pulumi-eks/issues/219
Related https://github.com/pulumi/pulumi-eks/issues/149

### DIFF
--- a/nodejs/eks/cni/aws-k8s-cni.yaml
+++ b/nodejs/eks/cni/aws-k8s-cni.yaml
@@ -81,7 +81,7 @@ spec:
       tolerations:
         - operator: Exists
       containers:
-        - image: 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.5.2
+        - image: 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.5.3
           imagePullPolicy: Always
           ports:
             - containerPort: 61678
@@ -104,10 +104,6 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
-            - name: WATCH_NAMESPACE
-              valueFrom:
-                fieldRef:
-                  fieldPath: metadata.namespace
           resources:
             requests:
               cpu: 10m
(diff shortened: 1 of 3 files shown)


nodejs/eks/cluster.ts
```diff
@@ -260,7 +262,9 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
     }
 
     // Create the VPC CNI management resource.
-    const vpcCni = new VpcCni(`${name}-vpc-cni`, kubeconfig, args.vpcCniOptions, { parent: parent });
+    const yamlPath = path.join(__dirname, "cni", "aws-k8s-cni.yaml");
+    const cniYamlText = fs.readFileSync(yamlPath).toString();
+    const vpcCni = new VpcCni(`${name}-vpc-cni`, cniYamlText, kubeconfig, args.vpcCniOptions, { parent: parent });
```
lukehoban: This looks fine as-is - but FWIW - I had been imagining a slightly more tactical change that kept this inside the `VpcCni` constructor, just not inside the `applyVpcCniYaml` implementation (which is run inside a different process).

nodejs/eks/cni/aws-k8s-cni.yaml
```diff
@@ -104,10 +104,6 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
-            - name: WATCH_NAMESPACE
-              valueFrom:
-                fieldRef:
-                  fieldPath: metadata.namespace
```
lukehoban: Did this go away due to a change in the `1.5.3` upstream?

nodejs/eks/cluster.ts
metral: I'll work on a change that has it in the `VpcCni` constructor.

nodejs/eks/cni/aws-k8s-cni.yaml
metral: Yes. I noted this in the commit:

> Based on: https://git.io/fj56i

Here's the upstream [PR](https://github.com/aws/amazon-vpc-cni-k8s/pull/578).

metral:

PR #230, opened 2019-08-19
fix(examples): typos, readme desc, and rm dupe tags of asg & cf

metral:
### Proposed changes

- fix(examples): typos, readme desc, and rm dupe tags of asg & cf
- test(makefile): bump up debug logging verbosity from level 2 -> 9


metral:
Closing in favor of: https://github.com/pulumi/pulumi-eks/pull/232

Issue #58, opened 2019-02-04
Failure to serialize YAML


lukehoban:
A user reported that they started seeing the error below in CI (they believe with no code changes). 

```
 +  pulumi:pulumi:Stack eks-cluster-eks-cluster-cd-test creating YAMLException: unacceptable kind of an object to dump [object Undefined]
 +  pulumi:pulumi:Stack eks-cluster-eks-cluster-cd-test creating     at writeNode (/builds/org/app/example/pulumi/eks-cluster/node_modules/js-yaml/lib/js-yaml/dumper.js:756:13)
 +  pulumi:pulumi:Stack eks-cluster-eks-cluster-cd-test creating     at writeBlockMapping (/builds/org/app/example/pulumi/eks-cluster/node_modules/js-yaml/lib/js-yaml/dumper.js:634:10)
 +  pulumi:pulumi:Stack eks-cluster-eks-cluster-cd-test creating     at writeNode (/builds/org/app/example/pulumi/eks-cluster/node_modules/js-yaml/lib/js-yaml/dumper.js:727:9)
 +  pulumi:pulumi:Stack eks-cluster-eks-cluster-cd-test creating     at writeBlockSequence (/builds/org/app/example/pulumi/eks-cluster/node_modules/js-yaml/lib/js-yaml/dumper.js:521:9)
 +  pulumi:pulumi:Stack eks-cluster-eks-cluster-cd-test creating     at writeNode (/builds/org/app/example/pulumi/eks-cluster/node_modules/js-yaml/lib/js-yaml/dumper.js:740:9)
 +  pulumi:pulumi:Stack eks-cluster-eks-cluster-cd-test creating     at dump (/builds/org/app/example/pulumi/eks-cluster/node_modules/js-yaml/lib/js-yaml/dumper.js:817:7)
 +  pulumi:pulumi:Stack eks-cluster-eks-cluster-cd-test creating     at Object.safeDump (/builds/org/app/example/pulumi/eks-cluster/node_modules/js-yaml/lib/js-yaml/dumper.js:823:10)
 +  pulumi:pulumi:Stack eks-cluster-eks-cluster-cd-test creating     at pulumi.all.apply (/builds/org/app/example/pulumi/eks-cluster/node_modules/@pulumi/cluster.ts:216:27)
[...]

PR #231, opened 2019-08-19
fix(iam): improve YAML error handling & reporting in IAM ops

metral:
### Proposed changes

fix(iam): improve YAML error handling & reporting in IAM ops
### Related issues (optional)

Fixes https://github.com/pulumi/pulumi-eks/issues/58

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -405,22 +405,34 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
 
     const roleMappings = pulumi.all([pulumi.output(args.roleMappings || []), instanceRoleMappings])
         .apply(([mappings, instanceMappings]) => {
-            return jsyaml.safeDump([...mappings, ...instanceMappings].map(m => ({
-                rolearn: m.roleArn,
-                username: m.username,
-                groups: m.groups,
+            let mappingYaml = "";
+            try {
+                mappingYaml = jsyaml.safeDump([...mappings, ...instanceMappings].map(m => ({
+                    rolearn: m.roleArn,
+                    username: m.username,
+                    groups: m.groups,
             })));
+            } catch (e) {
+                throw new Error(`The IAM role mappings provided could not be properly serialized to YAML for the aws-auth ConfigMap`);
+            }
+            return mappingYaml;
         });
     const nodeAccessData: any = {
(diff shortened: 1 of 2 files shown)


nodejs/eks/cluster.ts
```diff
@@ -314,22 +314,40 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
 
     const roleMappings = pulumi.all([pulumi.output(args.roleMappings || []), instanceRoleMappings])
         .apply(([mappings, instanceMappings]) => {
-            return jsyaml.safeDump([...mappings, ...instanceMappings].map(m => ({
+            mappings.map(m => {
+                if (!checkArnExists(pulumi.output(m.roleArn))) {
+                    throw new Error(`The IAM role ARN ${m.roleArn} could not be properly retrieved or does not exist`);
+                }
+            });
```
lukehoban: I don't think this part of the change is necessary - from the original bug report that part of the issue will be fixed more properly by https://github.com/pulumi/pulumi-terraform/issues/262.  We don't in general want to `.get` every resource id the user mentions to us just to verify they didn't make a mistake.

nodejs/eks/cluster.ts
```diff
@@ -314,22 +314,40 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
+            const mappingYaml = jsyaml.safeDump([...mappings, ...instanceMappings].map(m => ({
                 rolearn: m.roleArn,
                 username: m.username,
                 groups: m.groups,
             })));
+            if (!mappingYaml) {
+                throw new Error(`The IAM role mappings provided could not be properly serialized to YAML for the aws-auth ConfigMap`);
+            }
```
lukehoban: From the linked issue - doesn't `safeDump` throw?  If so, don't we need to catch the internal `sameDump` error and provider a better one?

nodejs/eks/cluster.ts
metral: Ok

nodejs/eks/cluster.ts

metral:

metral:
Ping @lukehoban. Any thoughts?

metral:
Going to go ahead and merge to close this out. Feel free to open up an issue if there is a different approach we should take.

PR #241, opened 2019-09-04
Refactor test options, and add new tests that use RunUpdateTest

metral:
### Proposed changes

- chore(dependencies): update Go deps to use `@pulumi/pulumi` v1.0.0
- chore(tests): use `dependency` over `peerDependency` & update deps to v1.0.0
- refactor(tests): use base test options to reduce dupe configs
- tests(all): use `RunUpdateTest` to test updates from latest release

### Related issues (optional)

https://github.com/pulumi/pulumi-eks/issues/124


nodejs/eks/examples/tags/package.json
```diff
@@ -6,9 +6,7 @@
     },
     "dependencies": {
         "@pulumi/pulumi": "^1.0.0-beta",
-        "@pulumi/aws": "^1.0.0-beta"
-    },
-    "peerDependencies": {
+        "@pulumi/aws": "^1.0.0-beta",
```
lukehoban: While you are touching all these - may as well bump these all to `^1.0.0`?

nodejs/eks/examples/all_test.go
```diff
@@ -228,7 +218,10 @@ func Test_AllTests(t *testing.T) {
 		},
 	}
 
-	longTests := []integration.ProgramTestOptions{}
+	longTests := shortTests
```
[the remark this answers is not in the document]
metral: Yes, that's correct. The `RunLatestTest` is it's own separate test: https://github.com/pulumi/pulumi/pull/3110

This was the simplest approach I found that didn't require redefining tests, but allowed me to enable `RunLatestTest: true` for each.

nodejs/eks/examples/tags/package.json
metral: Good catch, thanks.

PR #247, opened 2019-09-05
[WIP - DO NOT MERGE] - Debugging CI parallelism & RunUpdateTest setting in examples and tests

metral:
### Proposed changes

- Re-enable longTests with RunUpdateTest set
- Use dev build of CLI for better logs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
     - sudo mv aws-iam-authenticator /usr/local/bin
     # Install Pulumi
     # Random one-off build for printing more diagnostic info.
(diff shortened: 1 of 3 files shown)


chrsmith:
Closing this PR out. I have found out all that I was looking for.

It looks like this all boils down to when the system is very overloaded, the OS isn't notifying us of timer callbacks in a timely fashion. This in-turn leads to us failing to update the lease token in a reasonable timeframe.

Looking at the behavior of the system in these situations, I know what changes we can make on the CLI-side to make things a little more robust. (Though in general if the system is super-bogged down, we are going to have a bad time regardless.)

Thanks for the help in setting up the repro @metral !

PR #248, opened 2019-09-05
Use dev build of CLI for better logs

chrsmith:
Just looking to reproduce Travis CI failures.


metral:
Added your commit from this PR over in the [testing / debugging PR](https://github.com/pulumi/pulumi-eks/pull/247) that has all the other settings enabled to repro the increased traffic when doubling up tests via `RunUpdateTest`.

Current CI run to follow along: https://travis-ci.com/pulumi/pulumi-eks/builds/126149039

PR #249, opened 2019-09-06
Auto worker node draining

mazamats:
### Proposed changes

This change adds a Lambda function along with an ASG/Cloudwatch Lifecycle hook that will trigger whenever an AMI update happens.

The lambda will drain 1 node at a time and pause for 1 minute in between each node upgrade.

### Related issues

https://github.com/pulumi/pulumi-eks/issues/40


metral:
cc @lukehoban

mazamats:
Closing now that managed node groups handle graceful draining: https://www.terraform.io/docs/providers/aws/r/eks_node_group.html

PR #254, opened 2019-09-13
Sync `scripts/get-version` with pulumi/pulumi

ellismg:


metral:
@pgavlin Are we still intending to push this PR through?

pgavlin:
Yes.

PR #260, opened 2019-09-25
tests(examples): enable updateTests that run RunUpdateTest on examples

metral:
### Proposed changes

tests(examples): enable longTests that run RunUpdateTest for examples

### DIFF
--- a/nodejs/eks/examples/all_test.go
+++ b/nodejs/eks/examples/all_test.go
@@ -235,18 +235,7 @@ func Test_AllTests(t *testing.T) {
 		},
 	}
 
-	longTests := []integration.ProgramTestOptions{}
-	// longTests := shortTests
-	// for _, longTest := range longTests {
-	// 	longTest.RunUpdateTest = true
-	// }
-
-	tests := shortTests
-	if !testing.Short() {
-		tests = append(tests, longTests...)
-	}
-
-	for _, ex := range tests {
+	for _, ex := range shortTests {
 		example := ex.With(integration.ProgramTestOptions{
 			ReportStats: integration.NewS3Reporter("us-west-2", "eng.pulumi.com", "testreports"),
 			Tracing:     "https://tracing.pulumi-engineering.com/collector/api/v1/spans",
--- a/nodejs/eks/examples/examples_test.go
+++ b/nodejs/eks/examples/examples_test.go
@@ -38,8 +38,8 @@ func Test_Examples(t *testing.T) {
 		return
 	}
 
-	// Base options shared amongst all tests.
-	base := integration.ProgramTestOptions{
+	// Base options shared amongst all shortTests.
+	shortTestsBase := integration.ProgramTestOptions{
 		Config: map[string]string{
 			"aws:region": region,
 		},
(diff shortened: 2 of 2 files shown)


nodejs/eks/examples/all_test.go
```diff
@@ -236,10 +236,6 @@ func Test_AllTests(t *testing.T) {
 	}
 
 	longTests := []integration.ProgramTestOptions{}
-	// longTests := shortTests
-	// for _, longTest := range longTests {
-	// 	longTest.RunUpdateTest = true
-	// }
```
lukehoban: If we really don't need this - can we delete it?  If so - can we also clean up the rest of the code around here since `longTests` no longer is needed?

nodejs/eks/examples/examples_test.go
```diff
@@ -92,11 +92,10 @@ func Test_Examples(t *testing.T) {
...
 
-	longTests := []integration.ProgramTestOptions{}
-	// longTests := shortTests
-	// for _, longTest := range longTests {
-	// 	longTest.RunUpdateTest = true
-	// }
+	longTests := shortTests
+	for _, longTest := range longTests {
+		longTest.RunUpdateTest = true
+	}
```
lukehoban: Why not just modify all the tests upfront (like in `base`)?  Also `shortTests` vs. `longTests` naming is weird here - there is only one set of tests right?

lukehoban: Or is the goal to run *both* with and without `RunUpdateTest`?  If so - that doesn't seem to be what this will accomplish.

metral: > Why not just modify all the tests upfront (like in base)? Also shortTests vs. longTests naming is weird here - there is only one set of tests right?

The `RunUpdateTest` is its own, singular test - that is, it does not test both the locally-linked && latest release tests; hence the need to set both sets of tests. I threw it in a for-loop to not have to redefine the same tests in `shortTests` given that the only diff was the `RunUpdateTest: true` setting.

It seems that redefining the same tests as `updateTests` separately from `shortTests` instead of using the for-loop will help clear things up. I'll make these changes.

lukehoban: >  I threw it in a for-loop to not have to redefine the same tests in shortTests given that the only diff was the RunUpdateTest: true setting.

I don’t think that’s what the code as written accomplished - does it? If that is the intent - then I agree a for loop makes sense - but I think you would need to make a copy of the tests before mutating them - right?

metral: I totally overlooked that `shortTests` is a slice (pointer in Go) and that I was mutating it's value for `longTests` 🤦‍♂. Thank you for catching this!

I've changed it to use new, separate tests `updateTests`.

metral:

nodejs/eks/examples/examples_test.go
lukehoban: FWIW - I was trying to just suggest that you copy the tests and overwrite this in the copies instead of having to write every test out twice.  But at least this now looks correct! :-).

PR #263, opened 2019-10-13
Examples: Change the examples to run as individual tests

stack72:


metral:
For readers, the context for the change is:

> This makes it possible to run individual tests much more easily, and better aligns with go test practices. Note that if we do want more logical groupings, we can split tests into separate modules of files. But we can also just using naming conventions to allow running subsets of tests via regexps if that makes sense.

https://github.com/pulumi/home/issues/581#issuecomment-541224981

PR #272, opened 2019-11-01
Update securitygroup.ts - Quick comment fix

ben-nagengast:
### Proposed changes

Looks like this comment was copied from the above value and didn't get updated. I think this is a clearer representation of the purpose of the purpose of this property.

### Related issues (optional)


pgavlin:
Thanks for the fix!

PR #283, opened 2019-12-03
Support Fargate

### LINKED ISSUE #282: Support Fargate
AWS released support for [EKS on Fargate](https://aws.amazon.com/blogs/aws/amazon-eks-on-aws-fargate-now-generally-available/) today.

The new `aws.eks.FargateProfile` resource will allow defining when and how pods should be launched on Fargate instead of into existing NodeGroups.  This should be usable already with `@pulumi/eks` by just adding the `aws.eks.FargateProfile` on top of the existing cluster.

We could also make this even easier for users of `@pulumi/eks` to set up by providing an optional `fargate` property on the cluster to easily build a cluster with Fargate support.

lukehoban:
Adds support for adding Fargate support to an EKS cluster using `new eks.Cluster("cluster", { ..., fargate: true })`.

Fixes #282

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -15,9 +15,14 @@
 import * as aws from "@pulumi/aws";
 import * as k8s from "@pulumi/kubernetes";
 import * as pulumi from "@pulumi/pulumi";
+
+import * as childProcess from "child_process";
+import * as fs from "fs";
 import * as https from "https";
 import * as jsyaml from "js-yaml";
 import fetch from "node-fetch";
+import * as process from "process";
+import * as tmp from "tmp";
 import * as which from "which";
 
 import { VpcCni, VpcCniOptions } from "./cni";
@@ -345,29 +350,6 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
         provider: args.creationRoleProvider ? args.creationRoleProvider.provider : undefined,
     });
 
-    let fargateProfile: aws.eks.FargateProfile | undefined;
-    if (args.fargate) {
-        const fargate = args.fargate !== true ? args.fargate : {};
-        const podExecutionRoleArn = fargate.podExecutionRoleArn || (new ServiceRole(`${name}-podExecutionRole`, {
-            service: "eks-fargate-pods.amazonaws.com",
-            managedPolicyArns: [
-                "arn:aws:iam::aws:policy/AmazonEKSFargatePodExecutionRolePolicy",
-            ],
-        }, { parent: parent })).role.apply(r => r.arn);
-        const selectors = fargate.selectors || [
-            // For `fargate: true`, default to including the `default` namespaces and
-            // `kube-system` namespaces so that all pods by default run in Fargate.
-            { namespace: "default" },
-            { namespace: "kube-system" },
-        ];
-        fargateProfile = new aws.eks.FargateProfile(`${name}-fargateProfile`, {
-            clusterName: eksCluster.name,
-            podExecutionRoleArn: podExecutionRoleArn,
-            selectors: selectors,
-            subnetIds: pulumi.output(clusterSubnetIds).apply(subnets => computeWorkerSubnets(parent, subnets)),
-        }, { parent: parent });
-    }
-
     // Instead of using the kubeconfig directly, we also add a wait of up to 5 minutes or until we
     // can reach the API server for the Output that provides access to the kubeconfig string so that
     // there is time for the cluster API server to become completely available.  Ideally we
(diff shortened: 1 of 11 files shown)


nodejs/eks/cluster.ts
```diff
@@ -287,16 +288,16 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
-        ],
-      }, { parent: parent })).role;
+        eksRole = (new ServiceRole(`${name}-eksRole`, {
+            service: "eks.amazonaws.com",
+            description: "Allows EKS to manage clusters on your behalf.",
+            managedPolicyArns: [
+                "arn:aws:iam::aws:policy/AmazonEKSClusterPolicy",
+                "arn:aws:iam::aws:policy/AmazonEKSServicePolicy",
+            ],
+        }, { parent: parent })).role;
```
lukehoban: Lots of whitespace changes from running `format` on this file - I tried undoing them a half dozen times to avoid dirtying this PR - but finally gave up - suggest reviewing with "Hide whitespace changes" turned on.

nodejs/eks/examples/cluster/index.ts
```diff
@@ -26,6 +26,16 @@ const cluster2 = new eks.Cluster(`${projectName}-2`, {
     ],
 });
 
+// Create an EKS cluster with the fargate configuration.
+const vpc3 = new awsx.ec2.Vpc(`${projectName}-3`);
+const cluster3 = new eks.Cluster(`${projectName}-3`, {
+    vpcId: vpc3.id,
+    privateSubnetIds: vpc3.privateSubnetIds,
```
lukehoban: Note that `vpcId` and `privateSubnetIds` are effectively necessary to use `fargate: true`:
1. Fargate for EKS only works in private subnets, and the default VPC has only public subnets, so you must always provide an explicit VPC along with `fargate: true`
2. The current design of this library requires passing both subnets and VPC in - ideally we would augment this library to accept an `awsx.ec2.Vpc` as an alternative to a `vpc.id`, and if that is passed, infer the public and private subnets from it.  That's an orthogonal change to this PR - so something we can do separately.

nodejs/eks/examples/examples_test.go
```diff
@@ -39,6 +39,7 @@ func TestAccCluster(t *testing.T) {
 					info.Deployment.Resources,
 					info.Outputs["kubeconfig1"],
 					info.Outputs["kubeconfig2"],
+					info.Outputs["kubeconfig3"],
```
metral: We're going to have to rethink how we test for this example in the k8s test framework, as we currently parse the stack resources searching for nodes to compute a `desiredNodeCount` to anticipate during testing after a cluster has come up:

https://github.com/pulumi/pulumi-eks/blob/master/utils/utils.go#L487

nodejs/eks/examples/cluster/index.ts
metral: > 1. Fargate for EKS only works in private subnets, and the default VPC has only public subnets, so you must always provide an explicit VPC along with fargate: true

We should call out the need for private subnets in the comments / docs.

> 2. The current design of this library requires passing both subnets and VPC in - ideally we would augment this library to accept an awsx.ec2.Vpc as an alternative to a vpc.id, and if that is passed, infer the public and private subnets from it. That's an orthogonal change to this PR - so something we can do separately.

SGTM 👍

lukehoban:
It appears there is one more piece that will need to be integrated here in order to have CoreDNS function correctly in cases that coredns is run on Fargate. This is documented here:  https://docs.aws.amazon.com/eks/latest/userguide/fargate-getting-started.html#fargate-gs-coredns. 

To do this from Pulumi we will need to find a way to patch an existing resource created and managed outside of Pulumi (the `coredns` Deployment) after deploying the Fargate profile.

nodejs/eks/examples/examples_test.go
lukehoban: I added a change that skips validating the number of Nodes that exist if there are no NodeGroups attached.  I am hopeful that unblocks this, though we may need more flexibility in the test framework for future tests.

metral:
Looks good overall 🎉.

One thing we need to add is the conditional deployment of the AWS CNI daemonset. We should not be deploying this to fargate clusters since it's not needed, and because daemonsets are not allowed on fargate.

lukehoban:
> One thing we need to add is the conditional deployment of the AWS CNI daemonset. We should not be deploying this to fargate clusters since it's not needed, and because daemonsets are not allowed on fargate.

We do need to look into this - but i don’t believe this prevents things from working, and I think it will unfortunately be harder to do this conditionally because you may frequently have a mix of Fargate and EC2 in your cluster. I’d suggest we revisit CNI more generally as part of a follow up.

PR #292, opened 2019-12-17
Ensure that all data source calls are async

lukehoban:
Also allow passing a parent to inherit provider from.

Part of pulumi/pulumi#3528.


lukehoban:
Travis failure appears unrelated to the changes in this PR.

PR #299, opened 2020-01-03
[prototype] Tweaks to support multilang

lukehoban:
These changes don't change any behaviour, but allow the component to be used with the current multilang prototype in pulumi/pulumi#3711.

There are two changes:
1. Pass `id` explcitly instead of strongly typed references.  This change shouldn't be needed, but the current multilang prototype makes a breaking change to pass custom resource references as URNs instead of `id`s.  This breaks cases where strongly typed CustomResoures are passed as inputs.  We work around that sugar here for now by manually passing the `id` property.  (We should be able to undo this breaking change in the prototype).
2. Return more resource outputs as `registerOutput`s.  The only output proeprties that get serialized to guest langauges are those returned as part of `registerOutput`s (vs. just stored on `this.`), so we must return all output proeprties in the `registerOutput`s as well.


justinvp:
I'm going to close this as no longer relevant.

PR #301, opened 2020-01-08
Fix deprecated tag in JSDoc comments

justinvp:
The description in the doc comment should come before the `@deprecated` tag, otherwise the description will be included as part of the deprecation message.

Part of https://github.com/pulumi/docs/issues/2229


justinvp:
@metral, would you mind merging this despite Travis not being green? It's just changing some doc comments. Would like to close out the bug.

PR #308, opened 2020-01-21
Enable the ability for EKS to run on ap-east-1 / me-south-1

stack72:
Fixes: https://github.com/pulumi/examples/issues/517

The owner of the AMI that Amazon uses to run on EKS in ap-east-1
isn't actually owned by amazon - or rather, it's not the same owner id
that we have hard coded in our Eks setup

I have followed the EKS ami list here - https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html
to find the owner of the AMIs

I also added the Owner for me-south-1. The reason we needed to do
this is that these are regions that need to be opted-in to

### DIFF
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -415,9 +415,15 @@ ${customUserData}
                 },
             ];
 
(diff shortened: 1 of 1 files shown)


stack72:
@lukehoban just pushed an update to what each of the IDs are and where I was able to find them :)

Issue #340, opened 2020-02-26
EKS CI failing due to cluster creation timeouts > 20 minutes


metral:
### Problem description

Examples are failing in CI due to cluster creation timeouts lasting longer than 20 minutes.

### Errors & Logs

```
[ odejs/eks/examples/oidc-iam-sa ]     	* creating urn:pulumi:p-it-travis-job-oidc-iam-s-e6627796::example-oidc-iam-sa::eks:index:Cluster$aws:eks/cluster:Cluster::example-oidc-iam-sa-eksCluster: 
timeout while waiting for state to become 'ACTIVE' (last state: 'CREATING', timeout: 20m0s)
```

```
[ /nodejs/eks/examples/nodegroup ]     	* creating urn:pulumi:p-it-travis-job-nodegroup-c914f7ad::example-nodegroup::eks:index:Cluster$aws:eks/cluster:Cluster::example-nodegroup-iam-advanced-eksCluster: 
timeout while waiting for state to become 'ACTIVE' (last state: 'CREATING', timeout: 20m0s)
```

```
[ ks/nodejs/eks/examples/cluster ]     	* creating urn:pulumi:p-it-travis-job-cluster-836ebc28::example-cluster::eks:index:Cluster$aws:eks/cluster:Cluster::example-cluster-1-eksCluster: 
timeout while waiting for state to become 'ACTIVE' (last state: 'CREATING', timeout: 20m0s)
```

```
[ s/eks/examples/storage-classes ]     	* creating urn:pulumi:p-it-travis-job-storage-cl-109ba345::example-storage-classes::eks:index:Cluster$aws:eks/cluster:Cluster::example-storage-classes-2-eksCluster: 
timeout while waiting for state to become 'ACTIVE' (last state: 'CREATING', timeout: 20m0s)
```

```
[ ples/tests/managed-ng-aws-auth ]     	* creating urn:pulumi:p-it-travis-job-managed-ng-4612b205::managed-ng-aws-auth::eks:index:Cluster$aws:eks/cluster:Cluster::managed-ng-aws-auth-eksCluster: 
timeout while waiting for state to become 'ACTIVE' (last state: 'CREATING', timeout: 20m0s)
```

Refs:

- https://travis-ci.com/pulumi/pulumi-eks/builds/150597238
- https://travis-ci.com/pulumi/pulumi-eks/builds/150423630
- https://travis-ci.com/pulumi/pulumi-eks/builds/150405266

### Suggestions for a fix

[...]

PR #341, opened 2020-02-26
Extend Cluster Timeouts

stack72:
Fixes: #340

Sometimes clusters can time out whne creating, to allow this not
to be the case, we are going to specify a create timeout of 1hour
by default. This means the user *should not* need to worry about
custom timeouts.

For good measure, I added a delete timeout extension as well since
that can sometimes that a long time as well

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ vendor/
 **/Pulumi.*.yaml
 **/yarn-error.log
 yarn.lock
+.idea/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ vendor/
 **/Pulumi.*.yaml
 **/yarn-error.log
 yarn.lock
+.idea/
```
metral: Can we drop this since it does not seem pertinent to this PR, or is there some particular reason for its inclusion?

stack72: I added to the PR since I originally commited it on the PR and then had to remove it and then needed to recreate my environment. Happy to remove it if you want but just for reference, we don't have a standardised .gitignore - maybe this is something we should have

Issue #349, opened 2020-03-17
CI test flake failing due to unready pods in smoke test in TestAccOidcIam


metral:
### Problem description

TestAccOidcIam is flaking in CI due to the s3-job assertion test hitting an error on runs that seem related to the OIDC provider.

See s3-job Pods below where `*-7967v` failed first, but then `-mdqnj` succeeded.

```
    utils.go:296: Waiting for Pod "s3-job-f9bq4el9-7967v" to be ready. Retrying...
    utils.go:296: Waiting for Pod "s3-job-f9bq4el9-7967v" to be ready. Retrying...

    utils.go:303: Pod: s3-job-f9bq4el9-mdqnj | Ready Status: true

    utils.go:296: Waiting for Pod "s3-job-f9bq4el9-**rm7bc**" to be ready. Retrying...
    utils.go:296: Waiting for Pod "s3-job-f9bq4el9-rm7bc" to be ready. Retrying...
    utils.go:296: Waiting for Pod "s3-job-f9bq4el9-rm7bc" to be ready. Retrying...
   ...
```

### Affected product version(s)

v0.18.24

### Reproducing the issue

Have not been able to locally repro on 5 runs. CI hits it on occasion:

- https://travis-ci.com/github/pulumi/pulumi-eks/builds/153610667#L2038
- https://travis-ci.com/github/pulumi/pulumi-eks/builds/153305223#L1959

metral:
Re-opening as #351 did not resolve this issue.

metral:
It appears this is tied to the CA thumbprint mismatching, but the odd part is that it does not always happen.

```
2020/03/18 18:26:19 Can't upload to S3: WebIdentityErr: failed to retrieve credentials
caused by: InvalidIdentityToken: OpenIDConnect provider's HTTPS certificate doesn't match configured thumbprint
```

Related https://github.com/pulumi/pulumi-eks/issues/342

metral:
Over 10 runs were tried locally, as well as clearing RAM and DNS cache, issuing reboots, and using two machines.

This does not seem to be related to the OIDC thumbprint as it always returns the expected fingerprint for `oidc.eks.{region}.amazonaws.com`.

There's a sneaky suspicion this is related to nodes not being fully READY after joining the cluster by the time the pod or job are scheduled - only the control plane needs to be up for workloads to be accepted.

metral:
Seems that this is tied to:

- nodes not being quite ready
[...]

PR #353, opened 2020-03-19
test(oidc): make pod sleep after run, to avoid termination crashloop

metral:
### Proposed changes

test(oidc): make pod sleep after run to avoid termination crashloop

#351 moved from a Job for the S3 assertion test to a Pod because Job's create new pods until one succeeds, where as a regular Pod by default always restarts the same pod on failure until success is reached. For the CI smoketest, using a Job created issues when the Job's failed pods stick around even though the Job may have eventually succeeded.

The caveat though in moving to using a Pod instead, is that it should remain running for the sake of the smoke test, otherwise after it successfully runs, the Pod will terminate, come up again, succeed, and terminate as the [example](https://github.com/mhausenblas/s3-echoer) referenced in the [AWS blog](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) is not long-running by default.

This change adds a sleep command to the end of the S3 pod test to keep the pod up for the sake of the smoketest.

### Related issues (optional)

Fixes https://github.com/pulumi/pulumi-eks/issues/349

### DIFF
--- a/nodejs/eks/examples/oidc-iam-sa/index.ts
+++ b/nodejs/eks/examples/oidc-iam-sa/index.ts
@@ -120,7 +120,7 @@ const s3Pod = pulumi.all([bucketName, regionName]).apply(([bName, region]) => {
                         name: podName,
                         image: "amazonlinux:2018.03",
                         command: ["sh", "-c",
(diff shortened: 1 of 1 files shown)


metral:
This PR has passed 5 times in CI runs, so the flake _should_ now be resolved.

Issue #304, opened 2020-01-13
Replace dependency on aws-iam-authenticator with aws cli


seivan:
The [dependency](https://github.com/pulumi/pulumi-eks/blob/master/nodejs/eks/cluster.ts#L231) should be switched out to the `aws` CLI and use `aws eks get-token` instead. I mean, the aws cli is mandatory anyway, so it makes it one less dependency to have installed.

leezen:
Thanks for the suggestion!

Issue #338, opened 2020-02-22
Updating from 0.18.18 -> 0.18.19 results in preview showing k8s resources getting replaced


kennyjwilli:
### Problem description

Updating my @pulumi/eks from 0.18.18 to 0.18.19 results in the `pulumi up` preview indicating that many/all of my k8s resources should be replaced. 

My starting pulumi deps: 

```json
"@pulumi/pulumi": "1.9.1",
"@pulumi/kubernetes": "1.4.5",
"@pulumi/eks": "0.18.18",
"@pulumi/awsx": "0.19.1",
"@pulumi/aws": "1.18.0",
```

With these deps, I run `pulumi up` and receive the following output.

```
Previewing update (ComputeSoftware/dev):
     Type                 Name             Plan     Info
     pulumi:pulumi:Stack  cs-platform-dev           1 message
 
Diagnostics:
  pulumi:pulumi:Stack (cs-platform-dev):
    (node:26996) ExperimentalWarning: queueMicrotask() is experimental.
```

I then update to @pulumi/eks 0.18.19 and npm install. Running `pulumi up` results in many/all k8s resources getting replaced, according to the preview. Here are the first few lines from the preview: 

```
     pulumi:pulumi:Stack                                               cs-platform-dev                                                   1 message
     ├─ eks:index:Cluster                                              k8s-cluster-dev                                                   
 ~   │  ├─ aws:eks:Cluster                                             k8s-cluster-dev-eksCluster                            update      [diff: ~tags]
 +-  │  ├─ pulumi:providers:kubernetes                                 k8s-cluster-dev-provider                              replace     [diff: ~kubeconfig]
 +-  │  ├─ pulumi:providers:kubernetes                                 k8s-cluster-dev-eks-k8s                               replace     [diff: ~kubeconfig]
```

The diff on the kubeconfig is not useful since the output side simply says `=> output<string>`.

From the Pulumi Slack, @lukehoban said: 
> Theres a good change that this is just the preview being conservative - and that nothing will actually get replaced.  But it's certainly dangerous.

lukehoban:
In `0.18.19` the EKS package changed the default tags that are applied (added `Name`).  

This results in an `update` on the `aws:eks:Cluster`.  That leads to the outputs on the cluster being unknown (conservatively).  That leads to the `kubeconfig` being unknown.  That leads to the `pulumi:providers:kubernetes` provider needing to replace (conservative).

[...]

PR #362, opened 2020-03-30
refactor(aws-auth): replace aws-iam-authenticator with aws eks get-token

metral:
### Proposed changes

refactor(aws-auth): replace `aws-iam-authenticator` with `aws eks get-token`.

Note: for existing clusters, this change will recompute the kubeconfig used,
as its arguments and settings get updated to work with `aws eks get-token`.
It should not affect cluster access.

Examples of kubeconfig changes:

image
image

### Related issues (optional)

Closes https://github.com/pulumi/pulumi-eks/issues/304

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,9 @@ This includes:
 1. [Install Pulumi](https://www.pulumi.com/docs/reference/install).
 1. Install [Node.js](https://nodejs.org/en/download).
 1. Install a package manager for Node.js, such as [NPM](https://www.npmjs.com/get-npm) or [Yarn](https://yarnpkg.com/lang/en/docs/install).
-1. [Configure AWS Credentials](https://www.pulumi.com/docs/reference/clouds/aws/setup/).
-1. [Install AWS IAM Authenticator for Kubernetes](https://docs.aws.amazon.com/eks/latest/userguide/install-aws-iam-authenticator.html).
+1. [Install and Configure the AWS CLI](https://www.pulumi.com/docs/reference/clouds/aws/setup/).
+    * Use AWS CLI version >= `1.18.17`.
+    * See the [AWS docs](https://docs.aws.amazon.com/eks/latest/userguide/managing-auth.html) for more details.
 1. [Install `kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-kubectl).
 
 ## Installing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ### Improvements
 
+- refactor(aws-auth): replace aws-iam-authenticator with aws eks get-token
+  [#362](https://github.com/pulumi/pulumi-eks/pull/362)
+  **Note:** for existing clusters, this change will recompute the kubeconfig used,
+  as its auth arguments and settings get updated to work with `aws eks get-token`.
+  It should not affect cluster access or cause replacements of existing k8s
+  resources.
 - feat(nodegroup): use the latest recommended AMIs from the SSM store
   [#366](https://github.com/pulumi/pulumi-eks/pull/366)
 - feat(cluster): support HTTP(S) proxy for cluster readiness & OIDC config
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -163,11 +163,11 @@ function generateKubeconfig(
     clusterEndpoint: pulumi.Input<string>,
     certData: pulumi.Input<string>,
     opts?: KubeconfigOptions) {
-    let args = ["token", "-i", clusterName];
+    let args = ["eks", "get-token", "--cluster-name", clusterName];
     let env: { [key: string]: pulumi.Input<string>} | undefined;
 
     if (opts?.roleArn) {
-        args = [...args, "-r", opts.roleArn];
+        args = [...args, "--role", opts.roleArn];
     }
     if (opts?.profileName) {
         env = {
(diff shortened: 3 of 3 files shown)


metral:
> but updating the kubeconfig will show a replace for the cluster and resources on the cluster during preview

Not sure I'm following, could you elaborate?

Here's the preview of an update on an existing cluster from `examples/cluster`, the same for the examples from the OP. This PR `updates` the kubeconfig and it's dependents like the provider, it does not `replace` it.

image

lblackstone:
For the pulumi-kubernetes provider, if the kubeconfig changes in a first-class Provider resource, the preview will [show a `replace`](https://github.com/pulumi/pulumi-kubernetes/blob/138727c422c99300f65623609634bdeaf1335f17/pkg/provider/provider.go#L270-L278) for resources using that Provider.

This wouldn't be the case if the EKS cluster was managed in a separate stack, as this only happens when the kubeconfig is a computed value during preview.

metral:
>if the kubeconfig changes in a first-class Provider resource, the preview will show a replace for resources using that Provider.

Ah, right. Got it now.
Seems that the [conservative replace](https://github.com/pulumi/pulumi-kubernetes/blob/138727c422c99300f65623609634bdeaf1335f17/pkg/provider/provider.go#L270-L278) is related to https://github.com/pulumi/pulumi-eks/issues/338.

Other than the CHANGELOG updates in the PR, and notifying the community in Slack, any other suggestions for how best to roll this out and inform users that are not watching either or both?

lblackstone:
> Other than the CHANGELOG updates in the PR, and notifying the community in Slack, any other suggestions for how best to roll this out and inform users that are not watching either or both?

I think that's sufficient. If users have questions about it, you can point them to the CHANGELOG as a canonical reference.

metral:
@lukehoban Feedback?

metral:
>if the kubeconfig changes in a first-class Provider resource, the preview will show a replace for resources using that Provider.

In practice cluster access is not affected, but because the kubeconfig changes, it's dependencies from the k8s provider on will be signaled to be replaced during previews.

Thinking through this more, it seems best to signal this change along with the rest of the unreleased improvements and pending PRs by bumping and cutting a new minor release: 0.19.0.

Do folks agree?

README.md
```diff
@@ -37,8 +37,8 @@ This includes:
 1. [Install Pulumi](https://www.pulumi.com/docs/reference/install).
 1. Install [Node.js](https://nodejs.org/en/download).
 1. Install a package manager for Node.js, such as [NPM](https://www.npmjs.com/get-npm) or [Yarn](https://yarnpkg.com/lang/en/docs/install).
-1. [Configure AWS Credentials](https://www.pulumi.com/docs/reference/clouds/aws/setup/).
-1. [Install AWS IAM Authenticator for Kubernetes](https://docs.aws.amazon.com/eks/latest/userguide/install-aws-iam-authenticator.html).
+1. [Install and Configure the AWS CLI](https://www.pulumi.com/docs/reference/clouds/aws/setup/).
+  * AWS CLI version >= 1.18.17. See the [AWS docs](https://docs.aws.amazon.com/eks/latest/userguide/managing-auth.html) for more details.
```
lukehoban, line 41: Nit - This renders a bit funny (bullets inside numbered list).

lblackstone:
> replacements will not be required in practice. (Are we _sure_ that is true?)

As long as the actual cluster doesn't change, it will not replace the resources. The problem with preview is that the value of the kubeconfig isn't computed yet, so we can't tell for certain if the resources still exist, hence the replacement.

metral:
> But in the immediate term - can we include a clear message in README that indicates users may see kubernetes.Provider replacements during preview when updating to this version, but that replacements will not be required in practice. (Are we sure that is true?)

@lblackstone @lukehoban 

Tests were conducted with [two stacks](https://github.com/metral/stacked_cluster):
1. EKS cluster & k8s resources (k8s.Deployment) co-located in the same stack, and
2. a k8s resources (kx.Secret & k8s.Deployment) only stack, that ref'd and deployed into the cluster from the 1st stack.

Neither preview nor updates for either stack with this PR showed a replacement of any k8s resources, only updates to the kubeconfig, and the provider.

It does not appear that this PR change hits the conservative replacement for k8s resources when the kubeconfig changes. IIUC this may be because only the auth exec commands change from the `aws-iam-authenticator` to `aws eks get-token`, and not the cluster name and endpoint, which I believe is what p/k8s anchors on - is this correct Levi?

--- 

01-cluster preview & update:

image

---

02-apps preview & update:

image

lblackstone:
> not the cluster name and endpoint, which I believe is what p/k8s anchors on - is this correct Levi?

Yes, that's right. Thanks for verifying.

lblackstone:
I double-checked the logic in the k8s provider code, and it will preview a `replace` iff the `kubeconfig` is a computed value during preview.

However, Mike's test scenarios seem to show that this the `kubeconfig` value is resolved by the time this logic runs, both for the separate stack and single stack cases.

Here's the relevant logic in the k8s provider:

https://github.com/pulumi/pulumi-kubernetes/blob/39e2c0fc8d9ad3de5752bc2c0a2cdcadb0f50b7f/pkg/provider/provider.go#L270-L278

https://github.com/pulumi/pulumi-kubernetes/blob/39e2c0fc8d9ad3de5752bc2c0a2cdcadb0f50b7f/pkg/provider/provider.go#L314-L339

lblackstone:
I tested this locally as well, and wasn't able to cause a `replace` in preview. I think this is safe to merge.

Issue #401, opened 2020-06-24
nodegroup ASG name started hitting compile error: Type 'string' is not assignable to type 'Output<string>'.


metral:
### Problem description

Last night's [cron run](https://travis-ci.com/github/pulumi/pulumi-eks/builds/172787525) started to hit a compilation error for the ASG name. It is not clear what changed from previous CI runs since the `main` branch has been static for the last month.

### Errors & Logs

```
nodegroup.ts:560:9 - error TS2322: Type 'string' is not assignable to type 'Output<string>'.
  Type 'string' is not assignable to type 'OutputInstance<string>'.
560         autoScalingGroupName = outputs["NodeGroup"];
```

PR #402, opened 2020-06-24
Unblock CI errors due to recent type changes in pulumi-aws

metral:
### Proposed changes

- nodegroup(asgName): fix asgName definition
- fix(ex/default-sg): rm Output<string> tag values per string type reqs

### Related issues (optional)

Fixes https://github.com/pulumi/pulumi-eks/issues/401
Fixes https://github.com/pulumi/pulumi-eks/issues/404

### DIFF
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -552,12 +552,11 @@ ${customUserData}
         })),
     }, { parent, dependsOn: cfnStackDeps, provider });
 
-    let autoScalingGroupName = pulumi.output("");
-    cfnStack.outputs.apply(outputs => {
+    const autoScalingGroupName = cfnStack.outputs.apply(outputs => {
         if (!("NodeGroup" in outputs)) {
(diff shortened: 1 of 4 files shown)


nodejs/eks/nodegroup.ts
```diff
@@ -552,12 +552,11 @@ ${customUserData}
...
     }, { parent, dependsOn: cfnStackDeps, provider });
 
-    let autoScalingGroupName = pulumi.output("");
-    cfnStack.outputs.apply(outputs => {
+    const autoScalingGroupName = cfnStack.outputs.apply(outputs => {
         if (!("NodeGroup" in outputs)) {
             throw new Error("CloudFormation stack is not ready. Stack output key 'NodeGroup' does not exist.");
         }
-        autoScalingGroupName = outputs["NodeGroup"];
+        return outputs["NodeGroup"];
```
lukehoban, line 559: Interesting - looks like this would have been broken before - the autoscalingGroupName would always have been “” I believe?

Issue #410, opened 2020-07-28
Support for the ENI_CONFIG_LABEL_DEF environment variable


VRanga000:
### Problem description

When an EKS cluster worker node group is spread across multiple availability zones,
and we want to use custom CNI networking, EKS allows you to  specify the 
ENI_CONFIG_LABEL_DEF environment variable value for worker nodes
This is used to tell Kubernetes to automatically apply the ENIConfig for each Availability Zone
Ref: https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html (step 6(c)).

This env var is not currently exposed in the CNI module in pulumi-eks.

I have a PR to address this.

PR #411, opened 2020-07-28
Support for the ENI_CONFIG_LABEL_DEF environment variable

VRanga000:
### Proposed changes

When an EKS cluster worker node group is spread across multiple availability zones,
and we want to use custom CNI networking, EKS allows you to specify the
ENI_CONFIG_LABEL_DEF environment variable value for worker nodes
This is used to tell Kubernetes to automatically apply the ENIConfig for each Availability Zone
Ref: https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html (step 6(c)).

This PR exposes the env variable in the CNI module in pulumi-eks.

### Related issues (optional)

Addresses issue #410

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+- Support for the ENI_CONFIG_LABEL_DEF environment variable 
+  [#411](https://github.com/pulumi/pulumi-eks/pull/411)
+
 ### Improvements
 
 ## 0.19.3 (Released July 7, 2020)
--- a/nodejs/eks/cni.ts
+++ b/nodejs/eks/cni.ts
@@ -104,6 +104,14 @@ export interface VpcCniOptions {
      * Defaults to 9001.
      */
     eniMtu?: pulumi.Input<number>;
+    /*
+     * Specifies the ENI_CONFIG_LABEL_DEF environment variable value for worker nodes
+     * This is used to tell Kubernetes to automatically apply the ENIConfig for each Availability Zone
+     * Ref: https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html (step 5(c))
+     *
+     * Defaults to the official AWS CNI image in ECR.
+     */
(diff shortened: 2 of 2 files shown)


VRanga000:
closing this, will reopen after it passes internal review.

VRanga000:
The [build error in travis ci ](https://travis-ci.com/github/pulumi/pulumi-eks/builds/177836243) does not seem related to my change, seems like an env var is not getting set, from what I can tell:
```
aws: error: the following arguments are required: value
The command "aws configure set default.aws_secret_access_key $DEFAULT_AWS_SECRET_ACCESS_KEY" failed and exited with 2 during .
```

nodejs/eks/cluster.ts
```diff
@@ -322,10 +322,8 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
     }
 
     if (args.nodeGroupOptions && (
-        args.nodeSubnetIds ||
         args.nodeAssociatePublicIpAddress ||
         args.instanceType ||
-        args.instanceProfileName ||
```
metral, line 328: Why are these vars being removed?

metral:
> The build error in travis ci does not seem related to my change, seems like an env var is not getting set, from what I can tell:

No worries, I'll have to shepherd this PR through since the secrets are not available for public use.

nodejs/eks/cluster.ts
VRanga000: This logic makes it the code throw an exception if any of the vars in the second group are specified when nodeghroupoptions are specified. We do specify nodegroupoptions along with subnetids and instanceprofilename.  I think I was unable to exlude these from the main options and only set them in the nodeGrioupOptions. In any case, this is not related to the eniconfig change tho, so I can remove it from this PR and revisit separately.

VRanga000: Ive removed the unrelated changes.

metral:
Merged this PR in via #412

PR #422, opened 2020-08-14
Bump all pulumi-aws to v3.0.0

stack72:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- Upgrade to `pulumi-aws` v3.0.0
+
 ### Improvements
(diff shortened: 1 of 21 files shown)


metral:
Related: https://github.com/pulumi/home/issues/1045#issuecomment-675077958

> That said, the move from aws v2 -> v3 dropped the roles field from InstanceProfile so this a breaking change in the underlying AWS provider.

The move to AWS 3.0 will be included in a new minor p/eks v0.20.0, but this will cause breaking changes, as the underlying AWS provider API spec changed by removing `roles` (plural) from `aws.InstanceProfile`.

metral (approved):

PR #427, opened 2020-08-18
fix(cluster): add https req timeout & show time left waiting for healthz

### LINKED ISSUE #423: healthz check can take much longer than 300 seconds if the cluster is unreachable
### Problem description

The code below attempts to connect to the cluster API for up to 5 minutes (every 5 seconds up to 60 times). **_But_**, if the cluster is unreachable (because `endpointPrivateAccess: true` is set and the Pulumi client cannot reach the cluster as an example), then this code will actually take much longer than 5 minutes. I believe this is because the `https.request()` call will hang and wait for the operating system timeout to occur before an error is thrown. 

https://github.com/pulumi/pulumi-eks/blob/d44ac5b30ba6d66a022a2e9afc0fda0e85f88b71/nodejs/eks/cluster.ts#L474-L503

### Suggestions for a fix

Change this code to set a timeout on the `https.request()` call (I haven't looked to see if this is possible) or take the overall wall-clock time into account in this code block to give up after 5 minutes has elapsed.

Also, it would be helpful to users to provide more information in the info message that is printed (code below) with something to the effect of `Waiting for cluster endpoint (${i + 1}). Waiting for X more seconds...`

https://github.com/pulumi/pulumi-eks/blob/d44ac5b30ba6d66a022a2e9afc0fda0e85f88b71/nodejs/eks/cluster.ts#L497

metral:
### Proposed changes

fix(cluster): add https req timeout & show time left waiting for healthz

### Related issues (optional)

Fixes #423

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -480,26 +480,33 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
             // endpoint, and verify that it is reachable.
             const healthz = `${clusterEndpoint}/healthz`;
             const agent = createHttpAgent(args.proxy);
-            for (let i = 0; i < 60; i++) {
+            const maxRetries = 60;
+            const reqTimeoutMilliseconds = 1000; // HTTPS request timeout
+            const timeoutMilliseconds = 5000; // Retry timeout
+            for (let i = 0; i < maxRetries; i++) {
                 try {
                     await new Promise((resolve, reject) => {
                         const options = {
                             ...url.parse(healthz),
                             rejectUnauthorized: false, // EKS API server uses self-signed cert
                             agent: agent,
+                            timeout: reqTimeoutMilliseconds,
                         };
                         const req = https
                             .request(options, res => {
                                 res.statusCode === 200 ? resolve() : reject(); // Verify healthz returns 200
                             });
+                        req.on("timeout", reject);
                         req.on("error", reject);
                         req.end();
                     });
+                    pulumi.log.info(`Cluster is ready`, eksCluster, undefined, true);
                     break;
                 } catch (e) {
-                    pulumi.log.info(`Waiting for cluster endpoint (${i + 1})`, eksCluster, undefined, true);
+                    const retrySecondsLeft = (maxRetries - i) * timeoutMilliseconds / 1000;
+                    pulumi.log.info(`Waiting up to (${retrySecondsLeft}) more seconds for cluster readiness...`, eksCluster, undefined, true);
                 }
-                await new Promise(resolve => setTimeout(resolve, 5 * 1000));
+                await new Promise(resolve => setTimeout(resolve, timeoutMilliseconds));
             }
         }
         return clusterEndpoint;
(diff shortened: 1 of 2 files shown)


metral:
Tested using the following setup with the client machine not going through a bastion or in the same VPC as the cluster.

```
const vpc = new awsx.ec2.Vpc("myvpc", {
    tags: { "Name": "myvpc" },
});

const cluster2 = new eks.Cluster("mycluster", {
    vpcId: vpc.id,
    endpointPrivateAccess: true,
    endpointPublicAccess: false,
});
```

See video demo below that shows new request timeout and updated information (the video is sped up mid-way for demo purposes). When the update does time out and eventually error, as would be expected in this setup w/o a bastion in use, it does so shortly after the 5 min endpoint test window, versus the 15+min that it's currently taking.

https://drive.google.com/file/d/1nerx0eBXMCi1dWrI-gUimmO6nRg-WVRM/view

cc @clstokes ^

nodejs/eks/cluster.ts
```diff
@@ -477,26 +477,31 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
...
                             });
+                        req.on("timeout", reject);
                         req.on("error", reject);
                         req.end();
                     });
                     break;
                 } catch (e) {
-                    pulumi.log.info(`Waiting for cluster endpoint (${i + 1})`, eksCluster, undefined, true);
+                    const retrySecondsLeft = (maxRetries - i) * timeoutMilliseconds / 1000;
+                    pulumi.log.info(`Waiting for cluster endpoint (${i + 1}). Waiting up to (${retrySecondsLeft}) more seconds...`, eksCluster, undefined, true);
```
lblackstone, line 502: I think the retry count is somewhat confusing in the current string. Should either remove it if the timeout is sufficient, or show progress towards the max (retry 2/60) or similar.

metral:
@clstokes any feedback?

clstokes:
Does `reqTimeoutMilliseconds = 1000;` mean it will timeout after 1 second?
Is this _too_ low of a value? How does this behave when deploying from the
U.S. to a region on the other side of the world (or vice versa)?


Otherwise, looks good to me.

metral:
> Does `reqTimeoutMilliseconds = 1000;` mean it will timeout after 1 second?

yes

> How does this behave when deploying from the
U.S. to a region on the other side of the world (or vice versa)?

Ran tests without `endpointP*Access` set for each region below from a client in us-east-2 going over the public Internet.

- Brazil - endpoint ready on first attempt for 2 runs
- Frankfurt - endpoint ready on first attempt for 2 runs
- Ireland - endpoint ready on first attempt for first run, and after 10 sec for second run
- Oregon - endpoint ready on after ~1min on two runs, and ready after 10 sec on third run
- Ohio - endpoint ready on first attempt for 2 runs

There doesn't seem to be any issue or delays with the endpoint check for local or int'l regions.

If the cluster's are inaccessible e.g. no bastion with private access on the cluster, they'd timeout and quit shortly after the 5 min. window.

Issue #435, opened 2020-09-08
Migrate build from Travis to GitHub Actions


stack72:
This keeps the system inline with the other providers that we are building at Pulumi

PR #436, opened 2020-09-08
Migrate to GHA

stack72:
Fixes: #435


.github/workflows/master.yml
```diff
@@ -0,0 +1,213 @@
...
+  AWS_REGION: us-west-2
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  GO111MODULE: "on"
+  NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+  NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+  PROVIDER: eks
+  PULUMI_ACCESS_TOKEN: ${{ secrets.PULUMI_ACCESS_TOKEN }}
+  PULUMI_API: https://api.pulumi-staging.io
+  SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+  TRAVIS_OS_NAME: linux
```
metral, line 14: We can rm this envvar from all workflows

Makefile
```diff
@@ -1,24 +1,54 @@
...
-check_clean_worktree:
-	$$(go env GOPATH)/src/github.com/pulumi/scripts/ci/check-worktree-is-clean.sh
-
-# The travis_* targets are entrypoints for CI.
-.PHONY: travis_cron travis_push travis_pull_request travis_api
-travis_cron: default
-travis_push: only_build check_clean_worktree only_test publish_packages
-travis_pull_request: all
-travis_api: all
+PROJECT_NAME := Pulumi Infrastructure Components for AWS
```
metral, line 1:
```suggestion
PROJECT_NAME := Pulumi Amazon Web Services (AWS) EKS Components
```

Makefile
```diff
@@ -1,24 +1,54 @@
...
+
+VERSION         := $(shell pulumictl get version)
+TESTPARALLELISM := 12
```
metral, line 4: We should keep this is an envvar, as we've had to adjust this dial at times with Travis and would be great to carry over to GHA in case we need it too

Makefile
```diff
@@ -1,24 +1,54 @@
...
+	cd $(WORKING_DIR)/nodejs/eks/bin && yarn install
+	yarn link --cwd $(WORKING_DIR)/nodejs/eks/bin
+
+test_nodejs::
+	cd nodejs/eks/examples && go test -v -count=1 -cover -timeout 2h -parallel ${TESTPARALLELISM} .
+
+dev:: lint build_nodejs
+test:: install_nodejs_sdk test_nodejs
+
+#.PHONY: publish_packages
```
metral, line 35: Is this code supposed to be commented or was it meant to be deleted?

dev/build.sh
```diff
@@ -9,7 +9,7 @@ echo `date`
 
 pushd $PROJECT_DIR
 echo "=> Building commit: `git rev-parse --short HEAD`"
-make ensure && make only_build
+make build_nodejs && make install_nodejs_sdk
```
metral, line 12: we'll need to also update the CONTRIBUTING doc with these new commands: https://github.com/pulumi/pulumi-eks/blob/master/CONTRIBUTING.md

metral (changes requested):
LGTM! 🎉 

Couple of small fixes is all :)

Makefile
stack72: So this is difficult - GHA has no idea of EnvVars - it has secrets... We can of course make this a secret and then allow the fine tuning??

metral: That may be worth considering. We have not had to tweak this recently, but it has been a knob we've used on a handful of occasions with Travis

metral (approved):

Makefile
stack72: i will keep an eye on this and see what we need to do here...

PR #450, opened 2020-11-02
fix: correct spelling for encryptRootBlockDevice

### LINKED ISSUE #446: encryptRootBockDevice should be Block instead of Bock
[docs/reference/pkg/nodejs/pulumi/eks/_index.md](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/eks/)
property encryptRootBockDevice should be Block instead of Bock

metral:
### Proposed changes

fix: correct spelling for encryptRootBlockDevice

### Related issues (optional)

Fixes #446

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### Improvements
 
+- fix: correct spelling for encryptRootBlockDevice
+  [#450](https://github.com/pulumi/pulumi-eks/pull/450)
+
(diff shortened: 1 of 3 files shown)


stack72:
this is a breaking change right? Can we do it in a way where we deprecate the old and point to the new and then be able to take them into account?

metral:
@stack72 Thanks for input. PTAL

stack72 (approved):
this is awesome! thanks

PR #472, opened 2020-11-20
Adding the build portion for the provider and schema

stack72:


Makefile
```diff
@@ -3,16 +3,20 @@ PROJECT_NAME := Pulumi Amazon Web Services (AWS) EKS Components
 VERSION         := $(shell pulumictl get version)
 TESTPARALLELISM := 12
 
+PACK			:= eks
```
justinvp, line 6: Nit: odd spacing

justinvp (approved):
Looks great!

Makefile
stack72: ah I see it's tabs not spaces 🙄 - I will fix that up in the next set of work required here :)

justinvp (approved):
Latest commits LGTM

PR #473, opened 2020-11-20
Ensure we package the correct sdk folder based on language

stack72:


stack72:
Self merging here as we need this to be in master for the repository_dispatch to work correctly :/

PR #475, opened 2020-11-23
Schematize remaining surface area

justinvp:
- Expose `VpcCni` resource
- Expose `NodeGroup` component
- Expose `Cluster.storageClasses`
- Expose external types/resources from AWS and Kubernetes
- Expose functions as components:
    - `createManagedNodeGroup` => `ManagedNodeGroup` component
    - `createNodeGroupSecurityGroup` => `NodeGroupSecurityGroup` component
    - `getRoleProvider` => `ClusterCreationRoleProvider` component

I'll open a separate PR with examples (tests).

Fixes https://github.com/pulumi/pulumi-eks/issues/454
Fixes https://github.com/pulumi/pulumi-eks/issues/452

PR #482, opened 2020-12-04
Tracking a change to pulumi-aws that adds support for Node Spot Groups

stack72:
This is a breaking change of sorts but is a required for further node group work

### DIFF
--- a/examples/managed-nodegroups/index.ts
+++ b/examples/managed-nodegroups/index.ts
@@ -44,7 +44,7 @@ const managedNodeGroup2 = eks.createManagedNodeGroup("example-managed-ng2", {
         maxSize: 2,
     },
     diskSize: 20,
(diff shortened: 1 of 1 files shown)


lblackstone (approved):
Change LGTM, but curious where the functional change was made. It seems like we should be able to support either input shape.

stack72:
The change is in pulumi-aws which feeds back to TF

justinvp (approved):
LGTM

I was wondering about `instanceTypes`...

https://github.com/pulumi/pulumi-eks/blob/c0548fd03abe0bb24b9e8d33cb18e09872c82ad0/provider/cmd/pulumi-gen-eks/main.go#L581-L589

I'll update https://github.com/pulumi/pulumi-eks/pull/475 to project it as an array to match the updated AWS signature.

Issue #455, opened 2020-11-04
Add Python, .NET, Go examples


justinvp:

stack72:
Closing this out - we have some examples for these already

PR #484, opened 2020-12-05
Add Python Examples

justinvp:
Initial set... Still a bunch more to add.

Part of https://github.com/pulumi/pulumi-eks/issues/455


justinvp:
@lblackstone, this is ready for review. Easiest to review commit by commit. I'm out until Jan and won't be able to address any feedback, so if it looks OK, I suggest merging and opening issues for any feedback that can be addressed subsequently.

Note that this PR includes `managed-nodegroups-py`, but this example kept timing out in CI, and I didn't get a chance to investigate why, so I've temporarily not included it in the list of tests that run as part of the GHA workflow.

Just so you're aware, each test that we want to run in GHA needs to be listed here (this PR adds some more):

https://github.com/pulumi/pulumi-eks/blob/4498d12d103a76f7b7cf4bd70bb0a061d005dcd6/.github/workflows/run-acceptance-tests.yml#L343-L344

lblackstone (approved):

PR #492, opened 2020-12-18
fix(nodejs): Do not fallback on cluster name when creating node group

### LINKED ISSUE #491: createManagedNodeGroup disable node group auto naming
### Problem description

A managed node group created throuh `createManagedNodeGroup` will not have an auto generated name on AWS.

The issue is the same as #294 but is caused by 

https://github.com/pulumi/pulumi-eks/blob/4498d12d103a76f7b7cf4bd70bb0a061d005dcd6/nodejs/eks/nodegroup.ts#L850

introduced in #302:

> Refactoring the managed node group API to simplify its use (options in role vs role.ARN, make parent arg optional, **and default node group name**).

It causes the node group to be deleted before being replaced in case the config change (e.g. isntanceType). Is there a reason to not auto name it?

### Reproducing the issue

Creating a node group and changing it's instanceType should reproduce the issue.

### Suggestions for a fix

Removing the fallback to `name` should work.

Embraser01:
### Proposed changes

`createManagedNodeGroup` had a fallback on the cluster name for `nodeGroupName` which disabled the auto naming feature which led `deleteBeforeReplace` to be true.

The fix is probably a breaking change as it will update the NodeGroup after an update.

### Related issues

Fixes #491

### DIFF
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -847,7 +847,6 @@ export function createManagedNodeGroup(name: string, args: ManagedNodeGroupOptio
     const nodeGroup = new aws.eks.NodeGroup(name, {
         ...nodeGroupArgs,
         clusterName: args.clusterName || core.cluster.name,
-        nodeGroupName: args.nodeGroupName || name,
         nodeRoleArn: roleArn,
         scalingConfig: pulumi.all([
             args.scalingConfig,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### Improvements
 
+- fix(nodejs): Do not fallback on cluster name when creating node group
+  This is a breaking change as it will recreate the node group on the first deploy
+  [#492](https://github.com/pulumi/pulumi-eks/pull/492)
 - fix: correct spelling for encryptRootBlockDevice
   [#450](https://github.com/pulumi/pulumi-eks/pull/450)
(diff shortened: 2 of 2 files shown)


nodejs/eks/nodegroup.ts
```diff
@@ -847,7 +847,6 @@ export function createManagedNodeGroup(name: string, args: ManagedNodeGroupOptio
     const nodeGroup = new aws.eks.NodeGroup(name, {
         ...nodeGroupArgs,
         clusterName: args.clusterName || core.cluster.name,
-        nodeGroupName: args.nodeGroupName || name,
```
leezen, line 850: Ah, I thought in your original issue you posted, you were proposing something different, which was to not provide `nodeGroupName` if there's no arg passed in, but to keep it if there was. At least this way, this is _less_ breaking for folks who are explicitly passing in the arg.

Embraser01: The `nodeGroupName` will be taken if provided from the `...nodeGroupArgs` spread. Passing the `nodeGroupName` argument will still work as before

leezen: Ah, of course, sorry, yes. Thinking about this more, I guess one way to help folks who do rely on defaulting to the cluster name would be to indicate that they can still do this via a transformation. Let me ask one other person to take a look to make sure I'm not off in any of my thinking here.

lblackstone (approved):
LGTM

Issue #453, opened 2020-11-03
Update aws-node amazon-k8s-cni to v1.7.5


steerben:
### Problem description

Currently only amazon-k8s-cni v1.6.0 is supported.

v1.7.x uses additional volume mounts which can be seen in the updated resources provided by aws here:
```
curl -o aws-k8s-cni.yaml https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/v1.7.5/config/v1.7/aws-k8s-cni.yaml
```
Therefore just updating the image to the newest version alone does not work due to the missing volumes.

### Errors & Logs
Pod is not able to start due to
```
{"level":"error","ts":"2020-11-03T23:10:57.607Z","caller":"aws-k8s-agent/main.go:28","msg":"Initialization failure: open /var/run/aws-node/ipam.json.tmp017019115: no such file or directory"}
```

### Affected product version(s)

"@pulumi/eks": "^0.19.5",

### Suggestions for a fix
Update the cni resource https://github.com/pulumi/pulumi-eks/blob/master/nodejs/eks/cni/aws-k8s-cni.yaml with the newest version supported. 

Retrieve correct resources from
```
curl -o aws-k8s-cni.yaml https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/v1.7.5/config/v1.7/aws-k8s-cni.yaml
```

PR #496, opened 2020-12-22
Upgrade aws-k8s-cni to v1.7.5

stack72:
Fixes: #453

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,17 +2,8 @@
 
 ### Improvements
 
-- Go codegen fixes for external references
-  [#531](https://github.com/pulumi/pulumi-eks/pull/531)
-
-- Documentation addend for build dependencies
-  [#527](https://github.com/pulumi/pulumi-eks/issues/527)
-
-- Expose some more properties to Python, .NET, and Go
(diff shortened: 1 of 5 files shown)


stack72:
This file is taken from https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/v1.7.5/config/v1.7/aws-k8s-cni.yaml

leezen:
I'm not 100% sure how to review this at this point. Is there any divergence between what we want in the manifest vs. what's default to AWS? For a while, there was the divergence with fargate in https://github.com/pulumi/pulumi-eks/pull/291 but that divergence disappeared a year ago in https://github.com/aws/amazon-vpc-cni-k8s/pull/792 -- is there any else materially different that should be preserved?

emiioan:
Any updated on this? We just upgraded our eks clusters to 1.16 and we are using for now a fork and override the yaml file with 1.7.5 version but we will really want to use the upstream librabry. Also double check yaml compute function because for now if you provide  vpcCniOptions:   { warmIpTarget: 2}, env var "WARM_IP_TARGET" is not picked up by cni Daemonset.

leezen (approved):

emiioan:
Hello, when should we expect to have this merged into master branch?

Issue #500, opened 2021-01-05
Update axios to avoid CVE


jkinkead:
The axios npm package is [pinned to 0.19.0](https://github.com/pulumi/pulumi-eks/blob/d1ae8e6373878f476caa64b42e51625fb2e0a741/nodejs/eks/package.json#L17), which has [this CVE](https://github.com/advisories/GHSA-4w2v-q235-vp99) published against it.

axios should be updated to the a version which has a fix (>= 0.21.1).

PR #503, opened 2021-01-08
Bumping version of axios in nodejs sdk to fix security vulnerability

stack72:
Fixes: #500


leezen (approved):

tennix:
When will a newer version be released? @stack72

Issue #443, opened 2020-10-05
Support Service IPv4 range


jwenz723:
### Problem description

Add support for the new `Service IPv4 range` feature of EKS clusters: https://aws.amazon.com/about-aws/whats-new/2020/10/amazon-eks-supports-configurable-kubernetes-service-ip-address-range/

### Errors & Logs

N/A

### Affected product version(s)

N/A

### Reproducing the issue

N/A

### Suggestions for a fix

I don't have any suggestions currently. I'm not familiar with the package internals.

jwenz723:
I also captured this feature request [here](https://github.com/pulumi/pulumi-aws/issues/1148)

matanbaruch:
Any update on this?

matanbaruch:
@stack72

PR #509, opened 2021-01-19
Adding support for `kubernetesServiceIpAddressRange` to `eks.Cluster`

stack72:
Fixes: #443


examples/examples_nodejs_test.go
```diff
@@ -47,6 +47,22 @@ func TestAccCluster(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
+func TestAccKubernetesServiceIPv4RangeForCluster(t *testing.T) {
```
justinvp, line 50: Now that we have Python and .NET hooked up in CI, it'd be nice to add tests for those languages as well (and Go coming soon). I can help with that as a follow-up, if you'd like.

nodejs/eks/cluster.ts
```diff
@@ -475,6 +475,13 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
         }));
     }
 
+    let kubernetesNetworkConfig: pulumi.Output<aws.types.output.eks.ClusterKubernetesNetworkConfig> | undefined;
+    if (args.kubernetesServiceIpAddressRange) {
+        kubernetesNetworkConfig = pulumi.output(args.kubernetesServiceIpAddressRange).apply(
+            serviceIpv4Cidr => <aws.types.output.eks.ClusterKubernetesNetworkConfig>({serviceIpv4Cidr}),
+        );
+    }
```
justinvp, line 483: Technically this should be the _input_ type, and no need to cast inside the apply:


```suggestion
    let kubernetesNetworkConfig: pulumi.Output<aws.types.input.eks.ClusterKubernetesNetworkConfig> | undefined;
    if (args.kubernetesServiceIpAddressRange) {
        kubernetesNetworkConfig = pulumi.output(args.kubernetesServiceIpAddressRange).apply(
            serviceIpv4Cidr => ({ serviceIpv4Cidr })
        );
    }
```

examples/cluster-with-serviceiprange/Pulumi.yaml
```diff
@@ -0,0 +1,3 @@
+name: example-cluster-with-serviceIpRange
```
justinvp, line 1: Super nitpick (sorry!):

```suggestion
name: example-cluster-with-serviceiprange
```

examples/cluster-with-serviceiprange/package.json
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "example-cluster-with-serviceIpRange",
```
justinvp, line 2: Super nitpick:

```suggestion
    "name": "example-cluster-with-serviceiprange",
```

justinvp (approved):
LGTM otherwise.

examples/examples_nodejs_test.go
stack72: I will follow up with these when I get a local dev setup working for EKS in these langs

PR #520, opened 2021-01-27
Adding CI for Go examples

stack72:


.github/workflows/run-acceptance-tests.yml
```diff
@@ -450,7 +539,7 @@ jobs:
         nodeversion:
           - 14.x
         test-name:
-          - ClusterCs
+          - ClusterGo
```
justinvp: I don't think this is right. I think `test-dotnet:` and `test-go:` nodes each need their own `strategy:` section.

viveklak: yeah I just discovered and fixed it. Learning GHA as I go here.

viveklak:
Thanks for taking this on @lblackstone! I was planning on resurrecting this next week.

lblackstone (approved):

Issue #547, opened 2021-03-04
Configuration required for v1.7.5 of CNI


stack72:
As part of the upgrade to CNI 1.7.5, we set the values for the following to be hard coded:

```
            - name: ENABLE_POD_ENI
              value: 'false'
            - name: AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER
              value: 'false'
            - name: AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG
              value: 'false'
            - name: AWS_VPC_K8S_CNI_EXTERNALSNAT
              value: 'false'
```

We should investigate if these values need to be configurable

emiioan:
Hello,

Thanks for working and having cni 1.7.5 merged into master branch. I tested today and all good on crio runtime but it seems to be a permission issue for workers with docker runtime and selinux enabled. 
I managed to fix the permission issues by adding securitycontext: privileged on true to cni container, as having only the NET_ADMIN capability it's not enough. (https://github.com/aws/amazon-vpc-cni-k8s/issues/1291 ). 
For now we are looking to make some changes to selinux to see if we can make it work without privilege enabled. 

Meanwhile when do you thing we can have this yaml under a tag? Should we expected it to be integrated into v.0.23.0?

Thanks

shinebayar-g:
AFAIK `AWS_VPC_K8S_CNI_EXTERNALSNAT` needs to be `true` when multiple CNIs deployed into cluster.

stack72:
I think after what @shinebayar-g is saying, we need to actually implement these specific configuration options

PR will come shortly

emiioan:
Good, also please check this issue aws/amazon-vpc-cni-k8s#1291, I disabled selinux on ec2 instance node with docker container runtime and I was not able to make it working. As soon as I add privileged: true it works.

stack72:
Hey @emiioan 

so I am not sure from that issue where this needs to go in the yaml - can you give me a hint?

P.

stack72:
ok, I see it - so we should set this as `privileged: false` by default with the ability to override it, right?

emiioan:
Yes, you can add it here https://github.com/pulumi/pulumi-eks/blob/master/nodejs/eks/cni/aws-k8s-cni.yaml#L132  and able to overwrite it will be perfect: 

```
          securityContext:
	    privileged: false
            capabilities:
              add:
                - NET_ADMIN
```

stack72:
[...]

PR #549, opened 2021-03-05
Ensure new CNI configuration options are able to be controlled

stack72:
Fixes: #547

This is continuing the work on the CNI 1.7.5 upgrade


nodejs/eks/cmd/provider/cni.ts
```diff
@@ -37,6 +37,10 @@ interface VpcCniInputs {
     eniConfigLabelDef?: string;
     pluginLogLevel?: string;
     pluginLogFile?: string;
+    enablePodEni?: boolean;
+    cniConfigureRpfilter?: boolean;
+    cniCustomNetworkCfg?: boolean;
+    cniExternalSnat?: boolean;
```
justinvp, line 43: Are these meant to be exposed so users can set them? If so, we need to add them here:

https://github.com/pulumi/pulumi-eks/blob/30fafe512b3282afb6f63c9302d94e46115755e7/nodejs/eks/cni.ts#L20

And here:

https://github.com/pulumi/pulumi-eks/blob/30fafe512b3282afb6f63c9302d94e46115755e7/provider/cmd/pulumi-gen-eks/main.go#L1395-L1397

And rebuild.

stack72: Great catch! All added and exposed in the schema as well!

justinvp (approved):

PR #559, opened 2021-04-15
make the providerCredentialOpts error clearer

### LINKED ISSUE #558: The Profile error wording is confusing
If the user is using AWS profiles with the provider, and they don't set it during the creation, they get the following error message:

```
Error: providerCredentialOpts and AWS config setting aws:profile must be set together: Error: providerCredentialOpts and AWS config setting aws:profile must be set together
```

This error isn't terribly clear for the user. We should consider rewording it

## Expected behavior
We give the user an error that explains what they have to do

jaxxstorm:
Fixes #558

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -1379,15 +1379,15 @@ export class Cluster extends pulumi.ComponentResource {
         // Check that AWS provider credential options are set for the kubeconfig
         // to use with the given auth method.
         if (opts?.provider && !args.providerCredentialOpts) {
-            throw new Error("It looks like you're using an explicit AWS provider. Please specify this provider in providerCredentialOpts. For more details, see: https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/eks/#ClusterOptions-providerCredentialOpts")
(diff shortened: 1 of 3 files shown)


clstokes (approved):
Looks great - definite improvement on usability and _actionability_ of these messages for the end user.

stack72 (approved):

nodejs/eks/cluster.ts
```diff
@@ -1379,15 +1379,15 @@ export class Cluster extends pulumi.ComponentResource {
         // Check that AWS provider credential options are set for the kubeconfig
         // to use with the given auth method.
         if (opts?.provider && !args.providerCredentialOpts) {
-            throw new Error("providerCredentialOpts and an AWS provider instance must be set together");
+            throw new Error("It looks like you're using an explicit AWS provider. Please specify this provider in providerCredentialOpts. For more details, see: https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/eks/#ClusterOptions-providerCredentialOpts")
```
lukehoban, line 1382: Those are node.js docs - but this error will show up in all languages - right? Can we link to language-agnostic resource docs for this?

jaxxstorm: ah this is a good point, I'll remove this for now until we figure out a way to make the link language agnostic.

PR #579, opened 2021-05-18
fix: typo in error msg

floknip:
### Proposed changes

Added missing word "in" in the error message

Should it be something like "Please specify the profile in providerCredentialOpts" instead of
"Please specify this profile providerCredentialOpts."?

### Related issues (optional)

 #578


lblackstone (approved):
Thanks for the PR!

Issue #566, opened 2021-04-20
Issues creating nodegroup, Cryptic error message


andrewhertog:
## Expected behavior

Successfully create a nodegroup attached to a cluster

## Current behavior

Erroring with the following message:
```
    error: program failed: marshaling properties: awaiting input property cluster: missing one or more required arguments
    exit status 1

    error: an unhandled error occurred: program exited with non-zero exit code: 1
```
## Steps to reproduce

run `pulumi up` with the following code snippet:
```go

    cluster, err := eks.NewCluster(ctx, stack, &eks.ClusterArgs{
      SkipDefaultNodeGroup: pulumi.Bool(true),
      VpcId: pulumi.String(vpcId),
      SubnetIds: pulumi.StringArray{
        pulumi.String("subnet-<redacted>"),
        pulumi.String("subnet-<redacted>"),
        pulumi.String("subnet-<redacted>"),
      },
      ProviderCredentialOpts: &eks.KubeconfigOptionsArgs{
        ProfileName: pulumi.String(awsProfileName),
      },
      EndpointPrivateAccess: pulumi.Bool(true),
      EndpointPublicAccess: pulumi.Bool(false),
      ClusterSecurityGroup: clusterSecurityGroup,
      InstanceProfileName: instanceProfile.Name,
      NodeAssociatePublicIpAddress: pulumi.Bool(false),
    })
    if err != nil {
      return err
    }

_, err = eks.NewNodeGroup(ctx, fmt.Sprintf("%s-%s",stack,"systemNg") , &eks.NodeGroupArgs{
      Cluster: cluster.Core,
      InstanceProfile: instanceProfile,
      DesiredCapacity: pulumi.Int(1),
      MinSize: pulumi.Int(1),
      MaxSize: pulumi.Int(2),
      InstanceType: pulumi.String("t2.medium"),
      Labels: pulumi.StringMap{
        "ondemand": pulumi.String("true"),
        "nodegroup": pulumi.String("system"),
        "zone": pulumi.String("default"),
      },
      NodeSubnetIds: pulumi.StringArray{
        pulumi.String("subnet-<redacted>"), // same subnets as cluster
        pulumi.String("subnet-<redacted>"),
        pulumi.String("subnet-<redacted>"),
      },
    },pulumi.Parent(cluster))
```

## Context (Environment)

It created the cluster succesfully, but failed with the obscure message posted above

Running on Mac Catalina
[...]

PR #590, opened 2021-05-28
[nodejs] Add resource registration for VpcCni

justinvp:
The `VpcCni` resource can be passed back to the TypeScript implementation from other languages as a resource reference, and needs to be instantiated as an instance of `VpcCni` when deserialized so it can be used in a `dependsOn`.

Part of #566

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ provider::
 
 build_nodejs:: VERSION := $(shell pulumictl get version --language javascript)
 build_nodejs::
-	rm -rf nodejs/eks/bin
+	rm -rf nodejs/eks/bin/*
 	cd nodejs/eks && \
 		yarn install && \
 		yarn run tsc && \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -20,7 +20,7 @@ provider::
 
 build_nodejs:: VERSION := $(shell pulumictl get version --language javascript)
 build_nodejs::
-	rm -rf nodejs/eks/bin
+	rm -rf nodejs/eks/bin/*
```
justinvp, line 23: Otherwise, deleting the whole `bin` dir was preventing subsequent `yarn link` calls in the `install_nodejs_sdk` target from succeeding after a prior `yarn link`.

lblackstone (commented):
Changes LGTM. Do we have any tests that cover this case?

lblackstone (approved):

justinvp:
> Do we have any tests that cover this case?

We don't and I'm not sure we can yet because I hit another issue after this is fixed. Will open a new issue to track it.

lukehoban:
@justinvp Is this still needed?

justinvp:
> Is this still needed?

Yes.

PR #617, opened 2021-09-02
Support for setting init container options and per-pod security groups

alex-hunt-materialize:
### Proposed changes

Supports setting the init container image and DISABLE_TCP_EARLY_DEMUX
env var.

This allows users to upgrade their version of the CNI, and to support
per-pod security groups.

## Related issues

Fixes https://github.com/pulumi/pulumi-eks/issues/619
Fixes https://github.com/pulumi/pulumi-eks/issues/620

Relates to, but doesn't solve: https://github.com/pulumi/pulumi-eks/issues/622


alex-hunt-materialize:
I unfortunately am unable to test this change, since I can't figure out how to install it and use it with a python virtualenv (all our pulumi uses python). Any advice on how to do that would be greatly appreciated.

The tests seem to run locally, but all get marked as failed due to being unable to write the results to S3. I have no idea why local tests should write their output to S3. That seems like an unrelated bug.

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1479,8 +1477,13 @@ func vpcCniProperties(kubeconfig bool) map[string]schema.PropertySpec {
...
 			TypeSpec: schema.TypeSpec{Type: "boolean"},
-			Description: "Specifies whether to allow IPAMD to add the `vpc.amazonaws.com/has-trunk-attached` label to" +
-				"the node if the instance has capacity to attach an additional ENI. Default is `false`.",
+			Description: "Specifies whether to allow IPAMD to add the `vpc.amazonaws.com/has-trunk-attached` label to " +
+				"the node if the instance has capacity to attach an additional ENI. Default is `false`. " +
+				"If using liveness and readiness probes, you will also need to disable TCP early demux.",
+		},
+		"disableTcpEarlyDemux": {
+			TypeSpec:    schema.TypeSpec{Type: "boolean"},
+			Description: "Allows the kubelet's liveness and readiness probes to connect via TCP when pod ENI is enabled. This will slightly increase local TCP connection latency.",
```
lukehoban, line 1486: It's a little unfortunate, but for consistency, lets line wrap this similar to other descriptions in this file.  (A little surprised we aren't getting a int failure on this).

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1426,13 +1426,6 @@ func vpcCniProperties(kubeconfig bool) map[string]schema.PropertySpec {
 				"same VPC as your control plane resources) that are independent of your cluster's " +
 				"`resourcesVpcConfig`.\n\nDefaults to false.",
 		},
-		"externalSnat": {
-			TypeSpec: schema.TypeSpec{Type: "boolean"},
-			Description: "Specifies whether an external NAT gateway should be used to provide SNAT " +
-				"of secondary ENI IP addresses. If set to true, the SNAT iptables rule and off-VPC " +
-				"IP rule are not applied, and these rules are removed if they have already been " +
-				"applied.\n\nDefaults to false.",
-		},
```
lukehoban, line 1435: Should we deprecate this instead of removing it in case users are depending on it?

Probably not related to this PR - but the inconsistency that some of these are `cni...` and others aren't is a little odd.  It actually feels like we most likely want to standardize on *not* including the `cni` prefix (since these are all properties of the VpcCni anyway.  In which case, it might make sense to deprecate the other one instead?  (And file an issue to track aligning the remaining too `cni...` property names with the naming scheme?

lukehoban (commented):
Nice - thanks!

On test coverage -  it would be really nice to add a simple test that uses a newer CNI than the default - just to lock in that that works.  Perhaps alongside `examples/cluster` but configured with the new input parameter supported here?

provider/cmd/pulumi-gen-eks/main.go
viveklak: +1 on deprecating instead. If you want to back this change out and open an issue about the duplication instead, we can address that separately and in the process standardize the properties as Luke suggested.

viveklak (commented):
Thanks for taking this on! Looks great. Could you take a look at a couple of the suggestions add a test as Luke mentioned?

provider/cmd/pulumi-gen-eks/main.go
alex-hunt-materialize: done

provider/cmd/pulumi-gen-eks/main.go
alex-hunt-materialize: I've backed it out, and am filing a new issue for this.

alex-hunt-materialize:
I've backed out the `externalSnat` change, added `image` and `initImage` `VpcCniOptions` to `examples/cluster`, and filed https://github.com/pulumi/pulumi-eks/issues/624 to later fix the `externalSnat` issue.

viveklak (approved):

examples/cluster/index.ts
```diff
@@ -24,6 +24,10 @@ const cluster2 = new eks.Cluster(`${projectName}-2`, {
         "audit",
         "authenticator",
     ],
+    vpcCniOptions: {
+        image: "602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.9.0",
+        initImage: "602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni-init:v1.9.0",
+    },
```
viveklak, line 30: Would be nice to ensure the VPC CNI options were actually honored as well. The validation could be added [here](https://github.com/pulumi/pulumi-eks/blob/master/examples/examples_nodejs_test.go#L42).

viveklak:
> This also fixes a duplicate argument, where `externalSnat` and `cniExternalSnat` would potentially fight over setting the `AWS_VPC_K8S_CNI_EXTERNALSNAT` environment variable. I can find no reason for both to exist, and `cniExternalSnat` seemed more complete, so I kept that one.

Could you remove this from the PR body since it was backed out?

examples/cluster/index.ts
viveklak: > Would be nice to ensure the VPC CNI options were actually honored as well. The validation could be added [here](https://github.com/pulumi/pulumi-eks/blob/master/examples/examples_nodejs_test.go#L42). You might have to export some additional information, e.g. the spec from index.ts.

alex-hunt-materialize: My typescript/javascript skills are pretty much non-existent. I'm unsure how to export these.
I've tried:
```
export const image2 = cluster2.core.vpcCni.image;
export const initImage2 = cluster2.core.vpcCni.initImage;
```
but I get:
```
    index.ts(36,45): error TS2339: Property 'image' does not exist on type 'VpcCni'.
    index.ts(37,49): error TS2339: Property 'initImage' does not exist on type 'VpcCni'.
```

alex-hunt-materialize:
> > This also fixes a duplicate argument, where externalSnat and cniExternalSnat would potentially fight over setting the AWS_VPC_K8S_CNI_EXTERNALSNAT environment variable. I can find no reason for both to exist, and cniExternalSnat seemed more complete, so I kept that one.
> 
> Could you remove this from the PR body since it was backed out?

Done

benesch:
So, @viveklak, I'm afraid we don't have the expertise in house at @MaterializeInc to get the test suite properly amended for this change. This package in particular is really tricky to develop on locally!

The good news is that I managed to ship a forked of `pulumi-eks`: https://github.com/MaterializeInc/pulumi-eks/commit/d4b05f9370462c1a64a0e582e36ef2c1c0a04e77#diff-5d43b9a26b3f3c15075a30aed416693a0ccd1d793cce198952bc4cf0693ca63c. That relieves the time pressure on our end. We'd still love to see this merged, because we don't want to be maintaining a fork long term, but I think we've probably pushed this PR as far as we can.

viveklak:
> So, @viveklak, I'm afraid we don't have the expertise in house at @MaterializeInc to get the test suite properly amended for this change. This package in particular is really tricky to develop on locally!
> 
> The good news is that I managed to ship a forked of `pulumi-eks`: [MaterializeInc@d4b05f9#diff-5d43b9a26b3f3c15075a30aed416693a0ccd1d793cce198952bc4cf0693ca63c](https://github.com/MaterializeInc/pulumi-eks/commit/d4b05f9370462c1a64a0e582e36ef2c1c0a04e77#diff-5d43b9a26b3f3c15075a30aed416693a0ccd1d793cce198952bc4cf0693ca63c). That relieves the time pressure on our end. We'd still love to see this merged, because we don't want to be maintaining a fork long term, but I think we've probably pushed this PR as far as we can.

No worries and glad this worked for you. I will verify this on my end tomorrow and merge, then follow up with any test related updates. Thanks for taking the time to make the change!

alex-hunt-materialize:
@viveklak Any update on this? It's been working great for us on our fork for close to a month now. We'd love to point back at the upstream soon, but we can't until this is merged.

viveklak:
I have confirmed the changes work. However, the command dispatch workflow used to validate external PRs seems to be currently broken due to schema change detection not being able to get access to the originating workflow. We are looking to fix this but I would like to get this change in, so I opened https://github.com/pulumi/pulumi-eks/pull/631 which pulls in these changes and adds a test to validate. I'd like to close this in favor of https://github.com/pulumi/pulumi-eks/pull/631. Huge thanks for the contribution @alex-hunt-materialize and team!

lukehoban:
Thanks @alex-hunt-materialize! 🚀

alex-hunt-materialize:
Thank you for merging all for merging it :)

viveklak:
🙌🏾  - I will cut a release shortly.

PR #625, opened 2021-09-22
Fix a few typos

cnunciato:
There's probably more to be done than just this, so any guidance would be appreciated. Thanks!

Fixes https://github.com/pulumi/pulumi-hugo/issues/621.


lblackstone (approved):

lblackstone:
The changes LGTM, but I'm not sure if we need to also regenerate SDKs? @justinvp

ericrudder:
If you're editing anyway, the (c) should be 2021 ... the other text looks
sane (a lot more cases than i remember!)

susanev:
@justinvp @chrsmith can this be merged?

PR #627, opened 2021-10-01
Set the npm version to the correct package

iwahbe:
We have it pointing to our aws provider before.

### Proposed changes

Point npm badge to correct package.


justinvp (approved):

justinvp:
Actually, you'll need to re-generate the SDKs because the `python` step copies the README.

PR #639, opened 2021-11-29
NodeGroups accept strings as InstanceTypes

jazzdan:
### Proposed changes

The instance type parameter to NodeGroups should accept strings, in addition to the InstanceType enum, because the enum could be out of date. This is similar to what the [EC2 object itself accepts in Pulumi](https://github.com/pulumi/pulumi-aws/blob/8cdca73c8c37d91ad9bcdc2e6b616926df730ea2/sdk/nodejs/ec2/instance.ts#L709).


leezen (approved):

leezen:
Thanks for fixing!

jazzdan:
@leezen thanks for merging! :D

PR #653, opened 2022-01-27
Widen type for `instanceType` to allow `string`

leezen:
Allow `string` in addition to the `InstanceType` enum for `instanceType`

Related: https://pulumi-community.slack.com/archives/CRH5ENVDX/p1643301919171600

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -960,7 +960,7 @@ export interface ClusterOptions {
     /**
      * The instance type to use for the cluster's nodes. Defaults to "t2.medium".
      */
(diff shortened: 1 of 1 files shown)


danielrbradley (approved):

leezen:
Oddly, it looks like both the PR build and previous `master` build failed. Did something change in `tsc`?

leezen:
Actually, more likely is there was a `@pulumi/pulumi` release that went out -- looking into that.

leezen:
Indeed, seems to be a result of the change to `check` in https://github.com/pulumi/pulumi/pull/8631

PR #678, opened 2022-04-08
Update apiversion for eks 1.22

pabardina:
### Proposed changes

With the recent EKS release 1.22, we can't deploy a cluster because of an a deprecated api version.

```
no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1"
```
The apiextensions.k8s.io/v1beta1 API version of CustomResourceDefinition is no longer served as of v1.22.


pabardina:
Ok, another PR already exists

PR #690, opened 2022-05-05
Temporarily pin kubectl version

viveklak:
### Proposed changes
Pin the kubectl version to 1.23.6 to temporarily workaround https://github.com/aws/aws-cli/issues/6920 to fix CI runs.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -282,6 +282,13 @@ jobs:
         with:
           node-version: ${{matrix.nodeversion}}
           registry-url: https://registry.npmjs.org
+      - name: Install Kubectl
+        run: |
+          # https://github.com/aws/aws-cli/issues/6920
+          # curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
+          curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.23.6/bin/linux/amd64/kubectl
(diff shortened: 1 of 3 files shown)


guineveresaenger (approved):
reminder to create an issue to track unpinning

stack72 (commented):
Do we need this in master and release?

viveklak:
> 

Sounds like we do, yes. I initially grepped for kubectl but it looks like these workflows do run the tests as well and might be using the base image's kubectl. Which according to this is 1.24.0: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md

yasinntza:
@viveklak  how can I fix this issue for myself. I am still getting this error. I added github workflow and error image

Screen Shot 2022-05-09 at 19 20 22

image
?

viveklak:
@yasinntza the fundamental problem here was that aws-cli was lagging on picking up the deprecation of the above api version. That seems to have just been fixed: https://github.com/aws/aws-cli/issues/6920. I will be looking to revert this PR shortly.

AaronFriel:
Hey, thanks for the issue report. 1.24 deprecations are definitely an issue people can face with Kubernetes, but I think I'm missing a couple pieces of the puzzle. I have a couple questions:

1. Are you using a self-hosted runner or one of GitHub's runners (GitHub defaults to their virtual environments.) If the latter, it looks like their runners recently updated to 1.24.0, which would be out of our hands. https://github.com/actions/virtual-environments

2. Are you executing any kubectl commands directly or providers other than EKS?

PR #700, opened 2022-05-09
Add IP family support for cluster network config

### LINKED ISSUE #698: Add support for ClusterKubernetesNetworkConfig-> ipFamily to support IPv6
We need to add support for [ClusterKubernetesNetworkConfig-> ipFamily](https://www.pulumi.com/registry/packages/aws/api-docs/eks/cluster/#clusterkubernetesnetworkconfig)

_Originally posted by @davidroth in https://github.com/pulumi/pulumi-eks/issues/675#issuecomment-1106453162_

viveklak:
### Related issues (optional)
Fixes #698

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -483,9 +483,9 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
     }
 
     let kubernetesNetworkConfig: pulumi.Output<aws.types.input.eks.ClusterKubernetesNetworkConfig> | undefined;
-    if (args.kubernetesServiceIpAddressRange) {
-        kubernetesNetworkConfig = pulumi.output(args.kubernetesServiceIpAddressRange).apply(
-            serviceIpv4Cidr => ({ serviceIpv4Cidr }),
+    if (args.kubernetesServiceIpAddressRange || args.kubernetesServiceIpFamily ) {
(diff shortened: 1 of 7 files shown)


nodejs/eks/cluster.ts
```diff
@@ -483,9 +483,9 @@ export function createCore(name: string, args: ClusterOptions, parent: pulumi.Co
     }
 
     let kubernetesNetworkConfig: pulumi.Output<aws.types.input.eks.ClusterKubernetesNetworkConfig> | undefined;
-    if (args.kubernetesServiceIpAddressRange) {
-        kubernetesNetworkConfig = pulumi.output(args.kubernetesServiceIpAddressRange).apply(
-            serviceIpv4Cidr => ({ serviceIpv4Cidr }),
+    if (args.kubernetesServiceIpAddressRange || args.kubernetesServiceIpFamily ) {
```
stack72, line 486: do these not need to be different applies @viveklak ? What happens if a user doesn't set the serviceIpAddressRange but does specify the serviceIpFamily (and vice versa), we will try and set the other value

viveklak: Yeah this seems buggy but I have lost a bunch of context here. I would imagine you would want to include the ipv6 cidr for ipv6 family. I have lost a bunch of context here. @roothorp mentioned he would take this over and get it in.

Issue #727, opened 2022-06-30
managedPolicyArns should not update on clusters older than April 16th 2020


rshade:
### What happened?

#670 causes a regression on clusters older than April 16th 2020, by removing a necessary policy from the cluster causing nodeGroups not to spin up. 

### Steps to reproduce

Have a cluster older than April 16th 2020 and bump the eks provider. 

### Expected Behavior

It should not remove the policy on older clusters. 

### Actual Behavior

It removes the policy on older clusters. 

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #730, opened 2022-07-06
Export the ARN for the IAM Role attached to the EKS Cluster

stack72:
Fixes: #727

In #670, we removed the deprecated IAM Role Policy `AmazonEKSServicePolicy`.
Unfortunately, this has broken clusters older than 16th April 2020

Rather than adding this policy back by default, we now expose the
ARN of the IAM Role for the cluster so that a user can do as follows:

```
const cluster1 = new eks.Cluster(`${projectName}-1`);

const iamRole = cluster1.core.clusterIamRole;

const rpa = new aws.iam.RolePolicyAttachment("rpa", {
  role: iamRole.name,
  policy: "arn:aws:iam::aws:policy/AmazonEKSServicePolicy"
})
```


rshade (approved):
LGTM

lblackstone (approved):

PR #746, opened 2022-07-21
Remove cluster provider from schema

### LINKED ISSUE #602: Remove `provider` from cluster result
Until https://github.com/pulumi/pulumi/issues/7012 is addressed, the provider returned in the cluster result is unusable without creating resources within an apply. Users should just create a provider from the returned kubeconfig instead.

Until we can address the underlying issue, we should just remove this field and consider removing the AWS provider as well, although it is not clear if there is a comparable work around.

jaxxstorm:
fixes #602

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
-(None)
+- Remove cluster provider from schema
+  [#746](https://github.com/pulumi/pulumi-eks/pull/746)
 
 ## 0.41.2 (Released Jul 12, 2022)
 - Allow removal of default Kubernetes addons
--- a/dotnet/Cluster.cs
+++ b/dotnet/Cluster.cs
@@ -69,12 +69,6 @@ namespace Pulumi.Eks
         [Output("nodeSecurityGroup")]
         public Output<Pulumi.Aws.Ec2.SecurityGroup> NodeSecurityGroup { get; private set; } = null!;
 
-        /// <summary>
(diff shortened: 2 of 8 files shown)


viveklak:
I believe we also want to nuke it from the schema which should drop it from the sdk at the least: https://github.com/pulumi/pulumi-eks/blob/master/provider/cmd/pulumi-gen-eks/main.go#L164-L166

jaxxstorm:
@viveklak I actually think nuking it from the schema is a better way, so I went ahead and did that

stack72 (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Unreleased
-(None)
+- Remove cluster provider from schema
```
viveklak, line 2: Maybe we need a bit more explanation for the result of the change? Something along the lines of:
```suggestion
- Due to https://github.com/pulumi/pulumi/issues/7012 including the provider in the generated SDK never really worked. This is removed now. Note - existing uses in Nodejs should not be affected.
```

lukehoban:
This is a breaking change - right?  Building on @viveklak's comment - probably worth adding to the CHANGELOG at least here.

stack72:
> This is a breaking change - right? Building on @viveklak's comment - probably worth adding to the CHANGELOG at least here.

This is breaking for non-TS SDKs - but the code never actually worked in those providers which is why we chose to take it on - I will make sure we prefix this as a breaking change

Issue #526, opened 2021-01-28
Cleanup: Move dotnet, nodejs, and python code under sdk directory


justinvp:
So that these are next to the generated go code, and for consistency with all our other provider repos.

PR #748, opened 2022-07-27
[internal] Reorganising the generated sources into the sdk folder

stack72:
Fixes: #526


roothorp (approved):
Good to bring repo layouts in line with the standard!

Issue #638, opened 2021-11-29
Support bottlerocket AMI Family on EKS managed node group


ZhiHanZ:
## Hello!

AWS 
AWS recently provided support to run bottlerocket natively on EKS managed node group. https://aws.amazon.com/blogs/containers/amazon-eks-adds-native-support-for-bottlerocket-in-managed-node-groups/ 
Does pulumi have plan to support it?
- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

mikhailshilkov:
@ZhiHanZ Thank you for your question.

If I understand correctly, the it should be possible with the Pulumi AWS provider already by setting the right AMI and maybe user data. Since you opened the issue in this repo, I guess you are asking whether the `@pulumi/eks` package supports it.

I wonder if you tried setting AMI on the node group? Do you know what else does it take to enable bottlerocket?

ZhiHanZ:
Hi @mikhailshilkov , thank you for response. 
Here I think currently pulumi/eks only support AL2_x86_64 | AL2_x86_64_GPU | AL2_ARM_64 for amiType (https://github.com/pulumi/pulumi-eks/blob/master/sdk/go/eks/managedNodeGroup.go#L47),
and in aws official doc (https://docs.aws.amazon.com/eks/latest/APIReference/API_Nodegroup.html), they also support 
BOTTLEROCKET_ARM_64 | BOTTLEROCKET_x86_64 amiType, so I am wondering about whether here is a plan to support it.

mikhailshilkov:
Good point. However, this list is just a textual description which shouldn't restrict which values you can pass there. We should probably adjust it [here](https://github.com/pulumi/pulumi-eks/blob/e03751425604b5193c00cd2f8858d9b16f7901f0/provider/cmd/pulumi-gen-eks/main.go#L621-L623) but meanwhile you could try passing a bottlerocket value. I'm curious if it would just work.

ronaldour:
It actually restricts the values that can be used. It fails with error `expected ami_type to be one of [AL2_x86_64 AL2_x86_64_GPU AL2_ARM_64 CUSTOM], got BOTTLEROCKET_x86_64. Examine values at 'NodeGroup.AmiType'.` 
I think this should be updated to support those AMI types, otherwise we need to create a custom launch template to use Bottlerocket AMIs

michael-barker:
[...]

PR #755, opened 2022-08-01
managedNodegroup/amiType: Update docs to remove suggest version restriction

stack72:
Fixes: #638

the case due to the upgraded pulumi-aws dependency. We have changed the
API Docs to suggest that we support what AWS support as we use their SDK


rshade (approved):
LGTM

roothorp (approved):
Nice!

Issue #763, opened 2022-08-17
Consider switching from LaunchConfigurations to LaunchTemplates


jaxxstorm:
This may be a part of #750 but I thought I'd file it as a separate issue.

We currently use LaunchConfigurations for nodegroups: https://github.com/pulumi/pulumi-eks/blob/47faf9fdf5fc42e5652d2535a826c2c9c35fd64e/nodejs/eks/nodegroup.ts#L549

AWS is no longer adding support for features to launch configuration, and suggests using LaunchTemplates instead:
https://aws.amazon.com/blogs/compute/amazon-ec2-auto-scaling-will-no-longer-add-support-for-new-ec2-features-to-launch-configurations/

exposing LaunchTemplates will give us a whole bunch more flexibility to the way nodes are launch, like allowing mixedInstance types, spot instances and a whole host of other awesome stuff.

stack72:
@roothorp I suggest we look at this as part of the 1.0 work

Issue #750, opened 2022-07-29
Investigate the options of removing the need for CloudFormation in NodeGroups


stack72:
The cloudformation template is for the creation of an Autoscaling group - we can handle this via pulumi-aws ASG and be able to expose all of the options available in the ASG. We can make this property can either pass in an a user specified ASG or they can build the properties of the ASG automatically

@roothorp / @stack72 to come up with a design doc around this

jaxxstorm:
A little bit of backstory here.

My recollection for us implementing the nodegroups as a cloudformation stack was because it was the only way of refreshing an ASG without interrupting the nodes.

However, `InstanceRefresh` properties have been added to the ASG resource: https://www.pulumi.com/registry/packages/aws/api-docs/autoscaling/group/#instance_refresh_python

So this isn't really necessary

PR #766, opened 2022-08-25
Add NodeGroupV2

roothorp:
NodeGroupV2 is an alternative method for creating a node group, no longer using cfn  and moving from launch configurations to launch templates.

Fixes: https://github.com/pulumi/pulumi-eks/issues/750
Fixes: https://github.com/pulumi/pulumi-eks/issues/763


stack72 (commented):
@roothorp I have a couple of small questions / issues - would love @danielrbradley to give this a once over as well :) But this is looking great!!

I do think we will need to update the automated test to make sure both NodeGroup and NodeGroupV2 works

roothorp:
Hopefully resolved most of this @stack72 - I'd appreciate the oppurtunity to catch up and make sure the schema is properly working. Also, getting the tests to understand NodeGroupV2 looks a bit fiddly so I'll take a look at that tomorrow.

danielrbradley (approved):

PR #776, opened 2022-09-08
Modify Makefile to use PKG

### LINKED ISSUE #728: Change EKS repo to use PKG
We want to take advantage of a single binary to run this package so we should follow the same approach as per https://github.com/pulumi/pulumi-aws-apigateway/pull/48

This should only affect the packaging of the provider and not the packaging of the existing nodejs SDK for now. The nodejs SDK change will happen as a breaking change in v2.0

roothorp:
Fixes #728 

This change should be considered the first step in updating the shipping process for EKS, followed up with seperating out the node SDK from the provider. There is a draft for this in #777, but this will be a breaking change.


nodejs/eks/cmd/provider/cni.ts
```diff
@@ -167,7 +167,7 @@ function applyVpcCniYaml(args: VpcCniInputs) {
     // Check to ensure that a compatible kubectl is installed, as we'll need it in order to deploy k8s resources below.
     assertCompatibleKubectlVersionExists();
 
-    const yamlPath = path.join(__dirname, "..", "..", "cni", "aws-k8s-cni.yaml");
+    const yamlPath = path.join(__dirname, "../../cni/aws-k8s-cni.yaml");
```
roothorp, line 170: This feels like it defeats the point of using `join`, but pkg seems to want a specifically two-argument `join` in order to pick it up as an asset (same for the dashboard below).

https://github.com/vercel/pkg#detecting-assets-in-source-code

roothorp:
These tests are failing because we're hitting a limit for creating SecurityGroups.

nodejs/eks/cmd/provider/cni.ts
danielrbradley: Hmm, we use `path.join` to make this cross-platform. Using literal `/` won't work on Windows. Although, I guess this doesn't matter because this won't actually be run on Windows because it will be run at build-time only.

danielrbradley (approved):
I think this all looks good. 

Can we just run the `master` workflow from the PR to test doing a pre-release or do we need to merge first to ensure the dist steps work without hiccups?

Issue #800, opened 2022-10-25
`pulumi-eks` always download the latest `kubernetes` plugin 


aureq:
### What happened?

([Internal thread](https://pulumi.slack.com/archives/C0435R1H5FC/p1666677501385389?thread_ts=1666675408.522739&cid=C0435R1H5FC))

In Python, I have the following `requirements.txt` with all versions pinned, `pulumi-eks` seems to be dragging the latest versions of the `aws` and `kubernetes` plugins.

```
pulumi-aws==5.16.0
pulumi-eks==0.40.0
pulumi-kubernetes==3.21.4
```

After running `pip3 install -r requirements.txt`, I then run `pulumi plugin ls`.

```
❯ pulumi plugin ls
NAME        KIND      VERSION  SIZE    INSTALLED       LAST USED
aws         resource  5.18.0   427 MB  38 seconds ago  38 seconds ago
aws         resource  5.16.0   419 MB  1 minute ago    1 minute ago
eks         resource  0.40.0   200 MB  36 seconds ago  48 seconds ago
kubernetes  resource  3.22.0   84 MB   43 seconds ago  43 seconds ago
kubernetes  resource  3.21.4   83 MB   1 minute ago    1 minute ago

TOTAL plugin cache size: 1.2 GB
```

As shown above, both `aws` and `kubernetes` have 2 versions of their plugin installed.

If I run the same experiment but with `pulumi-eks==0.40.0` commented, running `pulumi plugin ls` shows

```
❯ pulumi plugin ls
NAME        KIND      VERSION  SIZE    INSTALLED       LAST USED
aws         resource  5.16.0   419 MB  6 seconds ago   6 seconds ago
kubernetes  resource  3.21.4   83 MB   30 seconds ago  30 seconds ago

TOTAL plugin cache size: 502 MB
```

Furthermore, if I delete the more recent plugins using `pulumi plugin rm resource kubernetes 3.22.0 --non-interactive --yes`, and the running `pulumi preview`. It appears the CLI download again the removed plugin.

What I've checked with the customer:
* they don't have any references to kubernetes 3.22.0 in their state file, only 3.21.4
* all their `requirements.txt` use pinned versions

### Steps to reproduce

1. pulumi plugin rm --all --yes
2. pulumi plugin ls
3. deactivate; 
4. rm -rf venv/;
5. python -m venv venv; 
6. source /home/aureq/work/customers/zendesk/1902/venv/bin/activate
[...]

PR #802, opened 2022-10-25
Pin all dependency versions

danielrbradley:
### Proposed changes

Quick fix to work around lack of lock files in plugin. This will stop us always pulling the latest versions on users' machines.

### Possible issues

We're going to cause downgrades in pretty much all non-nodejs users but this isn't too bad as it should then be a predictable version going forward.

Nodejs users will also be forced to resolve this dependency to this specific version rather than allowing them to choose another compatible version. This pinning should also be worked around by using custom package resolutions, if required.

This makes it a little harder to upgrade dependencies as we now have to manually edit the `package.json` rather than just run `yarn upgrade`.

### Related issues (optional)

Tactical fix for https://github.com/pulumi/pulumi-eks/issues/800

### DIFF
--- a/nodejs/eks/package.json
+++ b/nodejs/eks/package.json
@@ -11,15 +11,15 @@
     "homepage": "https://pulumi.com",
     "repository": "https://github.com/pulumi/pulumi-eks",
     "dependencies": {
-        "@pulumi/aws": "^5.1.2",
-        "@pulumi/kubernetes": "^3.0.0",
+        "@pulumi/aws": "5.10.0",
+        "@pulumi/kubernetes": "3.20.0",
         "@pulumi/pulumi": "^3.0.0",
-        "axios": "^0.21.1",
-        "netmask": "^2.0.1",
-        "js-yaml": "^3.13.0",
-        "which": "^1.3.1",
-        "https-proxy-agent": "^5.0.0",
-        "semver": "^7.3.7"
+        "axios": "0.27.2",
+        "netmask": "2.0.2",
+        "js-yaml": "3.14.1",
+        "which": "1.3.1",
+        "https-proxy-agent": "5.0.1",
+        "semver": "7.3.7"
     },
     "devDependencies": {
         "@types/axios": "^0.14.0",
--- a/nodejs/eks/package.json
+++ b/nodejs/eks/package.json
@@ -12,7 +12,7 @@
     "repository": "https://github.com/pulumi/pulumi-eks",
     "dependencies": {
         "@pulumi/aws": "5.10.0",
-        "@pulumi/kubernetes": "3.20.0",
+        "@pulumi/kubernetes": "3.20.2",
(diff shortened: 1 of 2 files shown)


roothorp (approved):

roothorp (approved):

nodejs/eks/yarn.lock
```diff
@@ -35,7 +35,7 @@
   resolved "https://registry.yarnpkg.com/@logdna/tail-file/-/tail-file-2.2.0.tgz#158a362d293f940dacfd07c835bf3ae2f9e0455a"
```
lukehoban, line 35: Curious - why doesn’t this lock file end up being used?  Is it because we use yarn here and end users use npm?  Cc @Frassle

lukehoban (approved):

nodejs/eks/package.json
```diff
@@ -11,15 +11,15 @@
     "homepage": "https://pulumi.com",
     "repository": "https://github.com/pulumi/pulumi-eks",
     "dependencies": {
-        "@pulumi/aws": "^5.1.2",
-        "@pulumi/kubernetes": "^3.0.0",
+        "@pulumi/aws": "5.10.0",
+        "@pulumi/kubernetes": "3.20.0",
```
ringods, line 15: @danielrbradley why this version and not 3.21.4 (the latest before v3.22.0 made SSA the default)?

nodejs/eks/yarn.lock
danielrbradley: The lock file is lost somewhere during the publishing pipeline. If we don't get vercel/pkg implemented soon, we should instead investigate why this is excluded.

danielrbradley: The secondary issue with the lock file is that we've only got a yarn.lock, but if the user doesn't have yarn available then it won't help.

nodejs/eks/package.json
ringods: Almost: it was 3.20.**2** 😉

nodejs/eks/yarn.lock
Frassle: > end users use npm

The engine just does an "npm install" when installing dependencies.

Frassle: Should do

danielrbradley: See https://github.com/pulumi/pulumi/issues/11147 - we should only ever use `npm ci` or `yarn install --frozen-lockfile`

nodejs/eks/package.json
danielrbradley: Trying bumping to the latest version as tests seem to be stuck having pinned the version. It's possible that our test entirely ignore the lock file too 🤦

danielrbradley: Well, moving to the latest versions breaks a whole load of tests ... so going back to the pinned versions again 🤷

PR #813, opened 2022-11-16
Fix defaults v mutually exclusive args conflict

### LINKED ISSUE #643: SDK default values cause validation failures
## Hello!

- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

We have code that has worked for months that is suddenly sporadically throwing this error: `details = "Setting nodeGroupOptions, and any set of singular node group option(s) on the cluster, is mutually exclusive. Choose a single approach."` We hit a number of failures on Friday, it began working again this weekend and is now not working again.

Our code in question is here:
* https://github.com/nginxinc/kic-reference-architectures/blob/master/pulumi/aws/eks/__main__.py#L61-L65
* https://github.com/nginxinc/kic-reference-architectures/blob/master/pulumi/aws/eks/__main__.py#L73-L83

There is nothing being set outside of the node options that should cause this error (based on the logic in https://github.com/pulumi/pulumi-eks/blob/master/nodejs/eks/cluster.ts#L361-L375).

The output looks like this:
* https://gist.github.com/qdzlug/931c07928dca1c3ad6097d89227edc16

This issue started following the issues with AWS, but I don't know if it's related. This is under Pulumi v3.19.0 and all current versions of the pulumi python modules. 

### Steps to reproduce

1. Clone the [kic-reference-architecture repo](https://github.com/nginxinc/kic-reference-architectures), being sure to init the submodule.
2. Configure the project; there are steps in the getting-started guide that can be followed.
3. Run the `start_all.sh` script.

Expected: The EKS step should stand up an EKS cluster, and the rest of the build should continue.
Actual: Errors are thrown as shown in the above gist.

roothorp:
PR #641 added a default value to the Cluster argument `nodeRootVolumeSize`, which causes all the generated SDKs to give this field a default value if not supplied. However, https://github.com/pulumi/pulumi-eks/blob/aab0ed9cd224a8b0a4d9f9edf65b535cfd36be38/nodejs/eks/cluster.ts#L231 checks that _either_ the argument `nodeGroupOptions` _or_ individual arguments including `nodeRootVolumeSize` are provided, and not both. Since `nodeRootVolumeSize` always has a value, this means that if you supply `nodeGroupOptions`, it will fail to create your cluster.

On the basis that the platform applies the default and the SDK does not need to, this PR removes the default for that specific field from the schema.

Fixes #643.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 - [Add support for non-default AWS partitions #788](https://github.com/pulumi/pulumi-eks/pull/788)
 - [Add support for launchTemplateTagSpecifications within NodeGroupV2 #810](https://github.com/pulumi/pulumi-eks/pull/810)
 - [Use pkg for packaging provider binary #776](https://github.com/pulumi/pulumi-eks/pull/776)
+- [Remove default for NodeRootVolumeSize that conflicted with NodeGroupOptions #813](https://github.com/pulumi/pulumi-eks/pull/813)
 
 ## 0.42.2 (Released Oct 12, 2022)
 - Fix internal registration of NodeGroupV2 resource.
--- a/examples/cluster-cs/Cluster.csproj
+++ b/examples/cluster-cs/Cluster.csproj
@@ -8,6 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="Pulumi.Aws" Version="5.*" />
+    <PackageReference Include="Pulumi.Awsx" Version="1.0.0-beta.*" />
   </ItemGroup>
 
 </Project>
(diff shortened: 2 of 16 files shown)


squaremo (commented):
Looks right and good and nicely self-contained to me. As a co-author though, I'll leave the official green-ticking to Daniel.

danielrbradley:
I think this approach is all correct, but I think we should probably remove all default values in the schema. We never want the SDK providing a default value rather - we're rather the SDK passes through the lack of value and let the nodejs implementation calculate the defaults provider-side.

danielrbradley (approved):

Issue #462, opened 2020-11-12
Improve typing of Cluster.kubeconfig output


EvanBoyle:
`Cluster.kubeconfig` is typed as `Output<any>`, which seems overly permissive given the implementation: 

https://github.com/pulumi/pulumi-eks/blob/master/nodejs/eks/cluster.ts#L182

Through tracing this code, it appears that we always get an object back. The any typing can lead to some easy to product errors, such as specifying the raw kubeconfig object in a property that requires a string: https://github.com/pulumi/pulumi-kubernetes/issues/1231

While that panic has been patched, it could still lead to issues elsewhere and we should consider using a more descriptive type if possible.

EvanBoyle:
Concrete proposal is to change the type of `kubeconfig` to be a string. This is however a breaking change.

justinvp:
> Concrete proposal is to change the type of `kubeconfig` (and the underlying impl that produces it) to be a string. This is however a breaking change.

The alternative is to add a new API like `kubeconfigJSON` which returns a string, and possibly deprecate `kubeconfig`.

In the meantime, it is possible to convert the object to a JSON string, e.g.:

## TypeScript

```typescript
const kubeconfig = cluster.kubeconfig.apply(JSON.stringify);
```

## Python

```python
import json
kubeconfig = cluster.kubeconfig.apply(lambda k: json.dumps(k))
```

swensons:
> > Concrete proposal is to change the type of `kubeconfig` (and the underlying impl that produces it) to be a string. This is however a breaking change.
> 
> The alternative is to add a new API like `kubeconfigJSON` which returns a string, and possibly deprecate `kubeconfig`.
> 
> In the meantime, it is possible to convert the object to a JSON string, e.g.:
> 
> ## TypeScript
> ```ts
> const kubeconfig = cluster.kubeconfig.apply(JSON.stringify);
> ```
> 
> ## Python
> ```python
> import json
> kubeconfig = cluster.kubeconfig.apply(lambda k: json.dumps(k))
> ```

hi, I'm not able to convert the kubeconfig using this strategy. My code:
```
print(aws_eks_cluster.kubeconfig.apply(lambda k: json.dumps(k)))
```

which yields:
```    Calling __str__ on an Output[T] is not supported.
    To get the value of an Output[T] as an Output[str] consider:
    1. o.apply(lambda v: f"prefix{v}suffix")
    See https://pulumi.io/help/outputs for more details.
[...]

PR #815, opened 2022-11-17
Change NodeGroup/NodeGroupV2/ManagedNodeGroup to accept `Cluster` rather than `CoreData`

### LINKED ISSUE #483: Python/.NET/Go: Support passing `Cluster` directly as an input
In the TypeScript API, there are a few places that accept a `cluster` argument typed as `Cluster | CoreData` (`CoreData` is an interface representing the outputs of the cluster):

https://github.com/pulumi/pulumi-eks/blob/26c027a9e6396a4e9346e720b80684871c94d215/nodejs/eks/nodegroup.ts#L260

In the schematized API, this is currently typed as a `CoreData` type, which means you need to pass it as follows (Python):

```python
cluster=cluster.core
```

Ideally, we'd be able to pass the cluster itself, just like in TypeScript:

```python
cluster=cluster
```

I haven't tried this yet, but it may all "just work" if the proper registrations are in place for `Cluster`.

### LINKED ISSUE #811: Creating a `NodeGroupV2` from Python fails, but works from TypeScript
Creating a `NodeGroupV2` from Python fails. It works from TypeScript.

This uses the `iam` utility from: https://github.com/pulumi/pulumi-eks/blob/master/examples/managed-nodegroups-py/iam.py

```python
import pulumi_aws as aws
import pulumi_eks as eks

import iam

role0 = iam.create_role("example-role0")
instance_profile0 = aws.iam.InstanceProfile("example-instanceProfile0", role=role0.name)

# Create an EKS cluster.
cluster = eks.Cluster(
    "example-managed-nodegroups", skip_default_node_group=True, instance_roles=[role0]
)

eks.NodeGroupV2(
    "example-ng2-simple-ondemand",
    cluster=cluster.core,  # TODO[pulumi/pulumi-eks#483]: Pass cluster directly.
    instance_type="t2.medium",
    desired_capacity=1,
    min_size=1,
    max_size=2,
    labels={"ondemand": "true"},
    instance_profile=instance_profile0,
)
```

Running this fails with:

```
* creating Auto Scaling Group (example-ng2-simple-ondemand): ValidationError: At least one Availability Zone or VPC Subnet is required.
    	status code: 400, request id: 2129668f-9777-4ef1-952f-32b9dd18fc63
```

The equivalent example in TypeScript works:

https://github.com/pulumi/pulumi-eks/blob/ff52acdcf9cd35eb9cb4e3200d294dc8c1737470/examples/nodegroup/index.ts#L42-L50

## Underlying Cause

The problem is manifested here:

https://github.com/pulumi/pulumi-eks/blob/ff52acdcf9cd35eb9cb4e3200d294dc8c1737470/nodejs/eks/nodegroup.ts#L946-L957

For a TypeScript, the last `else` block is executed, whereas with Python, the `else if (core.privateSubnetIds !== undefined)` case is executed because the passed-in `core` is an `Output` in the MLC case, so `core.privateSubnetIds` will also be a non-undefined `Output` as well. The fix is to do these checks inside an apply.

It looks like there are other places in the codebase that has this same problem. We need to do an overall audit.

justinvp:
This makes node groups work correctly in multi-lang scenarios (e.g. from Python, Go, .NET, Java, etc.) and aligns with how you'd typically use these in Node.js.

Fixes #483
Fixes #811


nodejs/eks/cluster.ts
```diff
@@ -1549,3 +1487,179 @@ export class Cluster extends pulumi.ComponentResource {
...
+ * than modifying the public surface area of the existing `Cluster` class, which is still being
+ * used directly by users using the Node.js SDK. Once we move Node.js over to the generated MLC SDK,
+ * we can clean all this up. Internally, this leverages the same `createCluster` helper method that
+ * `Cluster` uses.
+ *
+ * @internal
+ */
+ export class ClusterInternal extends pulumi.ComponentResource {
+    public readonly clusterSecurityGroup!: pulumi.Output<aws.ec2.SecurityGroup>;
+    public readonly core!: pulumi.Output<pulumi.Unwrap<CoreData>>;
```
justinvp, line 1606: If you look at `CoreData`, you can see its properties aren't "plain" -- they're _mostly_ all typed as `Output<T>`:

https://github.com/pulumi/pulumi-eks/blob/28745d145d53a918c681ee3ae233e67a84a287b8/nodejs/eks/cluster.ts#L128-L150

As of a couple days ago, I was using an internal copy of this interface from generating the Node.js SDK that had all plain properties.

However, I realized I could just use `pulumi.Unwrap` to get rid of all the `Output<T>`s, which means I don't have to include/maintain the internal copy of the interface.

nodejs/eks/nodegroup.ts
```diff
@@ -311,9 +311,9 @@ export interface NodeGroupV2Options extends NodeGroupOptions {
      *
      * Defaults to 50.
      */
-    minRefreshPercentage?: number;
+    minRefreshPercentage?: pulumi.Input<number>;
 
-    launchTemplateTagSpecifications?: awsInputs.ec2.LaunchTemplateTagSpecification[];
+    launchTemplateTagSpecifications?: pulumi.Input<pulumi.Input<awsInputs.ec2.LaunchTemplateTagSpecification>[]>;
```
justinvp, line 316: These should be `Input<T>` to match the signature in the MLC SDKs.

nodejs/eks/nodegroup.ts
```diff
@@ -428,31 +474,81 @@ export class NodeGroupV2 extends pulumi.ComponentResource implements NodeGroupV2
...
+    return createNodeGroupInternal(name, args, pulumi.output(core), parent, provider);
+}
 
-    if (!args.instanceProfile && !core.nodeGroupOptions.instanceProfile) {
-        throw new Error(`an instanceProfile is required`);
-    }
+function createNodeGroupInternal(
+    name: string,
+    args: Omit<NodeGroupOptions, "cluster">,
+    core: pulumi.Output<pulumi.Unwrap<CoreData>>,
```
justinvp, line 533: The main change here is that `core` needs to be typed as `pulumi.Output<pulumi.Unwrap<CoreData>>` rather than just `CoreData` because that's the type that will come back when the `ClusterInternal` is rehydrated.

Same goes for `createNodeGroupV2Internal` and `createManagedNodeGroupInternal`.

nodejs/eks/cluster.ts
```diff
@@ -1549,3 +1487,183 @@ export class Cluster extends pulumi.ComponentResource {
...
+    instanceRoles: pulumi.Output<aws.iam.Role[]>;
+    nodeSecurityGroup: aws.ec2.SecurityGroup;
+    eksClusterIngressRule: aws.ec2.SecurityGroupRule;
+    defaultNodeGroup: NodeGroupData | undefined;
+    eksCluster: aws.eks.Cluster;
+    core: CoreData;
+}
+
+/** @internal */
+export function createCluster(
```
justinvp, line 1506: This is just copied from the original `Cluster` constructor, so it can be used by both `Cluster` and `ClusterInternal`.

justinvp:
I keep running into:

```
Error: API rate limit exceeded for installation ID 807028.
```

I'll wait a bit before re-running the CI workflow.

provider/cmd/pulumi-resource-eks/schema.json
```diff
@@ -924,7 +924,8 @@
                     "description": "Type of capacity associated with the EKS Node Group. Valid values: `ON_DEMAND`, `SPOT`. This provider will only perform drift detection if a configuration value is provided."
                 },
                 "cluster": {
-                    "$ref": "#/types/eks:index:CoreData",
+                    "$ref": "#/resources/eks:index:Cluster",
+                    "plain": true,
```
danielrbradley, line 928: I _think_ we should be supporting both the data and resource options here ideally. 

The data version allows people to use the nodegroup components with clusters which are set up outside of EKS.

What complexities arise from making this a `oneOf`? Do you think this is feasible?

nodejs/eks/package.json
```diff
@@ -14,7 +14,7 @@
     "dependencies": {
         "@pulumi/aws": "5.10.0",
         "@pulumi/kubernetes": "3.20.2",
-        "@pulumi/pulumi": "^3.0.0",
+        "@pulumi/pulumi": "^3.47.0",
```
danielrbradley, line 17: Is 3.47.0 the minimum version with the fix or just the latest version?

examples/nodegroup-py/__main__.py
```diff
@@ -28,7 +28,10 @@
     instance_profile=instance_profile0,
 )
 
-clusterK8sProvider = k8s.Provider("clusterK8sProvider", kubeconfig=cluster.kubeconfig.apply(lambda k: json.dumps(k)))
+clusterK8sProvider = k8s.Provider(
+    "clusterK8sProvider",
+    kubeconfig=cluster.kubeconfig.apply(lambda k: json.dumps(k, sort_keys=True)),
```
danielrbradley, line 33: Why does `sort_keys` matter? Does it change every time otherwise? Or has this just shown up because we're moving things around?

provider/cmd/pulumi-resource-eks/schema.json
justinvp: Ah, ok. I was under the impression that the data part was more of an implementation detail that happened to "leak out" into the public API in Node, but your reasoning about why it's exposed makes sense. I'll give `oneOf` a try -- it should just work.

nodejs/eks/package.json
justinvp: It's the minimum version with the fix, and currently the latest version. Why?

examples/nodegroup-py/__main__.py
justinvp: It was changing, causing the provider to be recreated along with all the resources using that provider. It wouldn't always happen, but fairly consistently when running `pulumi up` multiple times.

An improvement over having users do this would be #462, adding a `kubeconfigJSON` output, which would return the canonicalized (sorted) JSON string, so users wouldn't have to deal with converting the object to JSON and keeping it sorted. I can add a commit to this PR that adds that.

nodejs/eks/package.json
danielrbradley: Cool - yup that's good - just checking it's not overly-restrictive

danielrbradley (approved):
Great - this looks good to go!

PR #825, opened 2022-11-23
Update README.md

danielrbradley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ If you are interested in contributing, please see the [contributing docs][contri
 
 ## Code of Conduct
 
-You can read the code of conduct [here][code-of-conduct].
+Please follow the [code of conduct][code-of-conduct].
 
 [contributing]: CONTRIBUTING.md
 [code-of-conduct]: CODE-OF-CONDUCT.md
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -7,22 +7,27 @@
...
 - The AWS CNI Plugin [`aws-k8s-cni`](https://github.com/aws/amazon-vpc-cni-k8s/) to manage pod networking in Kubernetes.
 
-## Pre-Requisites
 
-1. [Install Pulumi](https://www.pulumi.com/docs/reference/install).
-1. [Install `kubectl`](https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-kubectl) (v1.24.0 or higher).
-1. [Install AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html) (>=v1.24.0 for v1 or >=2.7.0 for v2).
-1. [Install Node.js](https://nodejs.org/en/download/) (required even if using this library from another Pulumi language).
+<div>
+    <a href="https://www.pulumi.com/templates/kubernetes/aws/" title="Get Started">
```
roothorp, line 20: I think users will still need the AWS CLI and kubectl, which isn't explicitly stated on this template page (but could maybe be safely assumed?)

roothorp (approved):

README.md
danielrbradley: Yeah, good to not have it duplicated here, but wherever the entry point is for new users, we should probably include some instalation instructions.

PR #839, opened 2022-12-13
Add prettier formatting

roothorp:
Add prettier as a dep, add minimal config in package.json, and run.


danielrbradley (approved):
This looks like a good improvement all round without causing too many spurious diffs 👍

PR #859, opened 2023-02-23
Allow users to override Name tag in autoscaling group

zeevallin:
### Proposed changes
Currently you get confilcts every time you update the cluster through Pulumi if you've provided a Name tag using the AutoScaling group tags. This fix will allow you to override the Name tag to ensure you can control the name of the nodes.

### Related issues (optional)
None (as far as I can tell)

### DIFF
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -843,14 +843,17 @@ ${customUserData}
     }
     const autoScalingGroupTags: InputTags = pulumi
         .all([eksCluster.name, args.autoScalingGroupTags])
-        .apply(
-            ([clusterName, asgTags]) =>
-                <aws.Tags>{
-                    Name: `${clusterName}-worker`,
-                    [`kubernetes.io/cluster/${clusterName}`]: "owned",
-                    ...asgTags,
-                },
-        );
+        .apply(([clusterName, asgTags]): aws.Tags => {
+            if (!asgTags) { asgTags = {} }
+            // The kubernetes.io/cluster/<cluster-name> tag is required to be set to "owned"
+            // Users should not be able to override this tag.
+            asgTags[`kubernetes.io/cluster/${clusterName}`] = "owned";
+            // Only set the Name tag if it's not already set to allow users to override it.
+            if (!asgTags["Name"]) {
+                asgTags["Name"] = `${clusterName}-worker`;
+            }
+            return <aws.Tags>asgTags;
+        });
 
     const cfnTemplateBody = pulumi
         .all([


nodejs/eks/nodegroup.ts
```diff
@@ -843,14 +843,17 @@ ${customUserData}
     }
     const autoScalingGroupTags: InputTags = pulumi
         .all([eksCluster.name, args.autoScalingGroupTags])
-        .apply(
-            ([clusterName, asgTags]) =>
-                <aws.Tags>{
-                    Name: `${clusterName}-worker`,
-                    [`kubernetes.io/cluster/${clusterName}`]: "owned",
-                    ...asgTags,
```
danielrbradley, line 851: The spread in the original implementation is after the explicit name - and therefore if a `Name` exists in `asgTags` it should overwrite the `Name` specified before the spread.

Do you have example code which demonstrates the issue you're seeing with the custom name not being applied?

danielrbradley: Quick test run:

```javascript
const clusterName = "test-cluster"
const asgTags = {
  Name: "overriden-name"
}
const result = {
  Name: `${clusterName}-worker`,
  [`kubernetes.io/cluster/${clusterName}`]: "owned",
  ...asgTags
};

console.log(result);
```

prints:
```
{
  kubernetes.io/cluster/test-cluster: "owned",
  Name: "overriden-name"
}
```

zeevallin: As it stands now it works to set your own name with the tag, that's not the issue. But everytime you re-run the command you get a diff on tags since the value fetched from AWS does not contain the `Name: {clusterName}-worker` value.

My merge request should make sure that the default name is never applied to the state if another name is provided.

zeevallin: This happens every time you run `preview` or `up` but you've provided your own name tag.
```
                ~ aws:autoscaling/group:Group: (update)
                    [id=k8s-core]
                    [urn=urn:pulumi:dev.eu-central-1::k8s::eks:index:Cluster$aws:iam/instanceProfile:InstanceProfile$eks:index:NodeGroupV2$aws:autoscaling/group:Group::k8s-core]
                    [provider=urn:pulumi:dev.eu-central-1::k8s::pulumi:providers:aws::default_5_16_2::5f73b234-f1c8-489b-8de0-2c66cf783459]
                  ~ launchTemplate: {
                      ~ version: "18" => output<string>
                    }
                  ~ tags          : [
                      + [9]: {
                              + key              : "Name"
                              + propagateAtLaunch: true
                              + value            : "k8s-worker"
                            }
                    ]
```

glennkentwell: `preview --diff` shows that the `${clusterName}-worker` value for the `Name` tag is appended to the array of tags; I guess last one wins?

```        + aws:autoscaling/group:Group: (create)
            [urn=urn:pulumi:staging::internal-accounts::eks:index:Cluster$aws:autoscaling/group:Group::data-worker-blue-v2]
            [provider=urn:pulumi:staging::internal-accounts::pulumi:providers:aws::default_6_43_0::93a16622-5417-40bf-9bc8-c9c54336e253]
            launchTemplate               : {
                name      : "data-worker-blue-v2-launchTemplate-ca33ba0"
                version   : output<string>
            }
            tags                         : [
                [0]: {
                   ... etc ...
                [6]: {
                    key              : "Name"
                    propagateAtLaunch: true
                    value            : "data-worker-blue"
                }
                [7]: {
                    key              : "kubernetes.io/cluster/main"
                    propagateAtLaunch: true
                    value            : "owned"
                }
                [8]: {
                    key              : "Name"
                    propagateAtLaunch: true
                    value            : "main-worker"
                }
            ]
```

and according to the AWS web console, the last value wins I guess:
image

glennkentwell (approved):

gmhabchi (approved):

blampe:
@zeevallin @glennkentwell @gmhabchi and anyone else interested in this PR, please check whether the issue still reproduces on the latest (v3.9.1) release. If it does, please create an issue!

Issue #864, opened 2023-03-02
TestAccOidcIam fails due to unexpected difference


squaremo:
This last worked on Feb 17 2023, and started failing conspicuously again once the build_sdk task was fixed: https://github.com/pulumi/pulumi-eks/actions/runs/4304083799/jobs/7509973841#step:23:835 for example.

This is the relevant bit of the captured test output:

```
      program.go:1260: Roundtripping checkpoint via stack export and stack import
      command.go:34: **** Invoke '/home/runner/.pulumi/bin/pulumi stack export --file stack.json' in '/tmp/p-it-fv-az554-1-oidc-iam-s-c3deeb82-3677261096'
      command.go:106: Command completed without output
      command.go:34: **** Invoke '/home/runner/.pulumi/bin/pulumi stack import --file stack.json' in '/tmp/p-it-fv-az554-1-oidc-iam-s-c3deeb82-3677261096'
  Import complete.
      command.go:106: Command completed without output
      program.go:1272: Performing empty preview and update(no changes expected)
      command.go:34: **** Invoke '/home/runner/.pulumi/bin/pulumi preview --non-interactive --diff --expect-no-changes' in '/tmp/p-it-fv-az554-1-oidc-iam-s-c3deeb82-3677261096'
  Previewing update (p-it-fv-az554-1-oidc-iam-s-c3deeb82)
  
  View Live: https://app.pulumi-staging.io/pulumi/example-oidc-iam-sa/p-it-fv-az554-1-oidc-iam-s-c3deeb82/previews/5a5c5c32-4654-471d-9494-0e8003a32f1c
  
    pulumi:pulumi:Stack: (same)
      [urn=urn:pulumi:p-it-fv-az554-1-oidc-iam-s-c3deeb82::example-oidc-iam-sa::pulumi:pulumi:Stack::example-oidc-iam-sa-p-it-fv-az554-1-oidc-iam-s-c3deeb82]
      ~ aws:s3/bucket:Bucket: (update)
          [id=pod-irsa-job-bucket-965e060]
          [urn=urn:pulumi:p-it-fv-az554-1-oidc-iam-s-c3deeb82::example-oidc-iam-sa::aws:s3/bucket:Bucket::pod-irsa-job-bucket]
[...]

PR #865, opened 2023-03-14
Bumps Pulumi AWS to version 5.31.0

kpitzen:
This adds a fix which was rolled into pulumi-aws caused by AWS adding SSE by default for new S3 buckets.  That caused some tests which expected no changes between pulumi steps to have changes as the SSE default was not recognized by pulumi.

### Related issues (optional)

Fixes https://github.com/pulumi/pulumi-eks/issues/864


squaremo:
Looks like this updates all examples, and the NodeJS and Go SDKs. Are the analogous changes to be made to the Python, .NET and Java SDKs? (And do you know of any documentation about what needs to change?)

kpitzen:
> Looks like this updates all examples, and the NodeJS and Go SDKs. Are the analogous changes to be made to the Python, .NET and Java SDKs? (And do you know of any documentation about what needs to change?)

Good catch - I'll make a change to update the Java dependencies.  Dotnet captures any minor version so the change for users is just to bump this which should catch the new version of pulumi-aws.

Python is a bit different in terms of how it handles dependencies - our templates _only_ specify major versions which I considered changing here since it only applies to new projects.  The actual fix is just bumping the version of pulumi-aws which is backing pulumi-eks, or updating pulumi-eks which will transitively update pulumi-aws.

kpitzen:
Also to your note about documentation - I don't have a clue!  I just references a prior PR which did a similar thing (three years ago).  Hence the missed Java update.  Ideally, we should have a versioning tool that updates versions for our pulumi dependencies automatically.

sdk/java/build.gradle
```diff
@@ -43,7 +43,7 @@ repositories {
 dependencies {
     implementation("com.google.code.findbugs:jsr305:3.0.2")
     implementation("com.google.code.gson:gson:2.8.9")
-    implementation("com.pulumi:aws:5.4.0")
+    implementation("com.pulumi:aws:5.31.0")
```
squaremo, line 46: Looks (from the failing action https://github.com/pulumi/pulumi-eks/actions/runs/4416938756/jobs/7742045747#step:16:40) like this is calculated and written here, rather than read from here. I would have changed it here, too.

kpitzen:
For any reviewers:  I spot checked some versions in `schema.json` for `pulumi-aws` versions 5.4.0 and 5.31.0 and the types of each resource are the same between them.

squaremo:
I confirmed (with a regexp -- come at me) that each line in that output just differs in the version in the path of the type name. I'm going to assume that's simply a limitation of schema-tools -- though not sure why it doesn't turn up in other updates.

PR #854, opened 2023-02-02
Create fargate profile with the given private subnets when provided

rdeeba:
### Proposed changes

When deploying an EKS cluster with Fargate profile, the configured subnets in FargateProfile were being ignored. 
I am unable to run integration tests on my side due to lack of permissions in my AWS account, I need someone validate that this tiny change didn't break anything.

Issue #835, opened 2022-12-09
Error when aws provider is disabled


joffreychambrin:
### What happened?

With pulumi-eks version 1.0.0 in python, we have the following error when the aws provider is disabled 
`Exception: Invoke: Default provider for 'aws' disabled. 'aws:index/getPartition:getPartition' must use an explicit provider.`

### Steps to reproduce

Create the following pulumi program 

```python
from pulumi import ResourceOptions
from pulumi_eks import Cluster
from pulumi_aws import Provider

provider = Provider("aws", region="us-west-2")
Cluster("hi", ResourceOptions(providers=[provider]))
```

Disable the aws default provider in the config
```yaml
config:
  pulumi:disable-default-providers:
    - aws
```

### Expected Behavior

No error

### Actual Behavior

`Exception: Invoke: Default provider for 'aws' disabled. 'aws:index/getPartition:getPartition' must use an explicit provider.`

### Output of `pulumi about`

❯ pulumi about
CLI
Version      3.49.0
Go Version   go1.19.4
Go Compiler  gc

Host
OS       darwin
Version  13.0.1
Arch     arm64

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

squaremo:
Thanks for posting this @joffreychambrin. I think it's reasonable to expect to be able to give an explicit provider, and turn the default provider off. There may be a technical reason why this isn't possible; but on first look, I'm going to leave this as `kind/bug`.

bacek:
Any progress on it? I hit exactly same issue.

jaxxstorm:
I believe this is fixed by #886

PR #886, opened 2023-06-19
Use parent provider instead of provider and fix fargate subnets parameter

vboulineau:
### Proposed changes

Some AWS Invoke calls made inside this plugin are using `.provider` attribute to make calls. This attribute comes from parent `.provider`.

However most of the time the `providers` attribute will be used instead because if you have use:
`pulumi:disable-default-providers: ["*"]`
You need to both pass an `aws` provider and an `eks` provider.

Then the issue is that `InvokeOptions` does not support `providers`, but only a single `provider`.
Instead of trying to find and get the right provider, I'm falling back to using `parent` instead (as it's already used by other calls).

Finally a bug was introduced by #854 which makes Fargate subnet IDs mandatory, fixing this (not familiar with TS, there might be a shorter way).

### Related issues (optional)

#854 Introduced a but that makes it mandatory to provide subnet IDs, fixing that.
#835


nodejs/eks/cluster.ts
```diff
@@ -918,7 +918,13 @@ export function createCore(
                         selectors: selectors,
                         subnetIds: pulumi
                             .output(clusterSubnetIds)
-                            .apply((subnets) => computeWorkerSubnets(parent, fargate.subnetIds ?? subnets)),
+                            .apply((subnets) => {
+                                if (fargate.subnetIds?.length && fargate.subnetIds.length > 0) {
```
thomas11, line 922: This change treats undefined and "defined but empty" the same. Is this really correct? Can there be a case where the user might want to express "no subnet ids"?

thomas11 (commented):
Sorry for the delay here, @vboulineau! I left a question, and once that's resolved there's a merge conflict to resolve.

nodejs/eks/cluster.ts
vboulineau: AFAIK that's what we want. If you don't specify (undefined or empty), you fallback to cluster subnets. You cannot give empty list of subnets to prevent falling back to cluster subnets (that's also how AWS API works AFAIK).

thomas11 (approved):

thomas11:
Thank you for your contribution, @vboulineau !

PR #894, opened 2023-07-21
Fix dynamic lookup of amiId for NodeGroup and NodeGroupV2

harstarb:
### Proposed changes

When `amiId` is not specified the `amiType` is incorrectly decided upon due to missing brackets after the null coalescing operator.

The `amiType` will always equal `amazon-linux-2-gpu`, confirmed in a REPL:
```
"test" ?? false ? "amazon-linux-2-gpu" : "amazon-linux-2"
> 'amazon-linux-2-gpu'
"test" ?? true ? "amazon-linux-2-gpu" : "amazon-linux-2"
> 'amazon-linux-2-gpu'
undefined ?? true ? "amazon-linux-2-gpu" : "amazon-linux-2"
> 'amazon-linux-2-gpu'
```

The fix is to use brackets:
```
"test" ?? (false ? "amazon-linux-2-gpu" : "amazon-linux-2")
> 'test'
"test" ?? (true ? "amazon-linux-2-gpu" : "amazon-linux-2")
> 'test'
undefined ?? (true ? "amazon-linux-2-gpu" : "amazon-linux-2")
> 'amazon-linux-2-gpu'
```

### DIFF
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -747,7 +747,7 @@ ${customUserData}
     // https://docs.aws.amazon.com/eks/latest/userguide/retrieve-ami-id.html
     let amiId: pulumi.Input<string> | undefined = args.amiId;
     if (!amiId) {
-        const amiType = args.amiType ?? args.gpu ? "amazon-linux-2-gpu" : "amazon-linux-2";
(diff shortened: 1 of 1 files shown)


rquitales (approved):
Thanks for the fix!

Issue #676, opened 2022-04-01
eks cluster creation fails with `Unhandled exception: TypeError: Cannot read properties of undefined (reading 'data')`


mohitreddy1996:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

We use Amazon EKS pulumi component (using typescript) to create EKS clusters. the configuration looks more or less as follows  - 

```
const cluster = new eks.Cluster(`eks-cluster`, {
        // vpcId,
        endpointPrivateAccess: true,
        endpointPublicAccess: true,
        subnetIds: // subnet ids,
        nodeGroupOptions: {
            // nodetype, desired capacity etc
            nodeAssociatePublicIpAddress: false,
        },
        providerCredentialOpts: {
            roleArn,
        },
        nodeAssociatePublicIpAddress: false,
        createOidcProvider: true
    }, { provider: awsProvider });
``` 

the resources plugin versions 
- aws - `4.38.1` 
- eks - `0.37.1`

We have this setup for a few months now and recently started seeing this error.

Error - `Unhandled exception: TypeError: Cannot read properties of undefined (reading 'data')`

I am assuming this is being raised from one of these blocks - https://github.com/pulumi/pulumi-eks/blob/master/nodejs/eks/cluster.ts#L563 (we do see log `Cluster is ready` which could be from here - https://github.com/pulumi/pulumi-eks/blob/master/nodejs/eks/cluster.ts#L544)

I wonder if there is a version mismatch or any new parameters which need to be set on Cluster definition.

guineveresaenger:
Hi @mohitreddy1996 - would it be possible for you to use v5.1.1, and report if the issue persists?

re-thc:
Have a similar error. Tried updating to latest including v5.1.1 and still get the same result. Below is the diagnostics...

Diagnostics:  
  pulumi:pulumi:Stack (test):  
    error: Running program '/builds/test/packages/inf/src' failed with an unhandled exception:  
    TypeError: Cannot read properties of undefined (reading 'data')  
        at /builds/test/node_modules/@pulumi/cluster.ts:570:103  
        at /builds/test/node_modules/@pulumi/output.ts:383:31  
        at Generator.next (<anonymous>)  
        at /builds/test/node_modules/@pulumi/pulumi/output.js:21:71  
        at new Promise (<anonymous>)  
[...]

PR #903, opened 2023-08-22
Guard against cluster certificate authority being undefined

danielrbradley:
Hopefully addresses #676

Although the types don't describe the certificate authority as being possibly undefined, the error reported would indicate it can be (these are the only reference of accessing `.data` in the cluster code). Therefore we'll add some extra guards just in case.

If the cluster certificate authority is undefined, then we'll pass the certData to generateKubeConfig as undefined. This will then not set the property within the kubeconfig, but at least it won't throw a hard error causing the whole deployment to fail.

Aside: Fix resolve() call which requires a value to be passed.

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -199,7 +199,7 @@ interface ExecEnvVar {
 export function generateKubeconfig(
     clusterName: pulumi.Input<string>,
     clusterEndpoint: pulumi.Input<string>,
-    certData: pulumi.Input<string>,
+    certData?: pulumi.Input<string>,
     opts?: KubeconfigOptions,
 ) {
     let args = ["eks", "get-token", "--cluster-name", clusterName];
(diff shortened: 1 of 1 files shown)


thomas11 (commented):
In `generateKubeconfig`, is it still ok to write `"certificate-authority-data": certData` unconditionally? What happens in the undefined case?

thomas11 (approved):

danielrbradley:
@t0yv0 unfortunately most of these calls are inside an apply which is called after the cluster creation. We _could_ add another program test but the EKS program tests are also not super reliable as it is and are quite costly given the time take for cluster creation.

I'd suggest getting unit tests around this code should probably be a part of a separate refactoring effort.

t0yv0:
That's very unfortunate. 

I was thinking of a replay test.

https://github.com/pulumi/pulumi-aws/blob/9b14ba69d12fd30ad570510b420ee4edfeefe7bb/examples/examples_test.go#L107 

https://github.com/pulumi/pulumi-hcloud/blob/b8a1e059ca986c852b13bd990be537a41c023ec8/examples/regress_170_test.go#L28

But I guess it's not applicable to providers written in TypeScript.

t0yv0:
Backlogged https://github.com/pulumi/pulumi-eks/issues/904

Issue #927, opened 2023-10-18
Add upgrade tests


mnlumi:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
Add tests to validate behavior of existing resources as they are upgraded between versions, especially major version changes. Tests should target the key components prioritized by popularity and usage.

t0yv0:
I'm picking this up but running into a tad of difficulty with providertest 0.0.1. Because this provider is not implemented in Go, Quick-mode upgrade tests are not available without further work (invert gRPC client/service connector). Now, PreviewOnly mode tests should be available. Instead of adding test programs I tried to exercise existing examples but ran into limitations that the 0.0.1 version hard-assumes that the tests are written in YAML. I'm looking into lifting that limitation and installing upgrade tests here over existing example programs.

PR #929, opened 2023-10-19
Add provider upgrade tests

t0yv0:
### Proposed changes

This PR adds upgrade tests that verify that taking a stack provisioned on a baseline version of the provider and upgrading it to the current version creates no replacements.

Fixes https://github.com/pulumi/pulumi-eks/issues/927

Dependent PR: https://github.com/pulumi/providertest/pull/19

- [ ] providertest is not correctly reinstalling the TypeScript SDKs for the upgrade test
- [ ] possibly as a result of that, the recorded baseline has the wrong AWS provider version listed
- [ ] if we want to record more example programs here, GHA automation is in order to parallelize, as each takes ~20 min
- [ ] need to ensure these tests are run in CI


provider/testdata/recorded/TestProviderUpgrade/cluster/1.0.4/state.json
```diff
@@ -0,0 +1,5309 @@
...
+                "modified": "2023-10-19T00:46:12.020491Z",
+                "sourcePosition": "project:///node_modules/@pulumi/runtime/stack.ts#36,23"
+            },
+            {
+                "urn": "urn:pulumi:p-it-antons-mac-cluster-0421759d::example-cluster::pulumi:providers:aws::default_5_31_0",
+                "custom": true,
+                "id": "4b6db602-50de-4739-ac98-93c6c721d82b",
+                "type": "pulumi:providers:aws",
+                "inputs": {
+                    "version": "5.31.0"
```
t0yv0, line 183: Hmm. This is not right unfortunately. The specification stated 5.42.0 - which should have been the provider that got installed in PATH. Either that's the provider really being used but the version here is taken from the SDK, or the ambient plugin installation didn't work right.

provider/provider_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/pulumi/providertest"
+)
+
+func TestExamples(t *testing.T) {
+	t.Run("cluster", func(t *testing.T) {
+		test(t, "../examples/cluster").Run(t)
+	})
+}
+
+func test(t *testing.T, dir string, opts ...providertest.Option) *providertest.ProviderTest {
```
iwahbe, line 15: I don't see where `t` is used.
```suggestion
func test(dir string, opts ...providertest.Option) *providertest.ProviderTest {
```

provider/provider_test.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"testing"
+
+	"github.com/pulumi/providertest"
+)
+
+func TestExamples(t *testing.T) {
+	t.Run("cluster", func(t *testing.T) {
+		test(t, "../examples/cluster").Run(t)
+	})
```
iwahbe, line 12: I don't see any opportunity for the call to `test(...)` to fail. We should just be able to elide the `func` here:
```suggestion
	t.Run("cluster", test(t, "../examples/cluster").Run)
```

t0yv0: Well well, you have to leave markers for the IDE to let you click on the test to run it. I really need to test it in VS Code.

t0yv0:
I've done some more work on providertest to lean harder into pre-existing ProgramTest features to do TypeScript programs much cleaner but I didn't integrate it back here yet. Tomorrow..

t0yv0:
With providertest 0.0.2 it should be handling TypeScript SDKs correctly.

PR #934, opened 2023-10-31
Add upgrade tests on top the v2 branch

thomas11:
If it works out, this PR replaces #927. By targeting the v2 upgrade branch, we avoid conflicts in versions and GH workflows.


mnlumi:
@thomas11 What's the status of this PR?

thomas11:
> @thomas11 What's the status of this PR?

It cannot be merged as is because there there certain diffs in the upgrade tests that we cannot and don't want to get rid of. For example, the version of explicit p-kubernetes providers changes in v2 which is expected. I believe we need providertest support for expected diffs, or something along those lines. Would be good to hear @t0yv0's thoughts on this. Looks like there's no providertest issue yet to track this?

t0yv0:
Would https://github.com/pulumi/providertest/pull/30 give enough control to pick and choose which diffs are OK?

PR #936, opened 2023-11-01
[WIP/DRAFT] Remove yarn link step from cron job - DO NOT MERGE

rquitales:
This should be handled within the makefile target to prevent the wrong folder being linked.

Cron job: https://github.com/pulumi/pulumi-eks/actions/runs/6724304133


rquitales:
Investigations complete. It is the pulumi CLI we need to pin.

PR #937, opened 2023-11-02
Pin Pulumi CLI used in CI to v3.89.0

rquitales:
### Proposed changes

This is required to avoid a bug in upstream v3.90.0 which causes the ` error: expected property name after '.'` error we see.

Note: the workflow files in pulumi-eks do not appear to be managed by ci-mgmt.

### Related issues (optional)

Fixes: https://github.com/pulumi/pulumi-eks/issues/931


thomas11 (approved):
We should file a bug upstream and link it here for tracking.

Issue #912, opened 2023-09-11
Introduce wheel builds and publishing for Python SDK


t0yv0:
Part of https://github.com/pulumi/home/issues/2883 rollout.

Ensure that the project:

- migrates Python SDKs to pyproject.toml
- publishes both wheels and source distributions to PyPI

t0yv0:
This looks easy to accomplish but the difficult part is upgrading to Pulumi 3.81.0. 

IT seems that's happening https://github.com/pulumi/pulumi-eks/pull/901 so I will block this until the upgrade lands.

PR #981, opened 2023-12-01
Feature: Python wheel-based builds and publishing

t0yv0:
### Proposed changes

This feature enables wheel-based builds and PyPI publishing for the provider's Python SDK. It is part of a larger rollout across all providers https://github.com/pulumi/home/issues/2883

### Related issues (optional)

Fixes https://github.com/pulumi/pulumi-eks/issues/912


mjeffryes (approved):
🎉

Issue #918, opened 2023-10-05
aws-auth get deleted on update (replace) render the cluster inaccessible force recreation of the cluster


ami-descope:
### What happened?

applying a change this the module, like adding fargate profile, renders the cluster inaccessible (available for cluster creator ofcourse)

flow:

apply new fargate setup -> passed
rollback to previous setup -> configmap needs to be updated so it gets replaced while running in parallel with other resources, if there is a pulumi error like resource exist, the aws-auth does not get recreated, cluster becomes inaccessible to anyone (cluster creator still have access)

### Example

new code

```ts
const cluster = new eks.Cluster(
            `${name}-eksCluster`,
            {
                vpcId: args.vpcId,
                subnetIds: args.vpcPrivateSubnetIds,
                nodeAssociatePublicIpAddress: false,
                skipDefaultNodeGroup: true,
                storageClasses: "gp2",
                createOidcProvider: true,
                enabledClusterLogTypes: ["api", "audit", "authenticator", "controllerManager", "scheduler"],
                version: config.eksClusterVersion,
                roleMappings: [
                    this.getAwsSSORole(),
                    this.getRoleMapping({
                        roleName: karpenterFargateRoleName,
                        groups: ["system:bootstrappers", "system:nodes", "system:node-proxier"],
                        username: "system:node:{{SessionName}}",
                    }),
                    this.getRoleMapping({
                        roleName: "PulumiGithubAction",
                        groups: ["system:masters"],
                        username: "pulumi:admin-usr",
                    }),
                    this.getRoleMapping({
                        roleName: "PulumiGithubAction-GCP",
                        groups: ["system:masters"],
                        username: "pulumi:admin-usr",
                    }),
                ],
                instanceRoles: [nodeRole],
                tags: tags,
            },
            { parent: this }
        );
```

old code:

```ts
const cluster = new eks.Cluster(
            `${name}-eksCluster`,
            {
                vpcId: args.vpcId,
                subnetIds: args.vpcPrivateSubnetIds,
                nodeAssociatePublicIpAddress: false,
                skipDefaultNodeGroup: true,
                storageClasses: "gp2",
                deployDashboard: false,
                createOidcProvider: true,
                enabledClusterLogTypes: ["api", "audit", "authenticator", "controllerManager", "scheduler"],
                version: config.eksClusterVersion,
                roleMappings: [
                    this.getAwsSSORole(),
                    this.getPulumiGithubActionRole(),
[...]

Issue #883, opened 2023-05-25
Nodes fail to join cluster because aws-auth not updated


liam-j-bennett:
### What happened?

Following [this doc](https://www.pulumi.com/docs/clouds/aws/guides/eks/#configuring-your-eks-clusters-worker-nodes-and-node-groups), creating a nodegroup after the cluster is instantiated, the instances are created, but fail to join the cluster with:

```
time="2023-05-25T08:55:22Z" level=warning msg="access denied" arn="arn:aws:iam::640907157373:role/obs-node-role-bae465a" client="127.0.0.1:50308" error="ARN is not mapped" method=POST path=/authenticate
```

This occurs as the node role for that nodegroup is not added to `aws-auth-cm.yaml`. This happens using both `NewNodeGroup` and `NewNodeGroupV2` functions.

Adding that role to the `aws-auth-cm` fixes the issue.

### Expected Behavior

I expect the `NewNodeGroup` or `NewNodeGroupV2` to update the `aws-auth-cm.yaml` with the node role of the new nodegroup, or the document needs to specify a patch that needs to be applied by the Kubernetes provider.

### Steps to reproduce

1. Create a EKS cluster
```
eksVpc, err := ec2.NewVpc(ctx, "eks-vpc", &ec2.VpcArgs{
			EnableDnsHostnames: pulumi.Bool(true),
			CidrBlock:          &vpcNetworkCidr,
		})
		if err != nil {
			return err
		}

		eksCluster, err := eks.NewCluster(ctx, "eks-cluster", &eks.ClusterArgs{
			VpcId:                        eksVpc.VpcId,
			PublicSubnetIds:              eksVpc.PublicSubnetIds,
			PrivateSubnetIds:             eksVpc.PrivateSubnetIds,
			NodeAssociatePublicIpAddress: pulumi.BoolRef(false),
			RoleMappings: eks.RoleMappingArray{
				eks.RoleMappingArgs{
					Groups: pulumi.StringArray{
						pulumi.String("system:masters"),
					},
					RoleArn:  pulumi.String(opsPrincipalRole),
					Username: pulumi.String("ops-user"),
				},
			},
			InstanceType:    pulumi.String("t2.medium"),
			DesiredCapacity: pulumi.Int(3),
			MinSize:         pulumi.Int(1),
			MaxSize:         pulumi.Int(5),
			EnabledClusterLogTypes: pulumi.StringArray{
				pulumi.String("api"),
				pulumi.String("audit"),
[...]

PR #983, opened 2023-12-05
Defaulted aws-auth ConfigMap creation to upsert, added optional ConfigMap update in place property to cluster

JaseKirby:
### Proposed changes

This PR does two things:

Defaults `aws-auth` ConfigMap management to an [upsert operation](https://www.pulumi.com/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply/#upsert-a-resource). This ensures pulumi takes full control of the ConfigMap regardless of whether it exists already or not eliminating any possible race conditions that could occur and error popping up stating the `aws-auth` ConfigMap already exists.

Adds an optional [`enableConfigMapMutable`](https://www.pulumi.com/registry/packages/kubernetes/api-docs/provider/#enableconfigmapmutable_nodejs) property to pass to the cluster's pulumi k8s provider that will allow for updating the `aws-auth` ConfigMap in place over the default replace operation. A replace operation for this particular ConfigMap can be especially dangerous if some failure blocks recreation after a delete as default M2M auth entries would get wiped as we have seen already intermittently in cluster provisioning workflows: `error: resource kube-system/aws-auth was not successfully created by the Kubernetes API server : configmaps "aws-auth" already exists`

The `enableConfigMapMutable` property is optional because docs denote it is still in preview and if consumers of this library re-use the exposed cluster k8s provider downstream it could have unintended effects on their k8s deployment workflows. AWS documentation like https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html always advises updating/applying/patching this ConfigMap in place.

### Related issues (optional)

Fixes: https://github.com/pulumi/pulumi-eks/issues/918
Fixes: https://github.com/pulumi/pulumi-eks/issues/883

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -713,6 +713,7 @@ export function createCore(
         `${name}-eks-k8s`,
         {
             kubeconfig: kubeconfig.apply(JSON.stringify),
+            enableConfigMapMutable: args.enableConfigMapMutable,
         },
         { parent: parent },
     );
@@ -874,9 +875,13 @@ export function createCore(
         `${name}-nodeAccess`,
         {
             apiVersion: "v1",
+            immutable: false,
             metadata: {
                 name: `aws-auth`,
                 namespace: "kube-system",
+                annotations: {
+                    "pulumi.com/patchForce": "true",
+                },
             },
             data: nodeAccessData,
         },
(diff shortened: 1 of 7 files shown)


rquitales:
Thanks for the PR @JaseKirby. It looks like this is the correct direction to solve the linked issues. It looks like only the nodejs SDK was updated though, so we'll need to replicate this in the other supported languages as well.

rquitales:
Thanks for the initial submission. I've updated the PR to also expose the new argument to the other supported SDK languages.

Issue #738, opened 2022-07-18
Unable to choose correct AMI ID if NodeGroup is ARM64


danielbichuetti:
### What happened?

When creating an EKS cluster, if you create a new NodeGroup which uses ARM instance (e.g., t4g-medium), creating cluster will fail, with error on EC2 ASG, regarding different AMI arch (x86_64) and node arch.

### Steps to reproduce

Create a Pulumi project, use AWS Crosswalk EKS to create cluster and node group.

```
cluster = Cluster(
    "cluster-on-eks",
    ClusterArgs(
        node_associate_public_ip_address=True,
        create_oidc_provider=True,
        skip_default_node_group=True,
    ),
)

medium_node_group = NodeGroup(
    "cluster-ng1",
    cluster=cluster.core,
    desired_capacity=2,
    instance_type="t4g.medium",
    ami_type="amazon-linux-2-arm64", # Need manually setup when is ARM64
    min_size=1,
    max_size=5,
    #labels={"instance": "medium", "lifecycle": "on-demand", "architecture": "arm64"},
    #taints={"arm": TaintArgs(value="true", effect="PreferNoSchedule")},
    opts=pulumi.ResourceOptions(
        parent=cluster,
        providers={
            "kubernetes": cluster.provider,
        },
    ),
)

```

### Expected Behavior

Cluster and node group of ARM64 type are created

### Actual Behavior

Fail to deploy the ASG and NodeGroup.

`Failed Scaling Activity: The architecture 'arm64' of the specified instance type does not match the architecture 'x86_64' of the specified AMI. Specify an instance type and an AMI that have matching architectures, and try again. You can use 'describe-instance-types' or 'describe-images' to discover the architecture of the instance type or AMI. Launching EC2 instance failed.
`
### Versions used

```
pulumi-awsx        1.0.0b9
pulumi-eks         0.41.2
```

### Additional context

Another reference at #582, but it seems to have been addressed.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

viveklak:
Thanks for opening the issue @danielbichuetti. Seems surprising since we do lookup the amiID corresponding to the amiType - which I believe you are providing to indicate this is arm64: https://github.com/pulumi/pulumi-eks/blob/62700bf751ae85414e0bd673d2c036cefdd6f2f6/nodejs/eks/nodegroup.ts#L513

[...]

Issue #582, opened 2021-05-19
Support ARM Instances in default NodeGroup


pquerna:
I think this block of code, which is using SSM to fetch the regional AMI ID, needs change to support ARM instances:
https://github.com/pulumi/pulumi-eks/blob/4f4a75b17de98cf2f9c3d34a960b59503cbc4f0a/nodejs/eks/nodegroup.ts#L470-L478

Specifically, if the InstnaceType is set to an ARM type, the `amiType` needs to be `amazon-linux-2-arm64` 

Without this change, the cloudformation get stuck trying to create the node group, which fails because the AMI picked is invalid for ARM instances.

lobshunter:
Met the same issue, hope this will be fixed soon.

PR #1017, opened 2024-01-25
Improve dynamic AMI ID lookup to support Graviton ARM64 instances in Cluster, NodeGroup and NodeGroupV2 resources

rquitales:
### Proposed changes

This pull request enhances the dynamic lookup process used to calculate the recommended AMI ID. Previously, consideration was limited to x86 and GPU-accelerated instances. With this enhancement, we now include the detection of Graviton ARM instances by parsing the instance type.

#### Changes made:
* Updated AMI ID lookup logic to accommodate Graviton/arm64 instances
* Introduced a Graviton cluster creation in the NodeJS examples integration test
* Added jest to enable unit testing

### Related issues (optional)

Fixes: #738
Fixes: #582


nodejs/eks/nodegroup.ts
```diff
@@ -1685,3 +1668,71 @@ function createManagedNodeGroupInternal(
...
+
+/**
+ * isGravitonInstance returns true if the instance type is a Graviton instance.
+ * We determine this by checking if the third character of the instance type is
+ * a "g".
+ *
+ * See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html.
+ */
+function isGravitonInstance(instanceType: pulumi.Input<string> | undefined): boolean {
+    if (!instanceType) {
```
lukehoban, line 1701: I think technically the first part can be multiple characters - like `Hpc7g`.

rquitales: Ah! Yes you're right. I'll need to account multi-char instance families.

rquitales: Updated to use regex.

lukehoban (approved):

PR #1018, opened 2024-01-25
Attempt to fix PYPI publishing

mjeffryes:
One more place where we still need to roll out the new PYPI API TOKEN

Fixes: https://github.com/pulumi/pulumi-eks/issues/1016


rquitales (approved):
LGTM

Issue #1024, opened 2024-01-26
Provider panics when current kubeconfig context is invalid


rquitales:
### What happened?

Currently, when using this provider, it verifies the presence of `kubectl` version v1.24.0 or higher on the local system. This validation is conducted at the following location: https://github.com/pulumi/pulumi-eks/blob/d274e13001cf8dcbee6d3c49d42a5adf1da4147e/nodejs/eks/dependencies.ts#L43.

However, the default behavior of `kubectl version` includes retrieving the remote API Server version in addition to the client version. This extra roundtrip step is unnecessary as we are solely interested in the client version and do not need to fetch the API server version for the current kubectx. If the current kubectx is invalid as well, then the provider would also panic.

### Example

-

### Output of `pulumi about`

-

### Additional context

-

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #1025, opened 2024-01-26
Only fetch kubectl client version

rquitales:
### Proposed changes
This PR appends the `--client=true` CLI option when grabbing kubectl's version. This will ensure we don't fetch the live server version of whichever cluster/context is in scope, as we don't need that at all. There are some cases where failure in getting the server version results in `kubectl version` completely failing, and not returning the client version as well.

This PR also adds jest tests for the functions in `dependencies.ts`.

### Related issues (optional)

Fixes: #1024


mjeffryes (approved):
Nice, thanks for all the tests

Issue #629, opened 2021-10-05
Enable ExtraRuntimeValidation in `TestAccAwsProfileRolePy`


emiliza:
As part of https://github.com/pulumi/pulumi-eks/pull/628/ 
> TODO: Temporarily skip the extra runtime validation due to test failure

This issue is to track re-enabling that part of the test.

PR #1026, opened 2024-01-27
Re-enable and enhance tests

rquitales:
### Proposed changes

This PR encompasses several enhancements related to tests. In summary, tests have been refactored to run faster, Python tests have been re-enabled, and various regressions/bugs have been addressed during test development.

#### Changes made:

- Re-enabled EKS validation for NodeJS Fargate tests
- Re-enabled all Python tests (#629)
- Resolved a bug causing unscheduled CoreDNS pods for Fargate-only clusters (#1030)
- Improved efficiency by running smoke tests concurrently, resulting in a reduction of test times from 45 minutes to 20 minutes in worst-case scenarios
- Improved test outputs by correcting swapped test assertions for expected and actual values
- Expanded the existing NodeJS cluster test to validate ARM64 node creation
- Fixed bitrotten AWS Profile python test

These changes are segregated into individual commits for easier review. Additionally, they are intended to be rebased and merged independently to ensure their autonomy upon merging into the mainline branch.

### Related issues (optional)

Fixes: #1030 
Fixes: #629

### DIFF
--- a/examples/examples_test.go
+++ b/examples/examples_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 
 	"github.com/pulumi/pulumi/pkg/v3/testing/integration"
+	"github.com/stretchr/testify/assert"
 )
 
 func getEnvRegion(t *testing.T) string {
@@ -73,4 +74,20 @@ func providerPluginPathEnv() (string, error) {
 		pathSeparator = ";"
 	}
 	return "PATH=" + os.Getenv("PATH") + pathSeparator + absPluginDir + pathSeparator + absTestPluginDir, nil
+}
+
+var envToUnset = [...]string{"AWS_SECRET_ACCESS_KEY", "AWS_ACCESS_KEY_ID", "AWS_SESSION_TOKEN"}
+
+// unsetAWSProfileEnv unsets the AWS_PROFILE and associated environment variables.
+// EKS token retrieval using the AWS_PROFILE seems to prefer the
+// the following variables over AWS_PROFILE so you end up with
+// authentication failures in the tests. So drop these environment
+// variables if set and reapply them after the test.
+func unsetAWSProfileEnv(t *testing.T) {
+	t.Helper()
+
+	for _, envVar := range envToUnset {
+		t.Setenv(envVar, "")
+		assert.NoError(t, os.Unsetenv(envVar)) // Explicitly unset the environment variable, as well.
+	}
 }
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -951,7 +951,18 @@ export function createCore(
                             const getAnnosOutputStr = getAnnosOutput.toString();
                             // See if getAnnosOutputStr contains the annotation we're looking for.
                             if (!getAnnosOutputStr.includes("eks.amazonaws.com/compute-type") ) {
-                                // No need to patch, the annotation is not present.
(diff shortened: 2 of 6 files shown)


nodejs/eks/cluster.ts
```diff
@@ -951,7 +951,18 @@ export function createCore(
                             const getAnnosOutputStr = getAnnosOutput.toString();
                             // See if getAnnosOutputStr contains the annotation we're looking for.
                             if (!getAnnosOutputStr.includes("eks.amazonaws.com/compute-type") ) {
-                                // No need to patch, the annotation is not present.
+                                // No need to patch the deployment object since the annotation is not present. However, we need to re-create the CoreDNS pods since
+                                // the existing pods were created before the FargateProfile was created, and therefore will not have been scheduled by fargate-scheduler.
+                                // See: https://github.com/pulumi/pulumi-eks/issues/1030.
+                                const cmd = `kubectl rollout restart deployment coredns -n kube-system`;
```
t0yv0, line 957: Did not expect to find this change there, curious is there any alternative to shelling out? What if the host system does not have kubectl installed?

examples/examples_test.go
```diff
@@ -74,3 +75,19 @@ func providerPluginPathEnv() (string, error) {
...
+// unsetAWSProfileEnv unsets the AWS_PROFILE and associated environment variables.
+// EKS token retrieval using the AWS_PROFILE seems to prefer the
+// the following variables over AWS_PROFILE so you end up with
+// authentication failures in the tests. So drop these environment
+// variables if set and reapply them after the test.
+func unsetAWSProfileEnv(t *testing.T) {
+	t.Helper()
+
+	for _, envVar := range envToUnset {
+		t.Setenv(envVar, "")
```
t0yv0, line 90: This gets troublesome with parallel tests as you might know.

t0yv0 (approved):
Thanks for the work! This is great. Had one question on the change to prod code that starts shelling out to kubectl, otherwise LGTM.

nodejs/eks/cluster.ts
rquitales: We have existing logic that already shells out to `kubectl` for updating the existing CoreDNS deployment object: https://github.com/pulumi/pulumi-eks/blob/62b0c2c788ec60715e6e139d85c570988e4b2ee2/nodejs/eks/cluster.ts#L944-L971
The additional code introduced in this PR simply builds upon that existing functionality.

Fwiw we do check that `kubectl` exist on the local system, and that it is a compatible version as well: https://github.com/pulumi/pulumi-eks/blob/62b0c2c788ec60715e6e139d85c570988e4b2ee2/nodejs/eks/dependencies.ts#L31-L62
Associated tests for the kubectl checking logic: https://github.com/pulumi/pulumi-eks/blob/62b0c2c788ec60715e6e139d85c570988e4b2ee2/nodejs/eks/dependencies.test.ts

Kubernetes does provide generated clients for NodeJS (https://github.com/kubernetes-client/javascript), so it might be nice to use that in the future for these operations.

examples/examples_test.go
rquitales: Yes. I'm just mimicking prior art for the existing AWS Profile tests we had. The tests run in non-parallel mode (which `t.Setenv` ensures as it panics if the tests run in parallel). I don't think there is an trivial way to get rid of this.

PR #1040, opened 2024-02-22
tests: log kubernetes object if object is not in ready state for easier debugging

rquitales:
### Proposed changes

This PR will enable printing the Kubernetes object manifest and status if the object fails to be in a healthy state after running our smoke tests. This will help us figure out why these resources are not ready, and hopefully fix our test flakes.

The diff logic for our upgrade tests based on snapshots is also relaxed to permit diffs when the test program depends on other providers without a version fixed.


provider/provider_test.go
```diff
@@ -82,20 +82,20 @@ func TestReportUpgradeCoverage(t *testing.T) {
...
 			if !diff.HasChanges {
 				continue
 			}
 
-			if strings.Contains(string(diff.URN), "pulumi:providers:kubernetes::") &&
+			if strings.Contains(string(diff.URN), "pulumi:providers:") &&
 				len(diff.Diffs) == 1 &&
 				diff.Diffs[0] == "version" {
-				log.Println("Ignoring version change for kubernetes provider")
+				log.Println("Ignoring version change for providers used in test programs")
```
thomas11, line 98: Could it be that the difference is in major version, and we shouldn't ignore that?

thomas11 (approved):

PR #1045, opened 2024-02-22
Use t3 instances instead of the default t2 which isn't available in all zones

thomas11:


thomas11:
Merged as part of #1050

PR #1050, opened 2024-02-28
Fix broken snapshot tests

rquitales:
### Proposed changes

This PR fixes the issue that we have with re-recording snapshot tests. The older baseline version of v1.0.4 is unable to successfully create Graviton clusters so the baseline version is now updated to v2.2.1.

The commit from #1045 is also cherry picked into this PR, and the providertest framework and therefore, pu/pu also upgraded.

New snapshots were recorded from this CI job: https://github.com/pulumi/pulumi-eks/actions/runs/8074191947

### Related issues (optional)

Fixes: #https://github.com/pulumi/pulumi-eks/issues/1042


thomas11 (approved):
Awesome, thanks!

Issue #868, opened 2023-03-27
Don't set `AWS_PROFILE` automatically if specified


jaxxstorm:
### What happened?

In https://github.com/pulumi/pulumi-eks/pull/561/files a change was made to specify the AWS_PROFILE for the kubeconfig automatically.

Unfortunately, this has the side effect of breaking multi user worklows is the profile name isn't the same for all users.

### Expected Behavior

The `AWS_PROFILE` should never be automatically inferred

### Steps to reproduce

Let's say I have an `AWS_PROFILE` called `pulumi-dev-sandbox` and I create a Pulumi program that provisions a cluster and specifies a provider:

```
cluster = eks.Cluster("example")

k8s_provider = k8s.Provider(
    "k8s-provider",
    kubeconfig=cluster.kubeconfig,
)
```

I have my get a kubeconfig that looks like this:

```
kubeconfig: {
        apiVersion     : "v1"
        clusters       : [
            [0]: {
                cluster: {
                    certificate-authority-data: "<REDACTED>"
                    server                    : "<REDACTED>"
                }
                name   : "kubernetes"
            }
        ]
        contexts       : [
            [0]: {
                context: {
                    cluster: "kubernetes"
                    user   : "aws"
                }
                name   : "aws"
            }
        ]
        current-context: "aws"
        kind           : "Config"
      ~ users          : [
          ~ [0]: {
                    name: "aws"
                  ~ user: {
                      ~ exec: {
                            apiVersion: "client.authentication.k8s.io/v1beta1"
                            args      : [
                                [0]: "eks"
                                [1]: "get-token"
                                [2]: "--cluster-name"
                                [3]: "example-eksCluster-90a12b6"
                            ]
                            command   : "aws"
                          ~ env       : [
                                [0]: {
                                        name : "KUBERNETES_EXEC_INFO"
                                        value: (json) {
                                            apiVersion: "client.authentication.k8s.io/v1beta1"
                                        }

                                    }
                              + [1]: {
                                      + name : "AWS_PROFILE"
                                      + value: "pulumi-dev-sandbox"
                                    }
                            ]
                        }
                    }
                }
        ]
    }
```

Then, let's say my colleague, who has entirely reasonable set the same profile as "dev-sandbox" comes along and runs the same pulumi program. It fails, because the profile name isn't available.

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
[...]

Issue #1038, opened 2024-02-14
The kubeconfig generated is missing the region argument


bryantbiggs:
### What happened?

I am trying to construct a Kubernetes provider thats suitable for deploying Helm charts onto an EKS cluster. However, I am getting a `cluster not found` error because the kubeconfig generated by the Pulumi EKS provider does not contain the region flag

### Example

Use this projects https://github.com/pulumi/pulumi-eks/tree/master/examples/aws-go-eks-helloworld example and inspect the generated kubeconfig - you will see it does not contain the `--region <region>` argument

### Output of `pulumi about`

x

### Additional context

This is what the `aws eks update-kubeconfig --name <name>` command generates

``` yaml
# truncated for brevity
exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - --region
      - <region>
      - eks
      - get-token
      - --cluster-name
      - <name>
      - --output
      - json
      command: aws
```

And this is what the Pulumi kubeconfig generates:

``` yaml
# truncated for brevity
exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - eks
      - get-token
      - --cluster-name
      - <name>
      command: aws
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mjeffryes:
Thanks for reporting this @bryantbiggs. I suspect we're not exporting the region because we pick it up from the environment or config so a Pulumi K8s program doesn't need the region in the kubeconfig, but it does seem like a meaningful omission if the user intends to use the config with other tools!

To your second comment: It's true that there's not a good in-repo example for generating a k8s provider from an eks cluster resource, but you can find some instructions in our docs (eg. third code block in this section: https://www.pulumi.com/docs/clouds/aws/guides/eks/#provisioning-a-new-eks-cluster). There's also examples of this in our pulumi/examples repo (https://github.com/pulumi/examples/blob/master/aws-ts-eks-distro/index.ts)  We're thinking of renaming the examples folder in the provider repos since these are actually used for e2e testing, not really as examples.

bryantbiggs:
> because we pick it up from the environment or config so a Pulumi K8s program doesn't need the region in the kubeconfig, but it does seem like a meaningful omission if the user intends to use the config with other tools!

[...]

PR #1082, opened 2024-03-20
Do not set AWS_PROFILE env in generated kubeconfig

rquitales:
### Proposed changes

AWS_PROFILE values can vary between users despite variations providing the same level of access to the AWS resources. Always setting the profile name in the kubeconfig will mean that other users of the Pulumi program will need to ensure that their profile names also match, which isn't ideal.

This change would enable users to switch profiles, without needing to regenerate a Kubeconfig as the underlying `eks get-token` command will use the ambient `AWS_PROFILE` env var to determine the best profile.

I have manually validated that this change works locally by explicitly specifying a custom profile on the first `pulumi up`, changing the profile to another name and creating an on cluster deployment to see if the generated kubeconfig can handle the profile name change.

### Related issues (optional)

Fixes: #868


examples/utils/utils.go
```diff
@@ -652,3 +660,20 @@ func clientSetFromKubeconfig(kubeconfig any) (*kubernetes.Clientset, error) {
...
+	kc, err := json.Marshal(kubeconfig)
+	if err != nil {
+		t.Errorf("unable to marshal provided kubeconfig: %s", err)
+	}
+	kubeAccess, err := KubeconfigToKubeAccess(kc)
+	if err != nil {
+		t.Errorf("unable to create KubeAccess from kubeconfig: %s", err)
+	}
+
+	_, err = kubeAccess.Clientset.Discovery().ServerVersion()
```
thomas11, line 675: It would be good to distinguish auth errors from other errors so we don't hide those.

thomas11 (approved):
Sadly, the upgrade tests are failing again due to a new AMI id :(

examples/utils/utils.go
rquitales: I'm working on another PR to improve our e2e tests, so I'll roll this suggestion into that PR as well!

fernandocarletti:
Hello! Just a suggestion for the future. This is a breaking change, since it does not offer backward compatibility with the old behavior (I know this the hard way :). It sounds this should be released as 3.0 instead as 2.3, since minor versions should keep backward compatibility.

I'm assuming Pulumi does use semver, based on [this](https://www.pulumi.com/docs/using-pulumi/pulumi-packages/schema/#:~:text=The%20version%20must%20be%20valid%20semver.), which made me overlook it before merging the dependabot PR (and yes, I know I should have checked the changelog more careful, there's a caution note there, that's why it is a friendly suggestion and not a rant hahaha).

gunzy83:
This absolutely is a breaking change. Took me a while to track down how to deal with this on upgrade (given I am already applying [transformations](https://github.com/pulumi/pulumi-eks/issues/1038#issuecomment-2036131649) to overcome deficiencies). 

To me it makes little sense that you can create the cluster with an AWS profile (saved into the Kubeconfig for the initial deploy of the cluster resource) but then immediately creates a provider (`${name}-provider`) inside the `Cluster` resource without the profile included in its Kubeconfig (which then throws `Unable to locate credentials. You can configure credentials by running "aws configure".` even if it is not used in the program). I would assume in the same stack you would still need the profile used to create the cluster which would need to remain consistent between machines?

Even though we are using an unconventional way of accessing AWS config (using `AWS_CONFIG_FILE` to a project local dir with local and CI config files in them), I find it baffling that an org trying to deal with things at scale would at least opt for consistent profile names for environments to reduce confusion and doesn't work on my machine problems. Anyway...

rquitales:
Thanks for your feedback on this change. We've previously received comments on the previous behavior from a number of customers and how it affects their user journeys in a multi-user setup. This coupled with the discussions in https://github.com/pulumi/pulumi-eks/issues/868 and https://github.com/pulumi/pulumi-aws/issues/1906#issuecomment-1099574436 led us to revisit the behavior of the generated kubeconfig and its behavior.
I do apologize for any pain this may have caused and I'll be sure to surface this with the rest of the team such that impacts like this are minimized in the future.

Issue #1092, opened 2024-03-25
Passing cluster.kubeconfig to VpcCni fails


rquitales:
### What happened?

Passing a kubeconfig object to the `VpcCni` resource fails in actually creating the resource. The following error message occurs.

```
error: Command failed: kubectl apply -f /var/folders/xl/bqjwsf01315fs4yp20dj3x640000gn/T/tmp-35899qZQuUxifngMB.tmp
    error: error validating "/var/folders/xl/bqjwsf01315fs4yp20dj3x640000gn/T/tmp-35899qZQuUxifngMB.tmp": error validating data: failed to download openapi: Get "http://localhost:8080/openapi/v2?timeout=32s": dial tcp [::1]:8080: connect: connection refused; if you choose to ignore these errors, turn validation off with --validate=false
```

### Example

```typescript
// Create an EKS cluster with the default configuration.
const cluster = new eks.Cluster(`${projectName}-1`);

// // Create a VPC CNI resource for the cluster.
const vpcCni = new eks.VpcCni(`${projectName}-1-cni`, cluster.kubeconfig);
```

### Output of `pulumi about`

-

### Additional context

Using `cluster.kubeconfigJson` or `cluster.getKubeconfig` works however, as they are string objects. The cluster resource itself also creates a `VpcCni` sub-resource using a kubeconfig object, however that works as expected. It's likely that a codepath for a manual `VpcCni` resource is not marshalling the kubeconfig object correctly.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

Issue #1057, opened 2024-02-28
Changes to CNI yaml file not deployed when upgrading `pulumi-eks` library version


ringods:
### What happened?

The `eks.VpcCni` resource deploys a Kubernetes YAML file with CNI components. This yaml file is changed between version 2.1.0 and 2.2.0:

* https://github.com/pulumi/pulumi-eks/blob/v2.1.0/nodejs/eks/cni/aws-k8s-cni.yaml (CNI 1.11)
* https://github.com/pulumi/pulumi-eks/blob/v2.2.0/nodejs/eks/cni/aws-k8s-cni.yaml (CNI 1.16)

If you first deploy the setup from the program below with `@pulumi/eks` v2.1.0, your cluster gets the CNI v1.11 components deployed. If you upgrade this to `@pulumi/eks` v2.2.0 or higher, the expectation is that the CNI v1.16 components are deployed, but they aren't.

Related: #924 

### Example

```ts
import * as awsx from "@pulumi/awsx";
import * as eks from "@pulumi/eks";
import * as k8s from "@pulumi/kubernetes";

const vpc = new awsx.ec2.Vpc('eks-network')

// Create an EKS cluster in the VPC
const cluster = new eks.Cluster("eks-cluster", {
    vpcId: vpc.vpcId, // Use the VPC created earlier
    publicSubnetIds: vpc.publicSubnetIds, // Use the VPC public subnets for the EKS cluster
    privateSubnetIds: vpc.privateSubnetIds, // Use the VPC private subnets for the EKS cluster
    instanceType: "t2.medium", // Specify the instance type for worker nodes
    desiredCapacity: 2, // Set the desired number of worker nodes
    minSize: 1, // Set the minimum number of worker nodes
    maxSize: 3, // Set the maximum number of worker nodes
});

// Create a new NodeGroupV2 using the existing EKS cluster
const nodeGroup = new eks.NodeGroupV2("my-node-group", {
    cluster: cluster, // reference to the EKS cluster object
    version: cluster.eksCluster.version, // specify the Kubernetes version for the nodes
    instanceType: "t3.medium", // specify the instance type
    desiredCapacity: 2, // specify the desired number of nodes
    minSize: 1, // specify the minimum number of nodes
    maxSize: 3, // specify the maximum number of nodes
    amiType: "amazon-linux-2", // specify the AMI type for the nodes
});

const provider = new k8s.Provider('clusterProvider', {
[...]

PR #1093, opened 2024-03-26
Ensure CNI manifest is applied to cluster when the base manifest and provider are updated

rquitales:
### Proposed changes

This PR ensures that changes to the base CNI manifest is applied as a Pulumi update whenever the provider version changes. This is done by now also storing the computed manifest in state to diff against. Previously, only changes to user VpcCni opts caused a diff. This would mean that some users who have already updated to the latest provider version, and manually triggered a CNI update would encounter a one-off spurious diff since the computed manifest would not have been stored in state.

Changes made:

- Add applied manifest to state to fix the issue where the manifest isn't reapplied when the provider is updated
- fix the issue where VpcCni could not be passed a kubeconfig object correctly
- updated CI to use the latest version of the Pulumi binary
- Added CNIAcrossUpdates integration test
- Added CNIAcrossUpdates to the list of tests to run in CI

### Related issues (optional)

Fixes: #1057
Fixes: #1092


thomas11 (approved):
I'm not sure I have enough context and knowledge in this area to fully review this, but from my PoV it looks good.

Issue #1078, opened 2024-03-13
`aws-eks-nodeagent` image is only pulled from us-east-2 region


pierskarsenbarg:
### What happened?

When you set up a cluster, it uses [this YAML](https://github.com/pulumi/pulumi-eks/blob/990aa235eb2b2fd9001449b2032b940c8c5968d8/nodejs/eks/cni/aws-k8s-cni.yaml) to set up the CNI, and as a result we only pull images from the us-west-2 region.

Some users are reporting issues because they don't allow images to be pulled from outside a certain set of regions. (other images are pulled from those US regions, but they seem to be included already in some of the latest EKS AMIs).

### Example

n/a

### Output of `pulumi about`

```
CLI          
Version      3.109.0
Go Version   go1.22.0
Go Compiler  gc

Plugins
NAME        VERSION
aws         6.25.1
awsx        2.5.0
docker      4.5.1
dotnet      unknown
eks         2.2.1
kubernetes  4.7.1

Host     
OS       darwin
Version  14.4
Arch     arm64

This project is written in dotnet: executable='/usr/local/share/dotnet/dotnet' version='8.0.201'

Current Stack: pierskarsenbarg/cni-issue-csharp/dev

TYPE                                                 URN
pulumi:pulumi:Stack                                  urn:pulumi:dev::cni-issue-csharp::pulumi:pulumi:Stack::cni-issue-csharp-dev
pulumi:providers:aws                                 urn:pulumi:dev::cni-issue-csharp::pulumi:providers:aws::default_6_25_1
aws:iam/role:Role                                    urn:pulumi:dev::cni-issue-csharp::aws:iam/role:Role::eksNodeRole
pulumi:providers:awsx                                urn:pulumi:dev::cni-issue-csharp::pulumi:providers:awsx::default_2_5_0
awsx:ec2:Vpc                                         urn:pulumi:dev::cni-issue-csharp::awsx:ec2:Vpc::eks-vpc
pulumi:providers:eks                                 urn:pulumi:dev::cni-issue-csharp::pulumi:providers:eks::default_2_2_1
pulumi:providers:aws                                 urn:pulumi:dev::cni-issue-csharp::pulumi:providers:aws::default_6_9_0
eks:index:Cluster                                    urn:pulumi:dev::cni-issue-csharp::eks:index:Cluster::eksCluster
[...]

PR #1094, opened 2024-03-26
Allow setting custom nodeagent container images

rquitales:
### Proposed changes

This PR exposes an additional option for `VpcCni` resources. The `nodeImage` option will allow users to specify a custom `aws-eks-nodeagent` container image.

#### Changes made:

- Added logic to enable customizing nodeagent container images
- Added jest unit test to validate images can be set
- Exposed the `nodeImage` option as a customizable argument in schema
- Rebuilt SDKs

### Related issues (optional)

Fixes: #1078

### DIFF
--- a/nodejs/eks/cmd/provider/cni.ts
+++ b/nodejs/eks/cmd/provider/cni.ts
@@ -36,6 +36,7 @@ interface VpcCniInputs {
     logLevel?: string;
     logFile?: string;
     image?: string;
+    nodeAgentImage?: string;
     initImage?: string;
     vethPrefix?: string;
     eniMtu?: number;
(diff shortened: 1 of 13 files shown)


nodejs/eks/cmd/provider/cni.ts
```diff
@@ -137,7 +149,10 @@ function computeVpcCniYaml(cniYamlText: string, args: VpcCniInputs): string {
         env.push({ name: "AWS_VPC_ENI_MTU", value: "9001" });
     }
     if (args.image) {
-        daemonSet.spec.template.spec.containers[0].image = args.image.toString();
```
rquitales, line 140: note: previously the logic depended on the ordering of containers which is error-prone. I've updated the code to do a scan instead.

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1682,9 +1682,14 @@ func vpcCniProperties(kubeconfig bool) map[string]schema.PropertySpec {
 		},
 		"image": {
 			TypeSpec: schema.TypeSpec{Type: "string"},
-			Description: "Specifies the container image to use in the AWS CNI cluster DaemonSet.\n\n" +
+			Description: "Specifies the aws-node container image to use in the AWS CNI cluster DaemonSet.\n\n" +
 				"Defaults to the official AWS CNI image in ECR.",
 		},
+		"nodeImage": {
```
thomas11, line 1688: Would "nodeAgentImage" be a clearer name here? To me, `image` and `nodeImage` could be the same thing without reading the description.

thomas11 (approved):
LGTM modulo naming question

provider/cmd/pulumi-gen-eks/main.go
rquitales: `nodeAgentImage` makes it clearer. Will update the PR. Thanks!

Issue #684, opened 2022-04-22
Allow creating IPV6 clusters


davidroth:
The underlying package already added support for IPV6 via: https://www.pulumi.com/registry/packages/aws/api-docs/eks/cluster/#clusterkubernetesnetworkconfig

It would be greate if pulumi-eks would allow creating clusters with IPV6 enabled.
See also: https://aws.amazon.com/blogs/containers/amazon-eks-launches-ipv6-support/

In order to support IPv6 clusters the following are required:

- [ ] Add support for enableIpv6 in CNI config [see change in 695](https://github.com/pulumi/pulumi-eks/pull/695)
- [ ] https://github.com/pulumi/pulumi-eks/pull/675
- [ ] https://github.com/pulumi/pulumi-eks/issues/698

yandooo:
@davidroth wondering whether you had a chance to evaluate the scope of work required here? I see you're working on CNI [ipv6 support](https://github.com/pulumi/pulumi-eks/pull/683).

davidroth:
@DCAL12 Unfortunately it is not. I dont know why this issue has been closed. See also the following issue which tracks the ipFamily support: https://github.com/pulumi/pulumi-eks/issues/698

See also [this MR](https://github.com/pulumi/pulumi-eks/pull/700) which was closed without being merged. 😕

rquitales:
Support for this feature should now be enabled in v2.3.0 of the EKS provider!

PR #1089, opened 2024-03-25
Expose Ip-Family for IPV6

davidroth:
Fixes https://github.com/pulumi/pulumi-eks/issues/698
Based on https://github.com/pulumi/pulumi-eks/pull/888 and rebased onto master

PR #888, opened 2023-06-20
Re-apply "expose ipFamily"

mikhailshilkov:
Re-applies pulumi/pulumi-eks#882 to fix #698 but now with green build (eventually)

PR #1098, opened 2024-03-26
Expose Ip-Family for IPV6 (#1089)

rquitales:
### Proposed changes
This PR exposes the `ipFamily` argument to enable creating EKS clusters that support ipv6 networking.
Modified from https://github.com/pulumi/pulumi-eks/pull/888 and rebased on top of latest changes in master.

An additional cluster with ipv6 is also created to test that cluster creation works as expected.

### Related issues (optional)
Closes: #698
Closes: #684


davidroth:
@rquitales It looks like the acceptance tests have been successfully completed.🥳
Happily awaiting this finally being merged (I absolutely need this change as soon as possible for a new project.) 🚀

examples/cluster-go/main.go
```diff
@@ -39,10 +45,79 @@ func main() {
...
+
+			subnetIDs = append(subnetIDs, subnet.ID().ToStringOutput())
+		}
+
+		// 3. Create an EKS cluster with IPv6 networking enabled.
+		cluster4, err := eks.NewCluster(ctx, "example-cluster-4", &eks.ClusterArgs{
+			IpFamily:         pulumi.StringPtr("ipv6"),
+			VpcId:            vpc.ID().ToStringOutput(),
+			SubnetIds:        pulumi.StringArray(subnetIDs),
+			UseDefaultVpcCni: func() *bool { t := true; return &t }(),
```
EronWright, line 89: This line seems unusual, could it be `pulumi.BoolPtr(true)`?

examples/cluster-go/main.go
```diff
@@ -28,6 +32,8 @@ func main() {
 			return err
 		}
 
+		// Create cluster that don't conflict between defaults and mutually exclusive arguments
```
EronWright, line 35: Can you rephrase this comment?

EronWright (approved):

examples/cluster-go/main.go
```diff
@@ -39,10 +45,79 @@ func main() {
...
+		var subnetIDs []pulumi.StringInput
+		for idx, az := range []string{"us-west-2a", "us-west-2b"} {
+			subnet, err := ec2.NewSubnet(ctx, fmt.Sprintf("ipv6-subnet-%d", idx), &ec2.SubnetArgs{
+				VpcId:                       vpc.ID().ToStringOutput(),
+				AssignIpv6AddressOnCreation: pulumi.Bool(true),
+				// EnableDns64:                             pulumi.Bool(true),
+				// EnableResourceNameDnsAaaaRecordOnLaunch: pulumi.Bool(true),
+				AvailabilityZone: pulumi.String(az),
+				CidrBlock:        pulumi.String(fmt.Sprintf("10.100.%d.0/24", len(subnetIDs))),
+				// Ipv6Native:    pulumi.Bool(true), // Create a ipv6 only subnet
```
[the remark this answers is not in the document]
rquitales: This was me trying out different configuration options. I'll remove it since it's not necessary.

examples/cluster-go/main.go
rquitales: `UseDefaultVpcCni` requires a `*bool` type, and that's because we define it as a plain type in schema (ref: https://github.com/pulumi/pulumi-eks/blob/449635cb9dfdfbdc6a03b2d0c3b8b12f96420f13/provider/cmd/pulumi-gen-eks/main.go#L319) when the MLCs were created for this provider. This is an artifact of the nodejs types driving the MLC types, and ideally should really not be a plain type here.

rquitales:
Note: the previously passed all pre-submits. A new commit was added that only modified comments (https://github.com/pulumi/pulumi-eks/pull/1098/commits/6c3675e153ee6a729ef26975415bff49988a2e88), so I'll be force pushing it since retesting requires another 30-45 mins to run. We also have post-submit CI runs that also exercise the same tests.

Issue #1048, opened 2024-02-26
eks.cluster resource needs examples and prop descriptions


sean1588:
The EKS cluster resource contains no code examples. I did not see any defined in the schema file either. We should author some code examples as this is one of our most highly trafficked resource pages and our most used component resource.

We are also missing the property description for the results prop of [getKubeConfig](https://www.pulumi.com/registry/packages/eks/api-docs/cluster/#method_GetKubeconfig_result). Would be great if we could get this resolved as well.

There are also some missing property descriptions for some of the nested types, we need to author these as well: https://www.pulumi.com/registry/packages/eks/api-docs/cluster/#coredata

PR #1112, opened 2024-04-01
add basic eks cluster example

sean1588:
related to: https://github.com/pulumi/pulumi-eks/issues/1048

Adds a basic example to the eks cluster resource page, since this page is missing examples. I took this from the pulumi-eks guide: https://www.pulumi.com/docs/clouds/aws/guides/eks/

I also added in the missing descriptions text for the [CoreData](https://www.pulumi.com/registry/packages/eks/api-docs/cluster/#coredata) nested type.

Screenshot from 2024-04-01 15-03-06


blampe:
@sean1588 I think you'll need to update the docstring [here](https://github.com/pulumi/pulumi-eks/blob/master/provider/cmd/pulumi-gen-eks/main.go#L152-L154) instead of in `schema.json`, and then you'll need to regenerate SDKs to reflect the new description. Running `make` (after you update `main.go`) should do the trick.

sean1588:
ahhh ok that makes sense. thanks so much @blampe!

sean1588:
thanks @blampe, I updated to move the docs string to the main.go file. I also added in the missing descriptions for the [CoreData](https://www.pulumi.com/registry/packages/eks/api-docs/cluster/#coredata) supporting type.

mjeffryes:
Looks like this is really close, but we still need to generate java and nodejs SDKs

blampe:
Looks like a flake, re-running.

> error downloading provider aws to file: failed to download plugin: aws-6.0.4: write /tmp/pulumi-plugin-tar1375576335: no space left on device

blampe (approved):
Thanks for this improvement!

I didn't look closely at the example, but if it matches what's in the guide then it's probably fine.

It would be nice to be able to more easily keep these docstrings in sync with examples like that, but we don't need to solve that here.

sean1588:
Thanks @blampe!! Yeah I took the example straight from the guide, but yeah would be great if we had a way to keep these in sync.

Issue #1113, opened 2024-04-01
Pulumi fails with execption when using CreationRoleProvider in C#


automagic:
### What happened?

When attempting to use the CreationRoleProvider, pulumi will fail with an exception.  Below is a full reproduction.

```
Do you want to perform this update? yes
Updating (dev)

View in Browser (Ctrl+O): https://app.pulumi.com/initech/webserver-asg/dev/updates/7

     Type                 Name               Status     Info
     pulumi:pulumi:Stack  webserver-asg-dev             'dotnet build -nologo .' completed successfully
     Type                             Name                       Status                  Info
 +   pulumi:pulumi:Stack              webserver-asg-dev          **creating failed**     1 error; 127 messages
 +   ├─ aws:iam:Role                  eks-creator                created (0.94s)         
 +   │  └─ pulumi:providers:aws       cluster-creation-provider  created (0.15s)         
 +   ├─ aws:iam:Policy                eks-creator-policy         created (0.94s)         
 +   └─ aws:iam:RolePolicyAttachment  eksRolePolicyAttachment    created (0.40s)         

Policies:
    ✅ aws-iso27001-compliance-ready-policies-typescript@v0.0.1

Diagnostics:
  pulumi:pulumi:Stack (webserver-asg-dev):
    TypeError: provider.getPackage is not a function: TypeError: provider.getPackage is not a function
        at convertToProvidersMap (/snapshot/eks/node_modules/@pulumi/pulumi/resource.js:334:25)
        at new Resource (/snapshot/eks/node_modules/@pulumi/pulumi/resource.js:205:133)
        at new CustomResource (/snapshot/eks/node_modules/@pulumi/pulumi/resource.js:400:9)
        at new Cluster (/snapshot/eks/node_modules/@pulumi/aws/eks/cluster.js:200:9)
        at createCore (/snapshot/eks/bin/cluster.js:331:24)
        at createCluster (/snapshot/eks/bin/cluster.js:827:18)
        at new ClusterInternal (/snapshot/eks/bin/cluster.js:906:25)
        at Object.construct (/snapshot/eks/bin/cmd/provider/cluster.js:21:29)
        at Provider.construct (/snapshot/eks/bin/cmd/provider/index.js:131:24)
        at Server.<anonymous> (/snapshot/eks/node_modules/@pulumi/pulumi/provider/server.js:315:52)

[...]

PR #1121, opened 2024-04-12
Throw error when `creationRoleProvider` is used in MLC programs

rquitales:
### Proposed changes

The `creationRoleProvider` option in the nodejs version of this provider allows passing a Pulumi AWS provider. However, for MLC programs, this isn't supported since only the JSON serialized provider is passed, not a fully initialized instance.

A user attempting to use this option will encounter a non-helpful error message: `Detail="provider.getPackage is not a function")`

This PR updates the provider to determine if the provider is being passed, and if so, returns a more helpful message on how to configure the EKS cluster settings.

### Related issues (optional)

Closes: #1113


examples/cluster-go/step2/main.go
```diff
@@ -0,0 +1,150 @@
...
+		if err != nil {
+			return err
+		}
+
+		////////////////////////////////////////////////
+		///     CreationRoleProvider should fail     ///
+		////////////////////////////////////////////////
+
+		// Create AWS provider
+		awsProv, err := aws.NewProvider(ctx, "aws", nil)
```
rquitales, line 103: The new resource being added in step 2.

nodejs/eks/cluster.ts
```diff
@@ -1905,6 +1905,12 @@ export class ClusterInternal extends pulumi.ComponentResource {
 
         super(type, name, args, opts);
 
+        if (args?.creationRoleProvider?.provider) {
+            throw new Error(
+                "The `creationRoleProvider.provider` option is not supported in non-nodejs Pulumi programs. Please use the `providerCredentialOpts` option instead.",
+            );
+        }
```
iwahbe, line 1912: Is there an issue for serializing resources in nodejs MLCs in Pulumi? If so, can we link it. If not, can we open the feature request and link it?

t0yv0:
> However, for MLC programs, this isn't supported since only the JSON serialized provider is passed, not a fully initialized instance.

I am a little surprised at that, since we have resource references that represent resources on the wire.  I was working on the feature here that allows a component to configure a Provider and then the program gets it through a method and uses that Provider to configure more resources: 

https://github.com/pulumi/pulumi/blob/master/tests/integration/construct_component_configure_provider/nodejs/index.ts that

t0yv0:
Is this this one https://www.pulumi.com/registry/packages/eks/api-docs/clustercreationroleprovider/ 

https://www.pulumi.com/registry/packages/eks/api-docs/cluster/#creationroleprovider_nodejs

t0yv0:
I think this change is not unreasonable but I'd double-check if other languages work. Presumably Node works. C# does not work. Go and Python may just work actually? If not let's indeed ticket a feature in p/p - there's very likely a bug of some sort because resource references are supposed to travel over the wire just fine, including provider references; perhaps the gap is C# specific .

rquitales:
> I think this change is not unreasonable but I'd double-check if other languages work. Presumably Node works. C# does not work. Go and Python may just work actually? If not let's indeed ticket a feature in p/p - there's very likely a bug of some sort because resource references are supposed to travel over the wire just fine, including provider references; perhaps the gap is C# specific .

I've confirmed that this issue affects all go, python and c#. I didn't get a chance to test java, but I would assume it should be the same outcome. Unfortunately, using the [ClusterCreationRoleProvider](https://www.pulumi.com/registry/packages/eks/api-docs/clustercreationroleprovider/) also doesn't work, since it would be a type mismatch resulting in `unknown resource type eks:index:ClusterCreationRoleProvider`.

t0yv0 (approved):
This is the right fix then thank you for confirming. Looking a bit to see how can we extract a good ticket for Pulumi Core here, what's going wrong...

PR #1131, opened 2024-04-24
Expose KubeletExtraArgs and BootstrapExtraArgs for managed nodegroups

rquitales:
### Proposed changes

This PR enables setting kubelet and bootstrap extra args by creating a custom launch template if the arguments are supplied. Existing ManagedNodeGroups will not have a diff since we only create and use a custom launch template if these options are used by the Pulumi program.

#### Tests:

- Updated existing NodeJS and Golang test cases to ensure MNGs can supply kubelet extra args and the launch template is used
- Manually tested and verified within the AWS console that the user data is set correctly when these options are supplied

### Related issues (optional)
Closes: #611

### DIFF
--- a/examples/managed-nodegroups/index.ts
+++ b/examples/managed-nodegroups/index.ts
@@ -22,6 +22,7 @@ export const kubeconfig = cluster.kubeconfig;
 const managedNodeGroup0 = eks.createManagedNodeGroup("example-managed-ng0", {
     cluster: cluster,
     nodeRole: role0,
+    kubeletExtraArgs: "--max-pods=500",
 });
 
 // Create a simple AWS managed node group using a cluster as input and the
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -222,7 +222,7 @@ export interface NodeGroupBaseOptions {
     kubeletExtraArgs?: string;
 
     /**
-     * Additional args to pass directly to `/etc/eks/bootstrap.sh`.  Fror details on available options, see:
+     * Additional args to pass directly to `/etc/eks/bootstrap.sh`.  For details on available options, see:
      * https://github.com/awslabs/amazon-eks-ami/blob/master/files/bootstrap.sh.  Note that the `--apiserver-endpoint`,
      * `--b64-cluster-ca` and `--kubelet-extra-args` flags are included automatically based on other configuration
      * parameters.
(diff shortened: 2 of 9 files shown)


rquitales:
Upgrade tests are failing due to kubeconfig diff and AMI ID diffs (see #1145). This is expected, and no other diffs are observed.

examples/managed-nodegroups/index.ts
```diff
@@ -22,6 +22,7 @@ export const kubeconfig = cluster.kubeconfig;
 const managedNodeGroup0 = eks.createManagedNodeGroup("example-managed-ng0", {
     cluster: cluster,
     nodeRole: role0,
+    kubeletExtraArgs: "--max-pods=500",
```
EronWright, line 25: (nit: examples have different values for max-pods)

blampe (commented):
edit: whoops ignore me

provider/cmd/pulumi-resource-eks/schema.json
```diff
@@ -980,6 +980,11 @@
                     "type": "string",
                     "description": "Type of Amazon Machine Image (AMI) associated with the EKS Node Group. Defaults to `AL2_x86_64`. See the AWS documentation (https://docs.aws.amazon.com/eks/latest/APIReference/API_Nodegroup.html#AmazonEKS-Type-Nodegroup-amiType) for valid AMI Types. This provider will only perform drift detection if a configuration value is provided."
                 },
+                "bootstrapExtraArgs": {
+                    "type": "string",
```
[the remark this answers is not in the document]
rquitales: It would, but I'm maintaining consistency with the existing `NodeGroup` resource which uses string inputs.
Ref: https://www.pulumi.com/registry/packages/eks/api-docs/nodegroup/#kubeletextraargs_nodejs

nodejs/eks/nodegroup.ts
```diff
@@ -1422,6 +1422,24 @@ export type ManagedNodeGroupOptions = Omit<
      */
     clusterName?: pulumi.Output<string>;
 
+    /**
+     * Extra args to pass to the Kubelet.  Corresponds to the options passed in the `--kubeletExtraArgs` flag to
+     * `/etc/eks/bootstrap.sh`.  For example, '--port=10251 --address=0.0.0.0'.
```
t0yv0, line 1427: Can you explain how to escape values that contain spaces and special characters here?

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -830,6 +831,30 @@ func generateSchema() schema.PackageSpec {
 					"version": {
 						TypeSpec: schema.TypeSpec{Type: "string"},
 					},
+					"kubeletExtraArgs": {
```
t0yv0, line 834: I'm remembering now that TF has a ConflictsWith helper which we don't (yet?) have, could be a useful shorthand for marking mutually exclusive fields.

t0yv0 (approved):
The changes look good to me code-wise. I don't have a good concept of the use-cases here though FWIW. I had the same question on whether string or `[]string` (or both?) is user friendly, sounds like it's beeing considered.

rquitales:
Will be leaving the input as a string in this PR to maintain consistency with the other nodegroup resource to make it less jarring for users managing both types of resources. We can look into extending the type to be a string array in the future.

Issue #587, opened 2021-05-26
EKS Add-on support (coredns, etc)


almson:
EKS now supports coredns, kube-proxy, and VPC CNI as managed add-ons. They're exposed through the API. It would be great to be able to declare them and update them using Pulumi.

https://aws.amazon.com/about-aws/whats-new/2021/05/eks-add-ons-now-support-coredns-kube-proxy/

lukehoban:
Great suggestion - would be great to add these to the `@pulumi/eks` API directly.

Issue #253, opened 2019-09-12
Support CoreDNS and KubeProxy updates


gitfool:
I'm very impressed with the amount of customization already in place for the [AWS VPC CNI](https://github.com/aws/amazon-vpc-cni-k8s). It's much appreciated.

Similarly, it would be nice to support updating [CoreDNS](https://github.com/coredns/coredns) and [kube-proxy](https://github.com/kubernetes/kube-proxy) as part of infra deploy, since Amazon does not update these automatically when updating an existing cluster.

References:
- https://docs.aws.amazon.com/eks/latest/userguide/update-cluster.html
> Amazon EKS does not modify any of your Kubernetes add-ons when you update a cluster. After updating your cluster, we recommend that you update your add-ons to the versions listed in the following table for the new Kubernetes version that you're updating to (steps to accomplish this are included in the update procedures).

(I've had to do this in the past when upgrading cluster versions in-place, so it really is needed.)

---

Depends on https://github.com/pulumi/pulumi-kubernetes/issues/264

MatiasNMont:
HI! I need to do this, but I do not know how to do this. There are some solutions?

stack72:
@roothorp can you think about how we can either expose a new option on EKS for managing add-ons OR if this is something that can be achieved using Kubernetes SSA as per https://www.pulumi.com/registry/packages/kubernetes/how-to-guides/managing-resources-with-server-side-apply/ and then we can write a guide on this?

PR #1153, opened 2024-05-16
Add Addon resource to EKS provider

rquitales:
### Proposed Changes

In late 2020, AWS introduced EKS managed addons, which allow AWS to handle the management of cluster addons as part of the EKS lifecycle. This pull request integrates the upstream `aws.eks.addon` resource into our EKS provider, enhancing visibility and management of this resource.

By incorporating this resource, users will be able to manage addons such as coredns and kube-proxy, along with other supported addons. We chose not to create specific Pulumi resources for each addon, as doing so and exposing schema configurations as resource properties could be fragile with future schema changes. Additionally, user requests to create CoreDNS and kube-proxy resources were proposed before AWS managed addons were available, when self-management was necessary to keep these components up to date. VpcCni is another example of an addon that we also intend to transition to AWS management.

#### Testing:

- Conducted manual testing to verify the successful installation and upgrade of new addons within a cluster
- Updated existing TypeScript and MLC (Python) tests to include the installation of `coredns` as an addon

### Related issues (optional)
Closes: #587
Closes: #253


examples/cluster-py/__main__.py
```diff
@@ -13,34 +13,51 @@
...
-                            vpc_id=vpc.vpc_id,
-                            public_subnet_ids=vpc.public_subnet_ids,
-                            node_group_options=eks.ClusterNodeGroupOptionsArgs(
-                                desired_capacity=1,
-                                min_size=1,
-                                max_size=1,
-                                instance_type="t3.small"
-                            )
+cluster2 = eks.Cluster(
+    "eks-cluster",
```
blampe, line 17: Not sure what's up the resource naming here, but should this be `f"{project_name}-2"` to match the others?

I'm assuming it's to prevent conflicts, but wouldn't this be auto-named? 🤷

rquitales:
Tests are failing due to orphaned AWS resources. Will need to run the cleanup script before re-running the test suite.

nodejs/eks/addon.ts
```diff
@@ -0,0 +1,54 @@
...
+import * as aws from "@pulumi/aws";
+import { Cluster, ClusterInternal } from "./cluster";
+
+/**
+ * AddonOptions describes the configuration options available for the Amazon VPC CNI plugin for Kubernetes. This is obtained from
+ * upstream `aws.eks.AddonArgs` and removes the deprecated `resolveConflicts` field. `clusterName` is also removed as we enable users to
+ * pass in the cluster object directly.
+ */
+export interface AddonOptions extends Omit<aws.eks.AddonArgs, "resolveConflicts" | "clusterName"> {
+    cluster: Cluster | ClusterInternal;
```
blampe, line 25: Out of curiosity, would I be able to use an `aws.eks.Cluster` here?

I can see a use case where maybe you didn't start with pulumi-eks but there's still some convenience it can provide on top of what you currently have. Edit: although looking at it more, it's not doing much on top so this wouldn't be very valuable.

blampe (approved):

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -961,6 +961,56 @@ func generateSchema() schema.PackageSpec {
...
+					},
+					"cluster": {
+						TypeSpec: schema.TypeSpec{
+							Ref: "#/resources/eks:index:Cluster",
+						},
+						Description: "The target EKS cluster.",
+					},
+					"configurationValues": {
+						TypeSpec:    schema.TypeSpec{Type: "string"},
+						Description: "Custom configuration values for addons with single JSON string. This JSON string value must match the JSON schema derived from describe-addon-configuration.",
```
flostadler, line 987: Why don’t we change this to a map and serialize it for users?

nodejs/eks/addon.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Addon manages an EKS add-on.
+ * For more information about supported add-ons, see: https://docs.aws.amazon.com/eks/latest/userguide/eks-add-ons.html
+ */
+export class Addon extends pulumi.ComponentResource {
+    public readonly addon: pulumi.Input<aws.eks.Addon>;
+
+    constructor(name: string, args: AddonOptions, opts?: pulumi.CustomResourceOptions) {
+        const cluster = args.cluster;
+
+        super("eks:index:Addon", name, args, {
```
flostadler, line 38: What’s the reason for making addons their own component vs making that an input to the eks cluster component?

This one is mostly wrapping the aws classic addon resource right now. Including it in the EKS component would make discoverability easier

flostadler: I thought some more about this and actually prefer this approach. The cluster resource is already quite overloaded with parameters and the addon is very self-contained.

provider/cmd/pulumi-gen-eks/main.go
rquitales: Updated this to be a map and added additional integration tests for this.

t0yv0:
Quick question about these addons.

We have some related code in pulumi-aws:

https://github.com/pulumi/pulumi-aws/issues/3009
https://github.com/pulumi/pulumi-aws/issues/2720
https://github.com/pulumi/pulumi-aws/issues/2732

Does this change look good in light of this code? Does it interact with defaultAddonsToRemove or should it? Thanks.

nodejs/eks/addon.ts
```diff
@@ -0,0 +1,57 @@
...
+            ...opts,
+            parent: cluster,
+        });
+
+        const addon = new aws.eks.Addon(
+            name,
+            {
+                ...args,
+                clusterName: cluster.core.cluster.name,
```
t0yv0, line 49: So, question, genuinely curious what's the benefit of wrapping  aws.eks.Addon in a component Addon?  

Can users use a vanilla aws.eks.Addon right now already without this helper?

Perhaps we have some more functionality we expect to land in this wrapper? 

Is this automatically setting `parent`? Or removing deprecated properties? Or is it automatic JSON.stringify?

Can we bake any of this directly in AWS Addon or a bad idea?

nodejs/eks/addon.ts
```diff
@@ -0,0 +1,57 @@
...
+                configurationValues: JSON.stringify(args.configurationValues),
+            },
+            { parent: this, provider: opts?.provider },
```
t0yv0, line 52: What if we have `provders: {"aws": awsProv}`? I think actually this might work already and `provider:*` might be redundant as it will inherit via `parent: this`, we were investigating resetting AWS versions underneath the awsx provider and https://github.com/pulumi/pulumi-awsx?tab=readme-ov-file#custom-aws-provider-versions it works fine while all the provider does is setting `{parent: this}`.

t0yv0 (approved):
The PR Looks great, thank you, learned a few TypeScript tricks here.

One higher level question I have is why we decided to wrap aws.eks.Addon in the first place, perhaps we can document internally for posterity, it looks like some thought was put into it though. I'd just appreciate reading into it more to inform future decisions of this sort.

Issue #1145, opened 2024-05-09
Upgrade provider snapshots to v2.4.0 of provider when released


rquitales:
### What happened?

#1136 updated the kubeconfig args to ensure that the awscli always outputs in JSON format. We'll need to update our baseline provider snapshots to account for this as our snapshot tests will always fail.

The next minor release will also contain a few other diffs related to addons, so we'll batch these updates into a single minor release to minimize state updates for the user.

In the longterm, we'd want to ignore update diffs, but error on replace diffs for scenarios like this.

### Example

-

### Output of `pulumi about`

-

### Additional context

-

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
Self-tagging @t0yv0

PR #1154, opened 2024-05-16
Update upgrade test snapshots to v2.5.1 of provider

rquitales:
This PR contains commits for 3 changes:

1. Updates the baseline version to v2.5.0
2. Updates the provider snapshots using the new baseline
3. Reduce parallelism in upgrade tests to 3 as a workaround for #1155

It is intended for this PR to be rebased and merged so that we can easily rollback any of these individual commits, if necessary.

Closes: #1145

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ PROJECT_NAME := Pulumi Amazon Web Services (AWS) EKS Components
 
 VERSION         := $(shell pulumictl get version)
 TESTPARALLELISM := 12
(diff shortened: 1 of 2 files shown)


rquitales:
Ah, looks like the upgrade test now fails because we just merged in a new field support so the older version of the provider doesn't recognize it.

blampe (approved):

flostadler:
I guess you now need to re-record the tests? :/

PR #1163, opened 2024-05-23
fix: use mime-multipart for launch template

rquitales:
### Proposed changes

Uses best practices from https://docs.aws.amazon.com/batch/latest/userguide/launch-templates.html to set the LaunchTemplate configuration. Tests were passing in CI, but failing in the upgrade snapshots workflow.

### Related issues (optional)

Fixes: #1162

### DIFF
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -1770,12 +1770,18 @@ function createMNGCustomLaunchTemplate(
                 args.clusterName,
             ])
             .apply(([clusterName, clusterEndpoint, clusterCertAuthority, argsClusterName]) => {
-                return `#!/bin/bash
+                return `MIME-Version: 1.0
+Content-Type: multipart/mixed; boundary="==MYBOUNDARY=="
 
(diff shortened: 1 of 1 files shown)


nodejs/eks/nodegroup.ts
```diff
@@ -1770,12 +1770,18 @@ function createMNGCustomLaunchTemplate(
...
+
+--==MYBOUNDARY==
+Content-Type: text/x-shellscript; charset="us-ascii"
+
+#!/bin/bash
 
-            /etc/eks/bootstrap.sh --apiserver-endpoint "${clusterEndpoint}" --b64-cluster-ca "${clusterCertAuthority}" "${
+/etc/eks/bootstrap.sh --apiserver-endpoint "${clusterEndpoint}" --b64-cluster-ca "${clusterCertAuthority}" "${
                     argsClusterName || clusterName
                 }"${bootstrapExtraArgs}
```
EronWright, line 1783: Seems like these lines should be unindented.

EronWright (approved):

nodejs/eks/nodegroup.ts
rquitales: These lines were indented from the linter. Since they're also typescript variables, the final text will be formatted as a single line.

PR #1166, opened 2024-05-24
fix: always set AMI ID for custom LaunchTemplates with ManagedNodeGroups

rquitales:
### Proposed changes

#1163 enabled setting custom user data in MIME format. However, AWS will attempt to merge the user data with its own copy if an AMI ID is not supplied in the LaunchTemplate. This results in the nodes being unable to join the cluster due to conflicting bootstrap script runs. This PR explicitly sets the ami id within the launch template.

Testing:

- Updated the existing tests to provision a cluster within a VPC to surface any node joining issues which CI was previously ignoring

### DIFF
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -1800,6 +1800,9 @@ Content-Type: text/x-shellscript; charset="us-ascii"
         {
             userData,
             metadataOptions,
+            // We need to always supply an imageId, otherwise AWS will attempt to merge the user data which will result in
+            // nodes failing to join the cluster.
+            imageId: getRecommendedAMI(args, core.cluster.version, parent),
         },
         { parent, provider },
     );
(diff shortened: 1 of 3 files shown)


nodejs/eks/nodegroup.ts
```diff
@@ -1800,6 +1800,9 @@ Content-Type: text/x-shellscript; charset="us-ascii"
         {
             userData,
             metadataOptions,
+            // We need to always supply an imageId, otherwise AWS will attempt to merge the user data which will result in
+            // nodes failing to join the cluster.
+            imageId: getRecommendedAMI(args, core.cluster.version, parent),
```
flostadler, line 1805: Do we want to enable overriding the AMI here?
Auto upgrading nodes might not be something users want to do

rquitales: That makes sense. WDYT about exposing that in another PR to reduce scope on this PR?

flostadler (approved):

Issue #1172, opened 2024-05-29
Unable to specify `nodeRootVolumeIops` and `nodeRootVolumeThroughput` in node group resources


rquitales:
### What happened?

When specifying these numeric inputs, a validation error occurs stating that the inputs are not numeric, despite them being numeric.

It appears that our regex check is wrong here:
https://github.com/pulumi/pulumi-eks/blob/161e7af6803b58fea43ffdc3c6f9ee85cf658134/nodejs/eks/nodegroup.ts#L758

It should be `^\\d+$`.

### Example

```typescript
const cluster = new eks.Cluster(`${projectName}`, {
    vpcId: vpc.vpcId,
    publicSubnetIds: vpc.publicSubnetIds,
    nodeGroupOptions: {
        desiredCapacity: 1,
        minSize: 1,
        maxSize: 1,
        nodeRootVolumeThroughput: 125,
        nodeRootVolumeType: "gp3",
        enableDetailedMonitoring: false,
    }
})
```

Stderr:

```shell
  pulumi:pulumi:Stack (example-cluster-dev):
    warning: using pulumi-resource-eks from $PATH at /Users/rquitales/code/providers/pulumi-eks/bin/pulumi-resource-eks
    error: Running program '/Users/rquitales/code/providers/pulumi-eks/examples/cluster/' failed with an unhandled exception:
    Error: Cannot create a cluster node root volume of gp3 type without provisioned throughput (nodeRootVolumeThroughput) as integer value.
        at /Users/rquitales/code/providers/pulumi-eks/nodejs/eks/nodegroup.ts:789:27
        at /Users/rquitales/code/providers/pulumi-eks/nodejs/eks/node_modules/@pulumi/output.ts:398:31
        at Generator.next (<anonymous>)
        at /Users/rquitales/code/providers/pulumi-eks/nodejs/eks/node_modules/@pulumi/pulumi/output.js:21:71
        at new Promise (<anonymous>)
        at __awaiter (/Users/rquitales/code/providers/pulumi-eks/nodejs/eks/node_modules/@pulumi/pulumi/output.js:17:12)
        at applyHelperAsync (/Users/rquitales/code/providers/pulumi-eks/nodejs/eks/node_modules/@pulumi/pulumi/output.js:239:12)
        at /Users/rquitales/code/providers/pulumi-eks/nodejs/eks/node_modules/@pulumi/output.ts:310:13
```

### Output of `pulumi about`

-

### Contributing

[...]

Issue #1160, opened 2024-05-22
Calling [toString] on an [Output<T>] is not supported


polkx:
### What happened?

When i use pulumi eks i get below error:

```
 Error: Invalid EC2 instance type: Calling [toString] on an [Output<T>] is not supported.
    To get the value of an Output<T> as an Output<string> consider either:
    1: o.apply(v => `prefix${v}suffix`)
    2: pulumi.interpolate `prefix${v}suffix`
    See https://www.pulumi.com/docs/concepts/inputs-outputs for more details.
    This function may throw in a future version of @pulumi/pulumi.: Error: Invalid EC2 instance type: Calling [toString] on an [Output<T>] is not supported.
    To get the value of an Output<T> as an Output<string> consider either:
    1: o.apply(v => `prefix${v}suffix`)
    2: pulumi.interpolate `prefix${v}suffix`
    See https://www.pulumi.com/docs/concepts/inputs-outputs for more details.
    This function may throw in a future version of @pulumi/pulumi.
        at isGravitonInstance (/snapshot/eks/bin/nodegroup.js:979:15)
        at getAMIType (/snapshot/eks/bin/nodegroup.js:998:25)
        at getRecommendedAMI (/snapshot/eks/bin/nodegroup.js:950:21)
        at createNodeGroupInternal (/snapshot/eks/bin/nodegroup.js:297:33)
        at createNodeGroup (/snapshot/eks/bin/nodegroup.js:155:12)
        at createCluster (/snapshot/eks/bin/cluster.js:842:60)
        at new ClusterInternal (/snapshot/eks/bin/cluster.js:906:25)
        at Object.construct (/snapshot/eks/bin/cmd/provider/cluster.js:21:29)
        at Provider.construct (/snapshot/eks/bin/cmd/provider/index.js:131:24)
        at Server.<anonymous> (/snapshot/eks/node_modules/@pulumi/pulumi/provider/server.js:315:52)
```

I think the problem is there:
https://github.com/pulumi/pulumi-eks/blob/b6200bbf073449479c2a47b52e96f1e79d89e1f4/nodejs/eks/nodegroup.ts#L1860

You get `Input<string> | undefined` and then you eliminate `undefined` but on `Input<string>` call `.toString()`

### Example

I use Scala language, Besom (Scala SDK for Pulumi) and scala-cli.
If you want to run the example below, you'll need to install the scala lang plugin. 
[...]

PR #1173, opened 2024-05-29
Handle arg inputs as potential Pulumi Outputs

rquitales:
### Proposed changes

This PR wraps all arg inputs that call `.toString()` with `pulumi.all` incase a `pulumi.Output` value is supplied instead of a `pulumi.Input` or plain type. As this is a MLC component, we needsto correctly deal with output values as we can't just assume that the engine sends Input<string> as a plain string.

##### Tests done:

- Updated existing test case to supply a pulumi.Output<string> instead of a plain string
- Manual testing

### Related issues (optional)

Fixes: #1160
Fixes: #1172

### DIFF
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -757,33 +757,38 @@ ${customUserData}
 
     const numeric = new RegExp("^\\d+$");
 
-    if (args.nodeRootVolumeIops && args.nodeRootVolumeType !== "io1") {
-        throw new Error(
-            "Cannot create a cluster node root volume of non-io1 type with provisioned IOPS (nodeRootVolumeIops).",
-        );
-    }
+    // We need to wrap the validation in a pulumi.all as MLCs could supply pulumi.Output<T> or T.
+    pulumi
+        .all([args.nodeRootVolumeIops, args.nodeRootVolumeType, args.nodeRootVolumeThroughput])
+        .apply(([nodeRootVolumeIops, nodeRootVolumeType, nodeRootVolumeThroughput]) => {
+            if (nodeRootVolumeIops && nodeRootVolumeType !== "io1") {
+                throw new Error(
+                    "Cannot create a cluster node root volume of non-io1 type with provisioned IOPS (nodeRootVolumeIops).",
+                );
+            }
 
-    if (args.nodeRootVolumeType === "io1" && args.nodeRootVolumeIops) {
-        if (!numeric.test(args.nodeRootVolumeIops?.toString())) {
-            throw new Error(
-                "Cannot create a cluster node root volume of io1 type without provisioned IOPS (nodeRootVolumeIops) as integer value.",
-            );
-        }
-    }
+            if (nodeRootVolumeType === "io1" && nodeRootVolumeIops) {
+                if (!numeric.test(nodeRootVolumeIops?.toString())) {
+                    throw new Error(
+                        "Cannot create a cluster node root volume of io1 type without provisioned IOPS (nodeRootVolumeIops) as integer value.",
+                    );
+                }
+            }
 
-    if (args.nodeRootVolumeThroughput && args.nodeRootVolumeType !== "gp3") {
-        throw new Error(
-            "Cannot create a cluster node root volume of non-gp3 type with provisioned throughput (nodeRootVolumeThroughput).",
-        );
-    }
+            if (nodeRootVolumeThroughput && nodeRootVolumeType !== "gp3") {
+                throw new Error(
+                    "Cannot create a cluster node root volume of non-gp3 type with provisioned throughput (nodeRootVolumeThroughput).",
+                );
+            }
 
-    if (args.nodeRootVolumeType === "gp3" && args.nodeRootVolumeThroughput) {
-        if (!numeric.test(args.nodeRootVolumeThroughput?.toString())) {
-            throw new Error(
-                "Cannot create a cluster node root volume of gp3 type without provisioned throughput (nodeRootVolumeThroughput) as integer value.",
-            );
-        }
-    }
+            if (nodeRootVolumeType === "gp3" && nodeRootVolumeThroughput) {
+                if (!numeric.test(nodeRootVolumeThroughput?.toString())) {
+                    throw new Error(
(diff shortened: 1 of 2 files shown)


nodejs/eks/nodegroup.ts
```diff
@@ -1867,24 +1880,26 @@ export function isGravitonInstance(instanceType: string): boolean {
...
 function getAMIType(
-    amiType: pulumi.Input<string> | undefined,
-    gpu: pulumi.Input<boolean> | undefined,
-    instanceType: pulumi.Input<string> | undefined,
-): string {
-    if (amiType) {
-        // Return the user-specified AMI type.
-        return amiType.toString();
-    }
+    amiType: pulumi.Input<string> | pulumi.Output<string> | undefined,
```
t0yv0, line 1883: `pulumi.Input<string> | pulumi.Output<string> | undefined` I *think* defines the same type space as ` `pulumi.Input<string> | undefined` because `pulumi.Input<string>` expands to `pulumi.Output<string> | Promise<string> | string`. Perhaps this is not needed?

nodejs/eks/nodegroup.ts
```diff
@@ -755,35 +755,40 @@ ${customUserData}
...
-    const numeric = new RegExp("^d+$");
+    const numeric = new RegExp("^\\d+$");
 
-    if (args.nodeRootVolumeIops && args.nodeRootVolumeType !== "io1") {
-        throw new Error(
-            "Cannot create a cluster node root volume of non-io1 type with provisioned IOPS (nodeRootVolumeIops).",
-        );
-    }
+    // We need to wrap the validation in a pulumi.all as MLCs could supply pulumi.Output<T> or T.
+    pulumi
```
t0yv0, line 761: I'm a little worried about this paradigm since the result of the pulumi.all(xs).apply(f) is not awaited anywhere but presumably it's awaited by the SDK.

IN addition, it changes the flow of the program to not fail fast but move forward and fall through all these checks even if some of them are failing. It will create error races, I think, which validator fails first. 

Looks like this approach is present in the codebase before the change so perhaps OK.

```typescript

    core.apply((c) => {
        if (c.nodeGroupOptions.nodeSecurityGroup && args.nodeSecurityGroup) {
            if (
                c.nodeSecurityGroupTags &&
                c.nodeGroupOptions.nodeSecurityGroup.id !== args.nodeSecurityGroup.id
            ) {
                throw new Error(
                    `The NodeGroup's nodeSecurityGroup and the cluster option nodeSecurityGroupTags are mutually exclusive. Choose a single approach`,
                );
            }
        }
    });
```

Have you tested the negative path, perhaps manually, where you trigger this error?

nodejs/eks/nodegroup.ts
```diff
@@ -755,35 +755,40 @@ ${customUserData}
         nodeAssociatePublicIpAddress = args.nodeAssociatePublicIpAddress;
     }
 
-    const numeric = new RegExp("^d+$");
+    const numeric = new RegExp("^\\d+$");
```
t0yv0, line 758: Could use a separate PR as slightly unrelated, but LGTM.

t0yv0:
Do we understand why the tests are failing?

nodejs/eks/nodegroup.ts
```diff
@@ -1867,24 +1880,26 @@ export function isGravitonInstance(instanceType: string): boolean {
...
-    instanceType: pulumi.Input<string> | undefined,
-): string {
-    if (amiType) {
-        // Return the user-specified AMI type.
-        return amiType.toString();
-    }
+    amiType: pulumi.Input<string> | pulumi.Output<string> | undefined,
+    gpu: pulumi.Input<boolean> | pulumi.Output<boolean> | undefined,
+    instanceType: pulumi.Input<string> | pulumi.Output<string> | undefined,
+): pulumi.Output<string> {
```
flostadler, line 1886: This will cause failures because `amiType` is used as a literal string in line 1840

flostadler (changes requested):

nodejs/eks/nodegroup.ts
t0yv0: Oof. Are there some linter settings that could've caught that or that's just on reviewers?

nodejs/eks/nodegroup.ts
rquitales: Yes, that is correct! Removed.

nodejs/eks/nodegroup.ts
rquitales: Good catch! I'm not sure if a linter would be able to catch this since we're formatting an object in a string literal. The existing test cases do catch this error though, hence the failing test.

nodejs/eks/nodegroup.ts
rquitales: Agreed. I've included it in this PR since it's a small change and don't want to go through a long CI process. I plan to rebase and merge so the commits can be rolled back easily should we need to.

nodejs/eks/nodegroup.ts
rquitales: Yes, the process in which we check these args may arise to some interesting edge cases, though we don't seem to have any reports of this currently. I'm not inclined to do a refactor within this PR since the scope is solely to just ensure we can handle pulumi.outputs being presented in the args and to resolve a P1.

I have done a quick manual validation of triggering the `The NodeGroup's nodeSecurityGroup and the cluster option nodeSecurityGroupTags are mutually exclusive. Choose a single approach` error and can hit it successfully during preview on a new stack. The concern you share regarding fail fast seems like it'd only occur when we are validating between a true Output that is unknown during preview/check. In that case, it wouldn't fail early. Since we're only validating against resource inputs I don't think it'd be a concern for not failing fast.

flostadler (approved):
LGTM! I don't think there's an easy fix for the more fundamental challenges  around running verifications async here. Let's re-evaluate the plain-ness of certain parameters for the next major versions

t0yv0:
>  I don't think there's an easy fix

There is a fix but it would make the code read a little ugly. 

```typescript
const os = new OutputSequencer()

os.sequence(output1);
os.sequence(output2);

```

I think based on the latest comment we're fine here though. Thank you!

Issue #1174, opened 2024-05-30
Upgrade tests are broken after release v2.5.0


flostadler:
### Problem description

The upgrade tests are targeting baseline version 2.3.0, but the tests use features from release 2.5.0. This causes the upgrade tests to fail.

### Errors & Logs

```
index.ts(36,3): error TS2345: Argument of type '{ cluster: eks.Cluster; nodeRole: Role; kubeletExtraArgs: string; enableIMDSv2: boolean; }' is not assignable to parameter of type 'ManagedNodeGroupOptions'.
    Object literal may only specify known properties, and 'kubeletExtraArgs' does not exist in type 'ManagedNodeGroupOptions'.
```
### Affected product version(s)

2.5.0+

### Reproducing the issue

Execute upgrade tests: `make test_nodejs_upgrade`

### Suggestions for a fix

Upgrade baseline version and re-record upgrade tests

Issue #1155, opened 2024-05-17
Upgrade tests has highly failure rate


rquitales:
### What happened?

The upgrade tests run during presubmits has a high error rate recently. The error presented is: `no space left on device`. On some re-runs of the job, it doesn't encounter this error. We already have a step right before testing to clear up disk space in CI. This frees up ~21 gb of space, so not sure what is using up all these resources.

Our current test code runs all the tests in parallel. We might have to reduce parallelism as a workaround for now until we can determine the root cause of this issue.

### Example

Example failing test: https://github.com/pulumi/pulumi-eks/actions/runs/9120670889/job/25114881783

Log output:
```
[resource plugin aws-6.36.0] installing
    pulumi:pulumi:Stack: (same)
      [urn=urn:pulumi:p-it-fv-az735-3-cluster-a3157e22::example-cluster::pulumi:pulumi:Stack::example-cluster-p-it-fv-az735-3-cluster-a3157e22]
  @ previewing update.....
  error: Running program '/tmp/p-it-fv-az735-3-cluster-a3157e22-1279121077/' failed with an unhandled exception:
  <ref *1> Error: failed to register new resource example-cluster-2 [eks:index:Cluster]: 2 UNKNOWN: Failed to encode GRPC debug logs: write /tmp/TestExamplesUpgradesclusterupgrade-preview-only419632818/001/preview-grpc-logs.json: no space left on device
      at Object.registerResource (/home/runner/work/pulumi-eks/pulumi-eks/nodejs/eks/node_modules/@pulumi/runtime/resource.ts:421:27)
      at new Resource (/home/runner/work/pulumi-eks/pulumi-eks/nodejs/eks/node_modules/@pulumi/resource.ts:507:13)
      at new ComponentResource (/home/runner/work/pulumi-eks/pulumi-eks/nodejs/eks/node_modules/@pulumi/resource.ts:1011:9)
      at new Cluster (/home/runner/work/pulumi-eks/pulumi-eks/nodejs/eks/cluster.ts:1671:9)
      at Object.<anonymous> (/tmp/p-it-fv-az735-3-cluster-a3157e22-1279121077/index.ts:14:18)
      at Module._compile (node:internal/modules/cjs/loader:1364:14)
      at Module.m._compile (/tmp/p-it-fv-az735-3-cluster-a3157e22-1279121077/node_modules/@pulumi/pulumi/vendor/ts-node@7.0.1/index.js:3009:23)
[...]

PR #1177, opened 2024-05-31
Use pulumitest framework for upgrade tests

rquitales:
### Proposed changes

This PR switches the underlying test framework to the newer pulumitest framework which uses Automation API under the hood. This resolves a few pain points related to the existing upgrade tests we have, which are extremely fragile. These changes are:

- Checking only for replacements, and ignoring diffs. This should help resolve our snapshots constantly being out of date due to changing upstream provider versions, or AMI IDs.
- Running preview against the cached state to reduce resource usage and speed up the upgrade tests

### Related issues (optional)
Related: #927

Fixes: #1145
Fixes: #1174
Fixes: #1155


examples/subnet-tags/index.ts
```diff
@@ -18,6 +18,7 @@ const vpc = new awsx.ec2.Vpc("vpc",
             {type: SubnetType.Public, tags: {"kubernetes.io/role/elb": "1", ...tags}},
             {type: SubnetType.Private, tags: {"kubernetes.io/role/internal-elb": "1", ...tags}},
         ],
+        subnetStrategy: "Legacy", // Upstream awsx default is being switched to "Auto", explicitly set to "Legacy".
```
flostadler, line 21: Should we maybe switch this to auto? Having legacy in the examples might be a bit discouraging

flostadler (approved):

examples/subnet-tags/index.ts
rquitales: Good idea! I was just throwing up something quick in the draft PR to get this working.

Issue #1181, opened 2024-06-04
Upgrade tests fails whenever there is a new recommended AMI ID


rquitales:
### What happened?

Our current examples uses the recommended AMI ID from AWS. However, whenever this value changes, our Pulumi programs within the context of an upgrade test will also cause a replacement update. This will fail our upgrade tests, so we should explicitly set the AMI ID for our clusters/nodes to avoid this issue.

### Example

https://github.com/pulumi/pulumi-eks/actions/runs/9370754621/job/25798678271?pr=1173

```
# trauncated log output:
           ~  aws:cloudformation:Stack example-ng-tags-spot-nodes update [diff: ~templateBody]
          @ previewing update.............
           -- aws:ec2:LaunchConfiguration example-ng-tags-ondemand-nodeLaunchConfiguration delete original [diff: ~imageId]
           -- aws:ec2:LaunchConfiguration example-tags-cluster1-nodeLaunchConfiguration delete original [diff: ~imageId]
           -- aws:ec2:LaunchConfiguration example-ng-tags-spot-nodeLaunchConfiguration delete original [diff: ~imageId]
              pulumi:pulumi:Stack example-tags-test  
          Resources:
              ~ 5 to update
              +-3 to replace
              8 changes. 64 unchanged
```

### Output of `pulumi about`

-

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #1182, opened 2024-06-04
tests: set amiID for upgrade tests

rquitales:
### Proposed changes

Explicitly set the AMI ID for our nodes and clusters to prevent constantly needing to update our test snapshots. The current AMI set in this PR is obtained from what we currently store in our test snapshots (example: https://github.com/pulumi/pulumi-eks/blob/b19f15552956f80a45493f887c3bf32e6cea08d0/provider/testdata/recorded/TestProviderUpgrade/tags/2.5.2/stack.json#L2051).

Explicitly setting the AMI ID is a much lower engineering lift than attempting to parse the automation API's change summary since it only returns an operation type/enum, and doesn't include what field is to be replaced. Furthermore, blanket ignoring a replacement of AMI IDs may inadvertently mask some replacement bugs we may have in the future. Note: we also have existing integration tests to also test that the provider can obtain and use a recommended ID., so we are not losing coverage of this logic flow.

### Related issues (optional)
Fixes: #1181

### DIFF
--- a/examples/cluster/index.ts
+++ b/examples/cluster/index.ts
@@ -35,8 +35,8 @@ const cluster2 = new eks.Cluster(`${projectName}-2`, {
 const cluster3 = new eks.Cluster(`${projectName}-3`, {
     vpcId: vpc.vpcId,
     publicSubnetIds: vpc.publicSubnetIds,
-    nodeAmiId: "ami-0384725f0d30527c7",
     nodeGroupOptions: {
+        amiId: "ami-0384725f0d30527c7",
         desiredCapacity: 1,
         minSize: 1,
         maxSize: 1,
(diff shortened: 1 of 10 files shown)


examples/cluster/index.ts
```diff
@@ -5,7 +5,9 @@ import * as eks from "@pulumi/eks";
 const projectName = pulumi.getProject();
 
 // Create an EKS cluster with the default configuration.
-const cluster1 = new eks.Cluster(`${projectName}-1`);
+const cluster1 = new eks.Cluster(`${projectName}-1` {
```
flostadler, line 8: This is missing a ‘,’

flostadler (approved):

Issue #895, opened 2023-07-24
Documentation appears to be missing supported inputs


MitchellGerdisch:
### What happened?

Docs page, for say, Nodegroup:
https://www.pulumi.com/registry/packages/eks/api-docs/managednodegroup/#inputs
does not show any setting related to, say, node group storage class (e.g. "gp2" or "gp3", etc).

But digging around the source code, one finds evidence of these inputs:
https://github.com/pulumi/pulumi-eks/blob/72e6429bb5494e80d62228de1d3f489593b19ad6/provider/cmd/pulumi-gen-eks/main.go#L442
https://github.com/pulumi/pulumi-eks/blob/72e6429bb5494e80d62228de1d3f489593b19ad6/nodejs/eks/nodegroup.ts#L148C1-L148C77

So, it seems like not everything is being presented in the docs and/or being exposed by the package.

### Expected Behavior

Docs should present supported inputs.

### Steps to reproduce

See notes above.

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
Thanks for reporting the lapse in our docs! @mnlumi I believe you've seen similar issues with incomplete doc generation. Do you have nay ideas what might be happening here. Iirc we had issues with nested fields in the past.

mnlumi:
@rquitales I'm not sure the exact cause, but there is this comment in [another pulumi-eks ticket related to nested properties MLC](https://github.com/pulumi/pulumi-eks/issues/626#issuecomment-1199446543): 
> This is a general bug from an MLC AFAICT that needs to be fixed as this also presents itself on AWSX.

There are also pages outside of pulumi-eks with missing content or blank fields like:
- https://www.pulumi.com/registry/packages/aws/api-docs/ec2/getami/ [Related issue](https://github.com/pulumi/pulumi-aws/issues/2437)
- https://www.pulumi.com/registry/packages/gcp/api-docs/serviceaccount/getaccount/#result
- https://www.pulumi.com/registry/packages/gcp/api-docs/iam/getworkloadidentitypool/
- https://www.pulumi.com/registry/packages/gcp/api-docs/artifactregistry/repositoryiambinding/
- https://www.pulumi.com/registry/packages/azapi/api-docs/getresourceaction/
- https://www.pulumi.com/registry/packages/azapi/api-docs/provider/

^to list a few.

[...]

Issue #784, opened 2022-10-04
ClusterNodeGroupOptions should support all node_root_volume_* options


betamike:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

`Cluster` supports configuring the root volume of nodes via a set of `node_root_volume_*` parameters in its initializer.  

The node group can also be configured by passing the `node_group_options` parameter. `ClusterNodeGroupOptions` only supports one (`node_root_volume_size`) and not the other five options.

Providing `node_group_options` is mutually exclusive with setting node group options on `Cluster`, making it impossible to use these options if `node_group_options` is already being used. This can be problematic if you need to set one of the parameters on `ClusterNodeGroupOptions` that don't appear on `Cluster`; `auto_scaling_group_tags` for example.

Overall it might be useful to understand why some of the options are duplicated; perhaps ease of use? It probably makes sense to commit to either keeping both sites in sync, or deprecate setting the node options on `Cluster` in favor of `node_group_options`.

markfickett:
> It probably makes sense to [...] deprecate setting the node options on Cluster in favor of node_group_options.
I got tripped up by this duplication, and would find it clearer to only have the `node_group_options`. I'd rather it be a little less convenient but have less opportunity for errors about mutually exclusive options or having multiple ways to do the same thing.

Issue #677, opened 2022-04-07
Cluster(.node_root_volume_type) not passed launch configuration


tma-unwire:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

When creating a new cluster using eks.Cluster(..., node_root_volume_type='gp3', ...) the volume type is not passed to the corresponding AWS Launch configuration.

### Steps to reproduce
Expected: use gp3 as volume type in Launch Configuration
Actual: uses gp2

stack72:
Related to #750

markfickett:
Looks somewhat related to #784 too.

rquitales:
We expect this to be supported once #1183 lands. Note though that we will not have `node_root_volume_type` as a field within the cluster resource. Instead, this can be set within the `cluster.node_group_options.node_root_volume_type` field.

PR #1183, opened 2024-06-04
fix: expose nodeRootVolume fields in NodeGroup and NodeGroupV2 resources

rquitales:
### Proposed changes

Update schema to expose the nodeRootVolume fields that is already present in Typescript.

- Exposed existing fields to schema and regenerated SDKs
- Added a new nodegroup in Python to test `gp3` volume type
- Manually inspecting created cluster from Python test to verify gp3 is created

### Related issues (optional)

Fixes: #895
Fixes: #677
Fixes: #784
Fixes: #718

### DIFF
--- a/examples/nodegroup-py/__main__.py
+++ b/examples/nodegroup-py/__main__.py
@@ -152,5 +152,20 @@ eks.NodeGroup(
     opts=pulumi.ResourceOptions(providers={"kubernetes": cluster2_provider}),
 )
 
+eks.NodeGroup(
+    "example-ng-advanced-spot-gp3-py",
+    cluster=cluster2,
+    instance_type="t3.medium",
+    desired_capacity=1,
+    min_size=1,
+    max_size=2,
+    node_root_volume_type="gp3",
+    spot_price="1",
+    labels={"preemptible": "true"},
+    taints={"special": eks.TaintArgs(value="true", effect="NoSchedule")},
+    instance_profile=instance_profile3,
+    opts=pulumi.ResourceOptions(providers={"kubernetes": cluster2_provider}),
+)
+
 # Export the cluster's kubeconfig.
 pulumi.export("kubeconfig2", cluster2.kubeconfig)
--- a/provider/cmd/pulumi-gen-eks/main.go
+++ b/provider/cmd/pulumi-gen-eks/main.go
@@ -658,7 +658,7 @@ func generateSchema() schema.PackageSpec {
 					},
 					"authenticationMode": {
 						TypeSpec: schema.TypeSpec{
-							Ref: "#/types/eks:index:AuthenticationMode",
+							Ref:   "#/types/eks:index:AuthenticationMode",
 							Plain: true,
 						},
 						Description: "The authentication mode of the cluster. Valid values are `CONFIG_MAP`, `API` or `API_AND_CONFIG_MAP`.\n\n" +
@@ -1587,9 +1587,9 @@ func generateSchema() schema.PackageSpec {
 			},
 			"eks:index:AccessEntryType": {
 				ObjectTypeSpec: schema.ObjectTypeSpec{
-					Type:        "string",
+					Type: "string",
 					Description: "The type of the new access entry. Valid values are STANDARD, FARGATE_LINUX, EC2_LINUX, and EC2_WINDOWS.\n" +
-					"Defaults to STANDARD which provides the standard workflow. EC2_LINUX and EC2_WINDOWS types disallow users to input a kubernetesGroup, and prevent associating access policies.",
+						"Defaults to STANDARD which provides the standard workflow. EC2_LINUX and EC2_WINDOWS types disallow users to input a kubernetesGroup, and prevent associating access policies.",
 				},
 				Enum: []schema.EnumValueSpec{
 					{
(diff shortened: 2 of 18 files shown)


nodejs/eks/storageclass.ts
```diff
@@ -20,7 +20,7 @@ import * as pulumi from "@pulumi/pulumi";
 /**
  * EBSVolumeType lists the set of volume types accepted by an EKS storage class.
  */
-export type EBSVolumeType = "io1" | "gp2" | "sc1" | "st1";
+export type EBSVolumeType = "io1" | "gp2" | "gp3" | "sc1" | "st1";
```
flostadler, line 23: This is a tricky one. This doesn't refer to the node root volumes, but the storage classes for EKS.

The default in-tree plugin doesn't support gp3.
AWS moved to having a separate EBS CSI driver that users need to install in order to use a gp3 storage class.

I'd keep this separate for now and tackle it if we want to tackle https://github.com/pulumi/pulumi-eks/issues/833

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1692,6 +1692,26 @@ func nodeGroupProperties(cluster, v2 bool) map[string]schema.PropertySpec {
 			TypeSpec:    schema.TypeSpec{Type: "integer"},
 			Description: "The size in GiB of a cluster node's root volume. Defaults to 20.",
 		},
+		"nodeRootVolumeDeleteOnTermination": {
```
flostadler, line 1695: This is a bit meta, but I'm wondering what we can do to prevent something like this from happening again in the future. We've already increased the test coverage in the last weeks, but I feel like this is gonna happen again unless we codegen the node SDK as well

t0yv0: https://github.com/pulumi/pulumi-eks/issues/1184 I didn't find a tracking issue so added one. This doesn't sound easy but sounds like something we definitely need to do.

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1534,9 +1534,9 @@ func generateSchema() schema.PackageSpec {
 			},
 			"eks:index:AccessEntryType": {
 				ObjectTypeSpec: schema.ObjectTypeSpec{
-					Type:        "string",
```
t0yv0, line 1537: I suspect there is no lint in this repo? Sadness. Do we track installing it already?

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1692,6 +1692,26 @@ func nodeGroupProperties(cluster, v2 bool) map[string]schema.PropertySpec {
...
+		"nodeRootVolumeEncrypted": {
+			TypeSpec:    schema.TypeSpec{Type: "boolean"},
+			Description: "Whether to encrypt a cluster node's root volume. Defaults to false.",
+		},
+		"nodeRootVolumeIops": {
+			TypeSpec:    schema.TypeSpec{Type: "integer"},
+			Description: "The amount of provisioned IOPS. This is only valid with a volumeType of 'io1'.",
+		},
+		"nodeRootVolumeThroughput": {
+			TypeSpec:    schema.TypeSpec{Type: "integer"},
```
t0yv0, line 1708: No floats here?

provider/cmd/pulumi-resource-eks/schema.json
```diff
@@ -255,10 +255,30 @@
...
+                "nodeRootVolumeIops": {
+                    "type": "integer",
+                    "description": "The amount of provisioned IOPS. This is only valid with a volumeType of 'io1'."
+                },
                 "nodeRootVolumeSize": {
                     "type": "integer",
                     "description": "The size in GiB of a cluster node's root volume. Defaults to 20."
                 },
+                "nodeRootVolumeThroughput": {
+                    "type": "integer",
```
t0yv0, line 275: Ditto, only integers?

examples/nodegroup-py/__main__.py
```diff
@@ -152,5 +152,20 @@
     opts=pulumi.ResourceOptions(providers={"kubernetes": cluster2_provider}),
 )
 
+eks.NodeGroup(
```
t0yv0, line 155: Thank you! Do we need to assert anything here or the example will just fail if this doesn't provision?

t0yv0 (approved):

provider/cmd/pulumi-gen-eks/main.go
flostadler: Ah you were quicker than me! Thanks for creating it

nodejs/eks/storageclass.ts
rquitales: Thanks for the extra info! I've dropped the commit for this change.

provider/cmd/pulumi-gen-eks/main.go
rquitales: Let me create a tracking issue and a follow-up PR for this!

examples/nodegroup-py/__main__.py
rquitales: I'll see if there's an easy way to also add some assertion for sanity checking. Right now, it's just manual validation.

provider/cmd/pulumi-gen-eks/main.go
rquitales: Yes, this is an integer only value. Ref: https://docs.aws.amazon.com/cli/latest/reference/emr/create-cluster.html

provider/cmd/pulumi-gen-eks/main.go
rquitales: #1185

flostadler (approved):

Issue #1205, opened 2024-06-19
Build broken due to ClusterInternal import in addon.ts


chasinglogic:
### What happened?

After upgrading to version 2.7.0 and 2.7.1 our typescript compiles are now broken for this package:

```
../../node_modules/@pulumi/eks/addon.d.ts:3:19 - error TS2305: Module '"./cluster"' has no exported member 'ClusterInternal'.

3 import { Cluster, ClusterInternal } from "./cluster";
                    ~~~~~~~~~~~~~~~


Found 1 error.

npm ERR! Lifecycle script `build` failed with error:
npm ERR! Error: command failed
npm ERR!   in workspace: aws-lbc
npm ERR!   at location: /Users/mathewrobinson/Work/platform/tillo-iac/infrastructure/aws-lbc
```

### Example

Any clean build for any of our packages which rely on version 2.7.0 have this error. So simply compile against it and you should see the same.

It looks like `ClusterInternal` is exported in the `cluster.ts` file but in the resulting `cluster.d.ts` file the symbol is entirely absent:

```
~/Work/platform/tillo-iac/node_modules/@pulumi/eks task/debug-node-group λ grep ClusterInternal addon.d.ts
import { Cluster, ClusterInternal } from "./cluster";
    cluster: Cluster | ClusterInternal;
[kube: production]
~/Work/platform/tillo-iac/node_modules/@pulumi/eks task/debug-node-group λ grep ClusterInternal cluster.d.ts
[kube: production]
!! ~/Work/platform/tillo-iac/node_modules/@pulumi/eks task/debug-node-group λ
```

### Output of `pulumi about`

CLI          
Version      3.120.0
Go Version   go1.22.4
Go Compiler  gc

Plugins
KIND      NAME    VERSION
language  nodejs  unknown

Host     
OS       darwin
Version  14.4.1
Arch     arm64

This project is written in nodejs: executable='/Users/mathewrobinson/.local/share/mise/installs/node/lts/bin/node' version='v20.12.2'

Current Stack: organization/eks/dev

TYPE                                                 URN
pulumi:pulumi:Stack                                  urn:pulumi:dev::eks::pulumi:pulumi:Stack::eks-dev
pulumi:providers:aws                                 urn:pulumi:dev::eks::pulumi:providers:aws::subaccount-provider
pulumi:providers:aws                                 urn:pulumi:dev::eks::pulumi:providers:aws::internal-services-provider
[...]

PR #1209, opened 2024-06-21
Remove ClusterInternal reference in typescript provider

rquitales:
### Proposed changes

The ClusterInternal type is a internal TS type. When users attempt to transpile their typescript programs to javascript, they'll encounter a symbol not found error, since internal types won't be exposed in the type annotations file. `ClusterInternal` isn't necessary since its used for MLC codegen, so this will not be a breaking change.

#### Testing:

- Updated existing typescript test to also run the build step

### Related issues (optional)

Fixes: #1205

### DIFF
--- a/examples/cluster/index.ts
+++ b/examples/cluster/index.ts
@@ -84,10 +84,10 @@ const coredns = new eks.Addon("coredns", {
 });
 
 // Export the clusters' kubeconfig.
-export const kubeconfig1 = cluster1.kubeconfig;
-export const kubeconfig2 = cluster2.kubeconfig;
-export const kubeconfig3 = cluster3.kubeconfig;
-export const kubeconfig4 = cluster4.kubeconfig;
+export const kubeconfig1: pulumi.Output<any> = cluster1.kubeconfig;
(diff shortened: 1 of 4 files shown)


examples/cluster/index.ts
```diff
@@ -84,10 +84,10 @@ const coredns = new eks.Addon("coredns", {
 });
 
 // Export the clusters' kubeconfig.
-export const kubeconfig1 = cluster1.kubeconfig;
-export const kubeconfig2 = cluster2.kubeconfig;
-export const kubeconfig3 = cluster3.kubeconfig;
-export const kubeconfig4 = cluster4.kubeconfig;
+export const kubeconfig1: pulumi.Output<any> = cluster1.kubeconfig;
```
rquitales, line 87: This is necessary due to us linking the eks sdk for testing purposes. Without the type assertions, we'd run into `tsc` errors like:

```
index.ts:93:14 - error TS2742: The inferred type of 'kubeconfig1' cannot be named without a reference to '../../nodejs/eks/node_modules/@pulumi/pulumi'. This is likely not portable. A type annotation is necessary.
```

blampe (approved):

Issue #1202, opened 2024-06-17
 TypeError: Cannot read properties of undefined (reading 'map') - eks cluster with 2 managed node groups


jamesongithub:
### What happened?

trying to create eks cluster with 2 managed node groups

### Example

```python
eks_vpc = awsx.ec2.Vpc("eks-vpc",
    enable_dns_support=True,
    enable_dns_hostnames=True,
    cidr_block="192.168.0.0/16",
    subnet_strategy=awsx.ec2.SubnetAllocationStrategy.AUTO,
)

eks_cluster = eks.Cluster("eks-cluster",
    # Put the cluster in the new VPC created earlier
    vpc_id=eks_vpc.vpc_id,
    # Public subnets will be used for load balancers
    public_subnet_ids=eks_vpc.public_subnet_ids,
    # Private subnets will be used for cluster nodes
    private_subnet_ids=eks_vpc.private_subnet_ids,
    # Change configuration values to change any of the following settings

    # Do not give worker nodes a public IP address
    node_associate_public_ip_address=False,
    # Change these values for a private cluster (VPN access required)
    endpoint_private_access=False,
    endpoint_public_access=True,
    skip_default_node_group=True,
)

node_role = aws.iam.Role("node-role", 
    assume_role_policy="""{
        "Version": "2012-10-17",
        "Statement": [
            {
                "Effect": "Allow",
                "Principal": { "Service": "ec2.amazonaws.com" },
                "Action": "sts:AssumeRole"
            }
        ]
    }"""
)

node_role_policy_attachment_eks_worker = aws.iam.RolePolicyAttachment("node-role-policy-eks-worker",
    role=node_role.name,
    policy_arn="arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy"
)

node_role_policy_attachment_ecr = aws.iam.RolePolicyAttachment("node-role-policy-ecr",
    role=node_role.name,
    policy_arn="arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly"
)

node_role_policy_attachment_eks_cni = aws.iam.RolePolicyAttachment("node-role-policy-eks-cni",
    role=node_role.name,
    policy_arn="arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy"
)

# Create a managed node group with labels
managed_node_group_infra = eks.ManagedNodeGroup("infra",
    cluster=eks_cluster.core,
    node_group_name="infra-nodes",
[...]

PR #1214, opened 2024-06-26
Fix undefined panic when using `.map`

rquitales:
### Proposed changes

Adds an `undefined` guard before attempting to use `.map()`. This is due to MLC programs being able to send `undefined` instead of an empty array.

This PR also updates the error handling to throw `pulumi.ResourceError` instead of a generic nodejs error.

### Related issues (optional)

Fixes: #1202

### DIFF
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -1195,7 +1195,8 @@ ${customUserData}
             if (nodeRootVolumeType === "gp3" && nodeRootVolumeThroughput) {
                 if (!numeric.test(nodeRootVolumeThroughput?.toString())) {
                     throw new pulumi.ResourceError(
-                        "Cannot create a cluster node root volume of gp3 type without provisioned throughput (nodeRootVolumeThroughput) as integer value.", parent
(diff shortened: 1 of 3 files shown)


nodejs/eks/nodegroup.ts
```diff
@@ -1652,16 +1674,16 @@ function createManagedNodeGroupInternal(
...
     // Check that the nodegroup role has been set on the cluster to
     // ensure that the aws-auth configmap was properly formed.
     const nodegroupRole = pulumi.all([core.instanceRoles, roleArn]).apply(([roles, rArn]) => {
-        // Map out the ARNs of all of the instanceRoles.
-        const roleArns = roles.map((role) => {
-            return role.arn;
-        });
+        // Map out the ARNs of all of the instanceRoles. Note that the roles array may be undefined if
+        // unspecified by the Pulumi program.
+        const roleArns: pulumi.Output<string>[] = roles ? roles.map((role) => role.arn) : [];
```
rquitales, line 1685: This is the actual fix to always output an empty array.

blampe (approved):

Issue #1195, opened 2024-06-13
Support AL2023 & Bottlerocket for node groups


flostadler:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

When creating a `ManagedNodeGroup`/`NodeGroup` with `kubeletExtraArgs` or `bootstrapExtraArgs` the component creates a `LaunchTemplate`/`LaunchConfig` with custom EC2 userdata to bootstrap the node.
Right now this `LaunchTemplate`/`LaunchConfig` is hardcoded to use AL2. AL2 will be EOL in June 2025 and users need enough time to upgrade to newer AMIs (Bottlerocket or AL2023).

This change will involve changes to the way userdata is structured because Amazon Linux 2023 (AL2023) introduces a new node initialization process nodeadm that uses a YAML configuration schema and Bottlerocket structures user data in the TOML format (see [AWS docs](https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html#launch-template-user-data)).

Necessary changes:
- [x] Remove AL2 hardcoding to start supporting AL2023 and Bottlerocket
    - [x] PR: https://github.com/pulumi/pulumi-eks/pull/1328
- [x] Support crafting custom user data
    - [x] AL2023
        - [x] PR: https://github.com/pulumi/pulumi-eks/pull/1332
    - [x] Bottlerocket
        - [x] PR: https://github.com/pulumi/pulumi-eks/pull/1334

Necessary bug fixes:
- [x] Fix self managed node groups wrongly assuming that nodes only have one block device
    - [x] PR: https://github.com/pulumi/pulumi-eks/pull/1337
- [x] ManagedNodeGroup with custom launch template always use x86 AMI even when graviton instances are used: #1323
- [x] eks.ManagedNodeGroup with amiType: 'AL2023_ARM_64_STANDARD' breaks when kubeletExtraArgs specified #1179
    - [x] PR: https://github.com/pulumi/pulumi-eks/pull/1328

### Affected area/feature

- `pulumi-eks`

flostadler:
#1323 is fixed now.
Continuing with this work now.

In order for ami type selection to work we'll also need to fix https://github.com/pulumi/pulumi-eks/issues/1179.

PR #1225, opened 2024-07-03
Managed Node Group Launch Template Improvements

### LINKED ISSUE #1224: Improve ManagedNodeGroup when creating a LaunchTemplate
## Hello!

## Issue details

In my current use case, I'm trying to implement a Managed Node Group with GPU support. Out of the box this isn't possible, however with a few small tweaks it should be entirely possible.

In addition, during my investigation I noticed a bug where it's not possible to specify the disk size of a managed node group when a custom launch template is created. This is because AWS requires the disk configuration to happen on the launch template, not the `diskSize` argument in this case.

My suggested implementation is in the PR #1225 

### Affected area/feature

`ManagedNodeGroup`

JustASquid:
### Proposed changes

Add the ability to specify `gpu`/`amiId`/`amiType` on a managed node group, and specify disk size when a launch template is created.

This PR is a draft because I haven't been able to test the changes (yet) - any guidance on this is helpful, I don't have easy access to a spare AWS account where I can spin up test clusters.

### Related issues (optional)

Resolves #1224

### DIFF
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -1544,6 +1544,43 @@ export type ManagedNodeGroupOptions = Omit<
      *   - maxSize: 2
      */
     scalingConfig?: pulumi.Input<awsInputs.eks.NodeGroupScalingConfig>;
+
+    /**
+     * The AMI type for the instance.
+     *
+     * If you are passing an amiId that is `arm64` type, then we need to ensure
+     * that this value is set as `amazon-linux-2-arm64`.
+     *
+     * Note: `amiType` and `gpu` are mutually exclusive.
+     */
+    amiType?: pulumi.Input<string>;
+
+    /**
+     * The AMI ID to use for the worker nodes.
+     *
+     * Defaults to the latest recommended EKS Optimized Linux AMI from the
+     * AWS Systems Manager Parameter Store.
+     *
+     * Note: `amiId` and `gpu` are mutually exclusive.
+     *
+     * See for more details:
+     * - https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html.
+     */
+    amiId?: pulumi.Input<string>;
+
+    /**
+     * Use the latest recommended EKS Optimized Linux AMI with GPU support for
+     * the worker nodes from the AWS Systems Manager Parameter Store.
+     *
+     * Defaults to false.
+     *
+     * Note: `gpu` and `amiId` are mutually exclusive.
+     *
+     * See for more details:
+     * - https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html.
+     * - https://docs.aws.amazon.com/eks/latest/userguide/retrieve-ami-id.html
+     */
+    gpu?: pulumi.Input<boolean>;
 };
 
 /**
@@ -1668,6 +1705,14 @@ function createManagedNodeGroupInternal(
         );
     }
 
+    if (args.amiId && args.gpu) {
+        throw new pulumi.ResourceError("amiId and gpu are mutually exclusive.", parent);
+    }
+
+    if (args.amiType && args.gpu) {
+        throw new pulumi.ResourceError("amiType and gpu are mutually exclusive.", parent);
+    }
+
     let roleArn: pulumi.Input<string>;
     if (args.nodeRoleArn) {
         roleArn = args.nodeRoleArn;
@@ -1740,8 +1785,11 @@ function createManagedNodeGroupInternal(
     }
 
     let launchTemplate: aws.ec2.LaunchTemplate | undefined;
-    if (args.kubeletExtraArgs || args.bootstrapExtraArgs || args.enableIMDSv2) {
+    if (args.kubeletExtraArgs || args.bootstrapExtraArgs || args.enableIMDSv2 || args.gpu || args.amiId || args.amiType) {
         launchTemplate = createMNGCustomLaunchTemplate(name, args, core, parent, provider);
+
+        // Disk size is specified in the launch template.
+        delete nodeGroupArgs.diskSize;
     }
 
     // Make the aws-auth configmap a dependency of the node group.
@@ -1836,9 +1884,20 @@ Content-Type: text/x-shellscript; charset="us-ascii"
         ? { httpTokens: "required", httpPutResponseHopLimit: 2, httpEndpoint: "enabled" }
         : undefined;
 
+    const blockDeviceMappings = [
+        {
+            deviceName: "/dev/xvda",
+            ebs: {
+                volumeSize: args.diskSize,
+                volumeType: "gp2"
+            },
+        },
+    ];
+
     return new aws.ec2.LaunchTemplate(
         `${name}-launchTemplate`,
         {
+            blockDeviceMappings,
             userData,
             metadataOptions,
             // We need to always supply an imageId, otherwise AWS will attempt to merge the user data which will result in


nodejs/eks/nodegroup.ts
```diff
@@ -1544,6 +1544,43 @@ export type ManagedNodeGroupOptions = Omit<
...
+
+    /**
+     * The AMI type for the instance.
+     *
+     * If you are passing an amiId that is `arm64` type, then we need to ensure
+     * that this value is set as `amazon-linux-2-arm64`.
+     *
+     * Note: `amiType` and `gpu` are mutually exclusive.
+     */
+    amiType?: pulumi.Input<string>;
```
t0yv0, line 1556: These looks like something copied over from NodeGroupBaseOptions, I wonder if we could factor this out in TypeScript somehow.

nodejs/eks/nodegroup.ts
```diff
@@ -1740,8 +1785,11 @@ function createManagedNodeGroupInternal(
     }
 
     let launchTemplate: aws.ec2.LaunchTemplate | undefined;
-    if (args.kubeletExtraArgs || args.bootstrapExtraArgs || args.enableIMDSv2) {
+    if (args.kubeletExtraArgs || args.bootstrapExtraArgs || args.enableIMDSv2 || args.gpu || args.amiId || args.amiType) {
         launchTemplate = createMNGCustomLaunchTemplate(name, args, core, parent, provider);
+
+        // Disk size is specified in the launch template.
+        delete nodeGroupArgs.diskSize;
```
t0yv0, line 1792: A separate PR/issue for the disk size option would be much appreciated.

t0yv0:
Thanks for contributing here! We indeed rely heavily on integration-level tests so an AWS testing account is needed to test the changes.

nodejs/eks/nodegroup.ts
JustASquid: I had a look, this inherits directly from the AWS config so I figured we shouldn't mess with it too much.

JustASquid:
Yep @t0yv0 - Is testing something you're prepared to help with? I do have access to my company's AWS account, but spinning up cluster(s) is a pretty heavy operation so would love to avoid if possible.

flostadler:
@JustASquid I can help with testing this. Do you want to have a first stab at creating a test?

nodejs/eks/nodegroup.ts
```diff
@@ -1544,6 +1544,33 @@ export type ManagedNodeGroupOptions = Omit<
...
+     *
+     * Defaults to the latest recommended EKS Optimized Linux AMI from the
+     * AWS Systems Manager Parameter Store.
+     *
+     * Note: `amiId` and `gpu` are mutually exclusive.
+     *
+     * See for more details:
+     * - https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html.
+     */
+    amiId?: pulumi.Input<string>;
```
flostadler, line 1559: The SDK for Go, python, java and dotnet is generated based on the schema here: https://github.com/pulumi/pulumi-eks/blob/master/provider/cmd/pulumi-gen-eks/main.go#L718
You also need to add these changes there.

Only the typescript SDK is hand-written and not generated.

JustASquid:
Cheers @flostadler - I just went on leave, should be able to take a stab at writing some tests in a couple weeks. Appreciate the help and guidance.

JustASquid:
@flostadler - I added the SDK generation and took a crack at adding some tests - surely they are missing something but hopefully it's a jumping off point.

examples/tests/managed-ng-ami-options/gpu/index.ts
```diff
@@ -0,0 +1,35 @@
...
+});
+
+// Export the cluster's kubeconfig.
+export const kubeconfig = cluster.kubeconfig;
+
+// Create a managed node group using a cluster as input.
+eks.createManagedNodeGroup(`${projectName}-managed-ng`, {
+    cluster: cluster,
+    nodeRole: role,
+    gpu: true
```
flostadler, line 34: Does this have any special instance type requirements or can we launch a GPU optimized AMI on a basic t3 instance for example?

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -735,6 +735,21 @@ func generateSchema() schema.PackageSpec {
 							"(https://docs.aws.amazon.com/eks/latest/APIReference/API_Nodegroup.html#AmazonEKS-Type-Nodegroup-amiType) " +
 							"for valid AMI Types. This provider will only perform drift detection if a configuration value is provided.",
 					},
+					"amiId": {
```
flostadler, line 738: You'll need to run `make build` and check in the changes to the schema and `sdk` folder. Otherwise those don't take effect.

Because of the disparity between the nodejs SDK and the rest of the SDKs it would be good to add a test in another language as well (e.g. Golang). Let me know if you need help with converting the test you've created into another language

examples/tests/managed-ng-ami-options/ami-id/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    instanceRoles: [role],
+});
+
+// Export the cluster's kubeconfig.
+export const kubeconfig = cluster.kubeconfig;
+
+// Find the recommended AMI for the EKS node group.
+// See https://docs.aws.amazon.com/eks/latest/userguide/retrieve-ami-id.html
+async function getEksAmiId(k8sVersion: string): Promise<string> {
+    const client = new SSMClient();
```
flostadler, line 34: You can do that in pulumi natively. See here for an example:
https://github.com/pulumi/pulumi-eks/blob/9d80b5df0dc491f3dd09c159a0183e40e8350bba/examples/custom-managed-nodegroup/index.ts#L38-L40

examples/tests/managed-ng-ami-options/ami-id/index.ts
```diff
@@ -0,0 +1,53 @@
...
+        throw new Error(`Could not find EKS optimized AMI for Kubernetes version ${k8sVersion}`);
+    }
+
+    return response.Parameter.Value;
+}
+
+const amiId = cluster.eksCluster.version.apply(version => pulumi.output(getEksAmiId(version)));
+
+// Create a managed node group using a cluster as input.
+eks.createManagedNodeGroup(`${projectName}-managed-ng`, {
```
flostadler, line 49: What do you think about combining the ami and gpu tests into one?
One test with two nodegroups should be fine.

We need to balance the number of eks clusters we're creating and here we're interested  in the node group.

nodejs/eks/nodegroup.ts
```diff
@@ -1740,7 +1775,7 @@ function createManagedNodeGroupInternal(
     }
 
     let launchTemplate: aws.ec2.LaunchTemplate | undefined;
-    if (args.kubeletExtraArgs || args.bootstrapExtraArgs || args.enableIMDSv2) {
+    if (args.kubeletExtraArgs || args.bootstrapExtraArgs || args.enableIMDSv2 || args.gpu || args.amiId || args.amiType) {
```
flostadler, line 1778: Why do we need to create a custom launch template if `amiType` is set? That's a valid input for `aws.eks.NodeGroup` on its own.

Aside from that, the list here gets a bit long now. What do you think about creating a method for deciding whether to create a custom launch template?

flostadler:
Thanks a lot @JustASquid! I left some comments, let me know if you need some help with any of the proposed changes

flostadler:
@JustASquid I'm starting to touch the same parts of the code base as part of refactoring needed for https://github.com/pulumi/pulumi-eks/issues/1195. I'd propose that I fold your changes into that to prevent duplication/conflicts.
What do you think about that?

flostadler:
Closing this in favor of https://github.com/pulumi/pulumi-eks/issues/1195.

PR #1233, opened 2024-07-10
Ci tidy

danielrbradley:
Fix up ~250 actionlint failures and add CI job to perform check on all PRs.

Stacked on:
- https://github.com/pulumi/pulumi-eks/pull/1232


.node-version
```diff
@@ -0,0 +1 @@
+18.14.0
```
flostadler, line 1: What's the reason for using a fully qualified version now instead of `18.x`?

danielrbradley: A `.node-version` file should generally be an exact version so that you're guarenteed consistent results in all build environments.

We should probably set up dependabot to bump this for us though. Also, I'll bump this to the latest version of 18.

flostadler: Yeah I was mostly worried about us not upgrading to the latest versions. Dependabot sounds good!

flostadler (approved):

PR #1242, opened 2024-07-10
Combined dependency upgrades

danielrbradley:
Combine all dependabot PRs and fix build.

- Fix dotnetgen.GeneratePackage arguments
- Re-generate python SDK
- Downgrade node version by 1 minor step as latest seems to sometime error when fetching


flostadler (approved):
Nice!

PR #1292, opened 2024-08-08
Add ignore changes for bootstrapClusterCreatorAdminPermissions of EKS clusters

### LINKED ISSUE #1291: Changing `authenticaion_mode` inadvertantly triggering replacement
### What happened?

In recent pulumi-eks versions, there's a new parameter called 'authentication_mode', which is documented here: https://www.pulumi.com/registry/packages/eks/api-docs/cluster/#authentication_mode_python

We're using pulumi-eks as it's a nice abstraction for eks components. However, when setting the 'authentication_mode' to 'CONFIG_MAP' (which is what the cluster was already using), this triggers a cluster replacement.

As turns out, once you set 'authentication_mode' to anything other than 'None' - this modifies the 'bootstrapClusterCreatorAdminPermissions' attribute of the underlying aws.eks.Cluster resource.

I referred to this issue:
https://github.com/pulumi/pulumi-aws/issues/3997 which suggested a fix.

However, using the pulumi-eks package, there's no way to control the 'access_config' attribute as it belongs to an upstream resource.

I'd really like to use 'authentication_mode', but can't have my clusters break over it

### Example

The field that triggers the replacement is:
```
~ accessConfig: {
~ bootstrapClusterCreatorAdminPermissions: false => true
}
```

### Output of `pulumi about`

pulumi==3.128.0
pulumi_aws==6.48.0
pulumi_awsx==2.13.0
pulumi_docker==4.5.5
pulumi_eks==2.7.7
pulumi_kubernetes==4.15.0

### Additional context

_No response_

### Contributing

flostadler:
When an EKS cluster was created with a provider version before 2.7.4 it was affected by this upstream bug: https://github.com/pulumi/pulumi-aws/issues/3997.
This caused the state to not actually reflect the default setting for this parameter. In the cloud this was set to true and in the state to false.
In 2.7.4 we pulled in the upstream fix for this which introduced bi-modal default behavior. If accessConfig is not specified, `bootstrapClusterCreatorAdminPermissions` defaults to true.
If `accessConfig` is provided but `bootstrapClusterCreatorAdminPermissions` is not set, it defaults to false. This parameter cannot be retrieved from the AWS APIs so it can't be corrected by a refresh.
Here's some more context: https://github.com/pulumi/pulumi-aws/issues/3997#issuecomment-2223201333

In order to fix this, this change marks changes in `bootstrapClusterCreatorAdminPermissions` to be ignored.
This is a safe option because it is a create only parameter.

Fixes #1291


provider/provider_test.go
```diff
@@ -157,3 +161,127 @@ func testProviderUpgrade(t *testing.T, example string) {
 	)
 	assertpreview.HasNoReplacements(t, result)
 }
+
+type testProviderCodeChangesOptions struct {
```
flostadler, line 165: This was adapted from here: https://github.com/pulumi/pulumi-aws/blob/2a3dd788d66627edd57ed38c86a6f761c65eea64/provider/provider_test.go#L121

I wanna move this to https://github.com/pulumi/providertest but first wanted to get this test working

PR #1296, opened 2024-08-09
chore: IgnoreDestroyErrors is set for all tests in CI

### LINKED ISSUE #1222: Integration Test Flake: DependencyViolation NodeSecurityGroup has a dependent object
We started getting an elevated amount of flakes related to the node SecurityGroup failing to be deleted e.g. https://github.com/pulumi/pulumi-eks/issues/1219. 
```
    aws:ec2:SecurityGroup (example-cluster-py-3-nodeSecurityGroup):
      error: deleting urn:pulumi:p-it-fv-az1210--cluster-py-37ea6979::example-cluster-py::eks:index:Cluster$aws:ec2/securityGroup:SecurityGroup::example-cluster-py-3-nodeSecurityGroup: 1 error occurred:
      	* deleting Security Group (sg-0e8b6d9244b79c723): DependencyViolation: resource sg-0e8b6d9244b79c723 has a dependent object
      	status code: 400, request id: 1b401e09-4dda-4402-8a4e-95c43e03dbec
```

There might be missing dependency links between resources (e.g. nodes depending on the node SecurityGroup).

t0yv0:
### Proposed changes

To stabilize CI the IgnoreDestroyErrors option is now set on all the tests, not just the Python Cluster test. This can be removed once the underlying problem with dangling ENIs is fixed and `pulumi destroy` can be relied on for eks.Cluster resources.

### Related issues (optional)

Fixes #1222

### DIFF
--- a/examples/examples_dotnet_test.go
+++ b/examples/examples_dotnet_test.go
@@ -38,7 +38,7 @@ func TestAccClusterCs(t *testing.T) {
 			},
 		})
 
-	integration.ProgramTest(t, &test)
+	programTestWithExtraOptions(t, &test, nil)
 }
 
 func getCSBaseOptions(t *testing.T) integration.ProgramTestOptions {
--- a/examples/examples_go_test.go
+++ b/examples/examples_go_test.go
@@ -52,7 +52,7 @@ func TestAccClusterGo(t *testing.T) {
 			},
 		})
 
-	integration.ProgramTest(t, &test)
+	programTestWithExtraOptions(t, &test, nil)
 	// Ensure that the provider error message is as expected.
 	assert.Contains(t, stdErr.String(), "not supported")
 }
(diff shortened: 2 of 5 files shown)


flostadler (approved):
Just making sure, does the account cleanup script clean out ENIs? If not we'd rack up costs and eventually lock the accounts when we hit the quota (5000).

I'm thinking it does, because otherwise we would've already run into this.

flostadler:
> Just making sure, does the account cleanup script clean out ENIs? If not we'd rack up costs and eventually lock the accounts when we hit the quota (5000).
> 
> I'm thinking it does, because otherwise we would've already run into this.

It does handle ENIs: https://github.com/pulumi/aws-account-cleanup/blob/f0d66abdea041018e487f2b1bc3d21af5737b040/application/main.go#L2268

PR #1312, opened 2024-08-22
Update acceptance test CI workflows

### LINKED ISSUE #1311: Acceptance tests failing because Cluster limit is exceeded for CI account
Our acceptance tests are creating too many clusters. This causes us to hit the account limit of 100 clusters.
```
aws:eks:Cluster (example-managed-nodegroups-eksCluster):
      error: 1 error occurred:
      	* creating EKS Cluster (example-managed-nodegroups-eksCluster-e0ca36e): operation error EKS: CreateCluster, https response error StatusCode: 400, RequestID: 7241367f-072f-427b-943e-28726e896097, ResourceLimitExceededException: Cluster limit exceeded for account.
```

We should attack this from two angles, 1) send quota increase requests to AWS to up the limit 2) better control the concurrency of our acceptance tests.
Right now our acceptance tests start a big test matrix of 31 jobs for the acceptance tests. That matrix is manually maintained ([see](https://github.com/pulumi/pulumi-eks/blob/9d80b5df0dc491f3dd09c159a0183e40e8350bba/.github/workflows/run-acceptance-tests.yml#L428)) which also poses the risk of forgetting to add new tests.
We should switch this out for `go-test-split-action` and control concurrency at the matrix level.

flostadler:
The acceptance test workflows were a bit outdated. They used a hacky test provider to create VPCs and some of the tests were missing from the test matrix.
Additionally too many of the tests were running in parallel causing us to hit the account level Cluster limit (100 clusters).

This change:
- removes the test provider for creating VPCs and uses AWSX instead
- introduces `go-test-split-action` for creating the test matrix instead of manually maintaining a list
- reduces test parallelism to ~2/3 of what we had before.

I also sent quota increase request to AWS, once those get approved we can up the parallelism again.
We should also be looking into test-recording and moving some of the tests to unit-tests in order to cut down on costs and reduce the cycle time.

Fixes #1311


.github/workflows/release.yml
```diff
@@ -443,37 +443,22 @@ jobs:
...
+          working-directory: examples
+          flags: -tags=nodejs
+          total: ${{ matrix.parallel }}
+          index: ${{ matrix.index }}
       - name: Run tests
-        run: make specific_test TestName=${{ matrix.test-name }} LanguageTags=nodejs
+        run: cd examples && go test -tags=nodejs -v -json -count=1 -cover -timeout 3h -parallel 12 . --run="${{ steps.test_split.outputs.run}}" 2>&1 | tee /tmp/gotest.log | gotestfmt
     strategy:
       fail-fast: false
+      max-parallel: 10
```
flostadler, line 458: 2/3 of parallelism

.github/workflows/release.yml
```diff
@@ -569,18 +554,22 @@ jobs:
...
+          working-directory: examples
+          flags: -tags=python
+          total: ${{ matrix.parallel }}
+          index: ${{ matrix.index }}
       - name: Run tests
-        run: make specific_test TestName=${{ matrix.test-name }} LanguageTags=python
+        run: cd examples && go test -tags=python -v -json -count=1 -cover -timeout 3h -parallel 12 . --run="${{ steps.test_split.outputs.run}}" 2>&1 | tee /tmp/gotest.log | gotestfmt
     strategy:
       fail-fast: false
+      max-parallel: 4
```
flostadler, line 569: 2/3 of parallelism

.github/workflows/release.yml
```diff
@@ -678,12 +667,103 @@ jobs:
...
           yarn install && yarn link @pulumi/eks
       - name: Run tests
-        run: make specific_test TestName=${{ matrix.test-name }} LanguageTags=dotnet
-    strategy:
-      fail-fast: false
-      matrix:
-        test-name:
-          - ClusterCs
+        run: cd examples && go test -tags=dotnet -v -json -count=1 -cover -timeout 3h -parallel 12 . 2>&1 | tee /tmp/gotest.log | gotestfmt
+  test-go:
```
flostadler, line 671: We were completely missing go tests in the release workflow...

.github/workflows/release.yml
```diff
@@ -678,12 +667,103 @@ jobs:
           cd ${{ github.workspace }}/bin
           yarn install && yarn link @pulumi/eks
       - name: Run tests
-        run: make specific_test TestName=${{ matrix.test-name }} LanguageTags=dotnet
-    strategy:
-      fail-fast: false
-      matrix:
-        test-name:
-          - ClusterCs
+        run: cd examples && go test -tags=dotnet -v -json -count=1 -cover -timeout 3h -parallel 12 . 2>&1 | tee /tmp/gotest.log | gotestfmt
```
flostadler, line 671: No need to have a matrix for a single test...

examples/extra-sg-go/main.go
```diff
@@ -3,9 +3,9 @@ package main
 import (
```
flostadler, line 3: This test was forgotten from the test matrix and was broken :(

.github/workflows/release.yml
corymhall: Doesn't this mean the parallelism is 10*12?

examples/aws-profile-role-py/vpc.py
corymhall, line 1: 🤯

corymhall (approved):
Looks good, just 1 question on the parallelism

.github/workflows/release.yml
flostadler: The max test parallelism is 10*12, yes. But there's only 2-3 tests per batch.

But you're right, when we'll eventually double our tests this would be an issue again. I'm gonna change this to 6. That should be a good middle ground.

PR #1321, opened 2024-08-26
Udate cron and master github workflows

flostadler:
In https://github.com/pulumi/pulumi-eks/pull/1312 the GitHub action workflows and acceptance tests were refactored and updated.
The cron and master workflows need the same updates.

In detail this includes:
- Adding go tests, they were missing :(
- introduces go-test-split-action for creating the test matrix instead of manually maintaining a list
- reduces test parallelism to roughly ~2/3 of what we had before (that's required due to EKS cluster account limits).

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -201,6 +201,7 @@ jobs:
       - test-nodejs
       - test-python
       - test-dotnet
+      - test-go
     runs-on: ubuntu-latest
     env:
       PROVIDER_VERSION: ${{ needs.prerequisites.outputs.version }}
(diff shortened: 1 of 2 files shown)


corymhall (approved):

.github/workflows/master.yml
```diff
@@ -656,12 +647,103 @@ jobs:
...
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-region: ${{ env.AWS_REGION }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          role-duration-seconds: 7200
+          role-session-name: ${{ env.PROVIDER }}@githubActions
+          role-to-assume: ${{ secrets.AWS_CI_ROLE_ARN }}
+      - name: Configure AWS CLI
```
t0yv0, line 735: There's ways to factor out repetition but Ok for now.

PR #1324, opened 2024-08-27
Fix ManagedNodeGroups with custom launch templates always using x86_64 AMI

### LINKED ISSUE #1323: ManagedNodeGroup with custom launch template always use x86 AMI even when graviton instances are used
### What happened?

A custom launch template is created for a Node Group when `kubeletExtraArgs`, `bootstrapExtraArgs` or `enableIMDSv2` is set. In that case, the recommended AMI is retrieved for AWS parameter store.

The input property `instanceType` is used to select the right architecture. But `ManagedNodeGroup` doesn't have an `instanceType` input property, it has a list of `instanceTypes`.
This conditional here will always result in `undefined` for ManagedNodeGroups: https://github.com/pulumi/pulumi-eks/blob/23ccdc367abaa38022df6d692a659b7466825018/nodejs/eks/nodegroup.ts#L1886

And in turn, we're always defaulting to x86 as the architecture.

As a workaround, users can set the `amiType` input property to an ARM based one. For example `AL2_ARM_64 `. A full list of AMI types can be seen here: https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateNodegroup.html#API_CreateNodegroup_RequestBody.

### Example

n/a

### Output of `pulumi about`

n/a

### Additional context

_No response_

### Contributing

flostadler:
A custom launch template is created for a Node Group when `kubeletExtraArgs`, `bootstrapExtraArgs` or `enableIMDSv2` is set. In that case, the recommended AMI is retrieved from SSM Parameter Store.

The input property `instanceType` is used to select the right architecture. But `ManagedNodeGroup` doesn't have an `instanceType` input property, it has a list of `instanceTypes`.

This change fixes this by checking the `instanceTypes` property of Managed Node Groups as well.

Fixes #1323

### DIFF
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -1844,15 +1844,17 @@ Content-Type: text/x-shellscript; charset="us-ascii"
         ? { httpTokens: "required", httpPutResponseHopLimit: 2, httpEndpoint: "enabled" }
         : undefined;
 
-    const blockDeviceMappings = args.diskSize ? [
-        {
-            // /dev/xvda is the default device name for the root volume on an Amazon Linux 2 & AL2023 instance.
-            deviceName: "/dev/xvda",
-            ebs: {
-                volumeSize: args.diskSize
-            },
-        },
-    ] : undefined;
+    const blockDeviceMappings = args.diskSize
+        ? [
+              {
+                  // /dev/xvda is the default device name for the root volume on an Amazon Linux 2 & AL2023 instance.
+                  deviceName: "/dev/xvda",
+                  ebs: {
+                      volumeSize: args.diskSize,
+                  },
+              },
+          ]
+        : undefined;
 
     return new aws.ec2.LaunchTemplate(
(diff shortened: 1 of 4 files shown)


corymhall (approved):

nodejs/eks/nodegroup.ts
```diff
@@ -1844,15 +1844,17 @@ Content-Type: text/x-shellscript; charset="us-ascii"
...
-        {
-            // /dev/xvda is the default device name for the root volume on an Amazon Linux 2 & AL2023 instance.
-            deviceName: "/dev/xvda",
-            ebs: {
-                volumeSize: args.diskSize
-            },
-        },
-    ] : undefined;
+    const blockDeviceMappings = args.diskSize
+        ? [
```
t0yv0, line 1848: Curious if this is some linter/gofmt-like change that we can enforce in CI so it's not a noise in PRs.

t0yv0 (approved):

PR #1329, opened 2024-08-29
Plain methods repro

corymhall:
1. `make build`
2. `make install_nodejs_sdk`
3. `cd examples/plain-method-repro`
4. `yarn install && yarn link @pulumi/eks`
5. `PATH=... pulumi up --diff`

You will see that the `namespace` resource is using a default provider rather than the specified one (because the specified one is undefined).


flostadler (commented):
I wonder if it's caused by the provider not being part of the properties: https://github.com/pulumi/pulumi-eks/blob/9044018c15a400c46dcd264161693f6f86516e18/provider/cmd/pulumi-gen-eks/main.go#L192-L195

Not fully sure what the engine does in that case, but it might be worth adding it to the properties to try

PR #1331, opened 2024-08-30
WIP: getting method-based K8S configuration working

t0yv0:


nodejs/eks/cmd/provider/index.ts
```diff
@@ -61,9 +61,15 @@ class Provider implements pulumi.provider.Provider {
...
+        const self: ClusterInternal = inputs.__self__;
         switch (token) {
+            case "eks:index:Cluster/getK8sProvider":
+                return {
+                    outputs: {
+                        res: self.getK8sProvider(),
+                    }
+                }
             case "eks:index:Cluster/getKubeconfig":
-                const self: ClusterInternal = inputs.__self__;
```
t0yv0, line 66: Probably better to keep this under `case` since the type of __self__ depends on token. We're just lucky there's only one component resource here.

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1677,6 +1692,28 @@ func generateSchema() schema.PackageSpec {
 					"com.pulumi:kubernetes": "4.4.0",
 				},
 			}),
+			"nodejs": rawMessage(map[string]interface{}{
+				"respectSchemaVersion": true,
+				"dependencies": map[string]string{
```
t0yv0, line 1697: Do we need all of this? Curious.

t0yv0:
I think we're done here @corymhall @flostadler but I do not quite recall where we landed.

flostadler:
@t0yv0 yeah, I think this can be closed.

PR #1332, opened 2024-09-03
Support custom userdata for AL2023

flostadler:
This adds the necessary handling for `nodeadm` user data. This is used for AL2023.

`nodeadm` is a tool used for bootstrapping kubernetes nodes. It's configuration interface is yaml based and can be set via user data. The user data needs to be in MIME `multipart/mixed` format. This allows interleaving the nodeadm configuration with scripts or other user data entries. See more here: https://awslabs.github.io/amazon-eks-ami/nodeadm/.


nodejs/eks/nodegroup.ts
```diff
@@ -1750,18 +1744,18 @@ function createManagedNodeGroupInternal(
...
     // amiType, releaseVersion and version cannot be set if an AMI ID is set in a custom launch template.
     // The AMI ID is set in the launch template if custom user data is required.
     // See https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html#mng-ami-id-conditions
     if (requiresCustomUserData(customUserDataArgs)) {
         delete nodeGroupArgs.version;
         delete nodeGroupArgs.releaseVersion;
         delete nodeGroupArgs.amiType;
-    }
+    } else if (amiType === undefined && args.operatingSystem !== undefined) {
+        // if no ami type is provided, but operating system is provided, determine the ami type based on the operating system
```
t0yv0, line 1757: 👍 these narrative comments quickly fall out of date, but looks like this is still fine, it's in the "else" branch because it can actually set ami in this case.

nodejs/eks/nodegroup.ts
```diff
@@ -1944,6 +1942,19 @@ function createMNGCustomLaunchTemplate(
     );
 }
 
+function getClusterServiceCidr(
+    networkConfig: pulumi.Output<aws.types.output.eks.ClusterKubernetesNetworkConfig>,
```
t0yv0, line 1946: Is there a way to know this earlier in the Input phase not Output? E.g. to throw a better preview? or doesn't matter.

nodejs/eks/nodegroup.ts
```diff
@@ -720,49 +720,38 @@ function createNodeGroupInternal(
...
-                    certificateAuthority: clusterCa.data,
-                };
+                extraUserData,
+                stackName,
+            };
 
-                return createUserData(os, clusterMetadata, userDataArgs);
-            },
-        );
+            return createUserData(os, clusterMetadata, userDataArgs);
```
t0yv0, line 753: Is clusterMetadata here type-checked or is it <any>? 

Magic.

I see

```
export declare function all<T1, T2, T3, T4, T5, T6, T7, T8>(values: [Input<T1>, Input<T2>, Input<T3>, Input<T4>, Input<T5>, Input<T6>, Input<T7>, Input<T8>]): Output<[Unwrap<T1>, Unwrap<T2>, Unwrap<T3>, Unwrap<T4>, Unwrap<T5>, Unwrap<T6>, Unwrap<T7>, Unwrap<T8>]>;

```
It might actually type-check.

nodejs/eks/userdata.ts
```diff
@@ -211,6 +215,102 @@ ${cfnSignal}
 `;
 }
 
+function createNodeadmUserData(clusterMetadata: ClusterMetadata, args: UserDataArgs): string {
+    if (args.bootstrapExtraArgs) {
+        throw new pulumi.ResourceError(
```
t0yv0, line 220: TIL pulumi.ResourceError. Should we use it for all validation failures? Does it surface nicely?

nodejs/eks/userdata.ts
```diff
@@ -211,6 +215,102 @@ ${cfnSignal}
...
+                    spec: {
+                        kubelet: {
+                            flags: kubeletFlags,
+                        },
+                    },
+                }),
+        });
+    }
+
+    // TODO: expose extra nodeadm config options in the schema
```
t0yv0, line 270: Ticket tracking this or this is done now? Or not needed?

examples/tests/self-managed-ng-os/index.ts
```diff
@@ -0,0 +1,70 @@
...
+const instanceProfile = new aws.iam.InstanceProfile("self-managed-ng-os", {role: role});
+
+// Create a new VPC
+const eksVpc = new awsx.ec2.Vpc("self-managed-ng-os", {
+  enableDnsHostnames: true,
+  cidrBlock: "10.0.0.0/16",
+});
+
+// Create an EKS cluster.
+const cluster = new eks.Cluster("self-managed-ng-os", {
```
t0yv0, line 17: This is going to generate a custom userdata implicitly somewhere? Trying to correlate the example to the goal of the PR.

t0yv0 (approved):
Had a few questions but code looking good.

nodejs/eks/nodegroup.ts
flostadler: Yeah it's type checked :)

nodejs/eks/userdata.ts
flostadler: It associates an error with a resource. I actually never tried using a regular error to see how that gets visualized, because we use `pulumi.ResourceError` throughout the components we have.

But good catch, I missed threading the resource through to here!

nodejs/eks/userdata.ts
flostadler: It's captured in the parent ticket. I just want to tackle it in a separate PR not to blow this one up

nodejs/eks/userdata.ts
```diff
@@ -211,6 +215,102 @@ ${cfnSignal}
...
+
+    // self-managed-v1 based node groups use cloudformation to bootstrap the nodes.
+    // we need to signal to CFN that the nodes have been  successfully created by using the cfn-signal script.
+    // see: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-signal.html
+    if (isSelfManagedV1NodeUserDataArgs(args)) {
+        parts.push({
+            contentType: 'text/x-shellscript; charset="us-ascii"',
+            content: `#!/bin/bash
+
+/opt/aws/bin/cfn-signal --exit-code $? --stack ${args.stackName} --resource NodeGroup --region ${args.awsRegion}
```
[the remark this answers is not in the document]
flostadler: Yes, that's the path to the `cfn-signal` script of EC2 instances: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-signal.html

This is baked into so many tools, there's pretty much no way AWS can change/remove this without releasing a new major release of their OS altogether

examples/tests/self-managed-ng-os/index.ts
flostadler: Yes, self managed node groups need user data because they deploy an auto scaling group for the nodes. Everything (lifecycle, config, etc.) is self managed and needs to be provided for those to work. So this requires the generated user data.

With Managed Node Groups, there's a base configuration AWS provides and everything users configure gets merged with the base configuration. In most cases this doesn't result in custom user data. But for certain configuration parameters it is necessary.

nodejs/eks/userdata.ts
t0yv0: I was just recently having trouble making validation errors presentable but I didn't notice this one. Maybe it's the answer.

nodejs/eks/nodegroup.ts
flostadler: We cannot know it earlier because EKS sets the service CIDR unless the user overrides it.

But overall it doesn't matter. This is used only in the user data and that is fully dependent on multiple outputs of the cluster.

nodejs/eks/userdata.ts
flostadler: I don't think it's the full answer. It still includes the GRPC trace and we cannot anchor them to specific fields

t0yv0: Ah ok, sad.

PR #1334, opened 2024-09-04
Add support for custom bottlerocket user data

flostadler:
### Proposed changes

This change adds the necessary options for configuring settings for Bottlerocket Operating Systems.
Bottlerocket configuration is driven by a user data script in TOML format.

The provider will set the base configuration that's necessary for nodes to successfully register with the kubernetes cluster. Users will have the ability to add additional settings or override the base configuration by using the `bottlerocketSettings` parameter. An overview of the settings can be seen here: https://bottlerocket.dev/en/os/1.20.x/api/settings/.

### New dependencies
This adds two dependencies to the provider.
1. `@iarna/toml"`: For converting the configuration into valid TOML. This allows us to expose `bottlerocketSettings` as an object instead of a string so users do not need to worry about TOML formatting&serialization
2. `ipaddr.js`: For calculating the `cluster-dns-ip`. We could definitely create our own IP parser, it's not overly complex for IPv4. But IPv6 can be more tricky with shortened formats. This is unnecessary complexity for the provider.

Both of them are maintained libraries with 0 other dependencies.

### Remarks
Bottlerocket will only be supported with `NodeGroupV2` and `ManagedNodeGroup` components because the older `NodeGroup` (aka `NodeGroupV1`) uses CloudFormation under the hood and nodes need to signal that they're ready. Bottlerocket can't execute scripts as part of the boot up because it doesn't have a shell, so this is not supported.


t0yv0:
Just from reading this description, is change detection TOML-normalizing and consistent? E.g. it's not possible for accidental meaningless changes in object field order serialization to trigger update diffs?

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -890,6 +890,17 @@ func generateSchema() schema.PackageSpec {
...
+					"bottlerocketSettings": {
+						TypeSpec: schema.TypeSpec{
+							Type: "object",
+							AdditionalProperties: &schema.TypeSpec{
+								Ref: "pulumi.json#/Any",
+							},
+						},
+						Description: "The configuration settings for Bottlerocket OS.\n" +
+							"The settings will get merged with the base settings the provider uses to configure Bottlerocket.\n\n" +
+							"For an overview of the available settings, see https://bottlerocket.dev/en/os/1.20.x/api/settings/.",
```
t0yv0, line 902: Typifying https://bottlerocket.dev/en/os/1.20.x/api/settings-index/ seems possible with some legwork. There seems to be an OpenAPI definition somewhere.

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -890,6 +890,17 @@ func generateSchema() schema.PackageSpec {
...
 					},
+					"bottlerocketSettings": {
+						TypeSpec: schema.TypeSpec{
+							Type: "object",
+							AdditionalProperties: &schema.TypeSpec{
+								Ref: "pulumi.json#/Any",
+							},
+						},
+						Description: "The configuration settings for Bottlerocket OS.\n" +
+							"The settings will get merged with the base settings the provider uses to configure Bottlerocket.\n\n" +
```
t0yv0, line 901: Wonder if we can link to a doc or inline something about which base settings are these?

examples/tests/managed-ng-os/index.ts
```diff
@@ -52,34 +64,82 @@ const managedNodeGroupAL2023Arm = eks.createManagedNodeGroup("al-2023-arm-mng",
...
+const managedNodeGroupBottlerocketUserData = eks.createManagedNodeGroup("bottlerocket-mng-userdata", {
+  ...scalingConfig,
+  cluster: cluster,
+  operatingSystem: eks.OperatingSystem.Bottlerocket,
+  instanceTypes: ["t3.medium"],
+  nodeRole: role,
+  labels: {
+    "increased-pod-capacity": "true",
+  },
+  bottlerocketSettings: {
```
t0yv0, line 119: Nice!!

nodejs/eks/nodegroup.ts
```diff
@@ -1837,56 +1851,38 @@ function createMNGCustomLaunchTemplate(
...
-                        userDataOverride: undefined,
-                    };
-
-                    const userData = createUserData(os, clusterMetadata, userDataArgs, parent);
-                    return Buffer.from(userData, "utf-8").toString("base64");
-                },
-            );
+                    bottlerocketSettings,
+
+                    // TODO: expose this as an option for managed node groups
```
t0yv0, line 1879: Looks like a pre-existing TODO lacking a matching ticket. We usually have `TODO[pulumi/pulumi-eks#12345]: description` format for these.

nodejs/eks/userdata.ts
```diff
@@ -346,3 +378,141 @@ function buildKubeletFlags(args: UserDataArgs): string[] {
...
+ * 10th IP address in the service CIDR. See: https://github.com/awslabs/amazon-eks-ami/blob/4f9b5560aee4dd9adbd4f1c711d237fd6f0b4e70/templates/al2/runtime/bootstrap.sh#L462-L485
+ *
+ * The calculation is done by round tripping the network address of the CIDR to a big integer, adding 10 to it, and then converting it back.
+ * This way the calculation is safe for both IPv4 and IPv6.
+ *
+ * @param serviceCidr - The service CIDR to parse and calculate the cluster DNS IP from.
+ * @returns The cluster DNS IP address.
+ * @throws {pulumi.ResourceError} If the service CIDR fails to parse or the cluster DNS IP is out of range.
+ */
+export function getClusterDnsIp(serviceCidr: string, parent: pulumi.Resource | undefined): string {
```
t0yv0, line 484: Wild.

t0yv0 (approved):

t0yv0:
One small technicality is that I always worry about the `any` type, namely that we're going to forget that something is an Input and assume it's the promptly available object. One place where this happens is here:

```
    const customUserDataArgs = {
        kubeletExtraArgs: args.kubeletExtraArgs,
        bootstrapExtraArgs: args.bootstrapExtraArgs,
        bottlerocketSettings: args.bottlerocketSettings,
    };
```

We assign args.bottlerocketSettings of type 'Input<any>' to a slot of type '<any>'. 

I think it is fine though as the code does pulumi.all discipline before processing the data, just it's not enforced by types.

provider/cmd/pulumi-gen-eks/main.go
flostadler: I did look into this, but they do not have any types for the actual configuration options: https://github.com/bottlerocket-os/bottlerocket/blob/a3e8bda116dcf4b06fdf6e4cea42da1a005e926f/sources/api/openapi.yaml#L108-L109

There's also the problem of choosing the right version of the typed object for the right bottlerocket version. E.g. there might be a change in the bottlerocket config, we'd have to accept all different versions of bottlerocket config. This would essentially cause us to fall back to accepting anything and we need an escape hatch, otherwise users will always have to wait for us to release updates.

So IMO this wouldn't really bring any benefits, rather complications for users.
My reasoning would be different if bottlerocket offered rich types though.

flostadler:
> One small technicality is that I always worry about the `any` type, namely that we're going to forget that something is an Input and assume it's the promptly available object. One place where this happens is here:
> 
> ```
>     const customUserDataArgs = {
>         kubeletExtraArgs: args.kubeletExtraArgs,
>         bootstrapExtraArgs: args.bootstrapExtraArgs,
>         bottlerocketSettings: args.bottlerocketSettings,
>     };
> ```
> 
> We assign args.bottlerocketSettings of type 'Input' to a slot of type 'any'.
> 
> I think it is fine though as the code does pulumi.all discipline before processing the data, just it's not enforced by types.

Yeah this is tricky. We want to kind of accept arbitrary data, but on the other hand exclude inputs. The good thing is that this is guarded by tests. So if we ever introduce a bug here it shouldn't hit users.

flostadler:
> Just from reading this description, is change detection TOML-normalizing and consistent? E.g. it's not possible for accidental meaningless changes in object field order serialization to trigger update diffs?

Good point! I added deep sorting (and tests) to ensure this cannot happen

PR #1337, opened 2024-09-05
Support configuring disk size of self-managed bottlerocket node groups

flostadler:
### Proposed changes

The provider wrongly assumed that an AMI only has a single block device. But Bottlerocket has two.
Bottlerocket has two block devices, the root device stores the OS itself and the other is for data like images, logs, persistent storage. We need to allow users to configure the block device for data.

With this change, the provider will choose the block device that gets modified depending on the OS. If the OS is bottlerocket, the data device gets modified.
This also adds E2E tests verifying that the node storage capacity correctly reflects user settings.

### Related issues (optional)

- This builds on top of #1334

### DIFF
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -1322,7 +1322,7 @@ function createNodeGroupV2Internal(
                     ebs: {
                         encrypted: pulumi
                             .output(args.nodeRootVolumeEncrypted)
-                            .apply((val) => (val === true ?? false ? "true" : "false")),
+                            .apply((val) => (val ? "true" : "false")),
                         volumeSize: args.nodeRootVolumeSize ?? 20, // GiB
                         volumeType: args.nodeRootVolumeType ?? "gp2",
                         iops: args.nodeRootVolumeIops,
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -847,7 +847,8 @@ function createNodeGroupInternal(
     const { rootBlockDevice, ebsBlockDevices } = pulumi
         .all([os, amiInfo])
         .apply(([os, amiInfo]) => {
-            if (os !== OperatingSystem.Bottlerocket) {
+            if (os === OperatingSystem.AL2) {
+                // Old behavior to stay backwards compatible
                 return {
                     rootBlockDevice: {
                         encrypted: args.nodeRootVolumeEncrypted ?? false,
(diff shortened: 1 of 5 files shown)


nodejs/eks/nodegroup.ts
```diff
@@ -841,14 +894,8 @@ function createNodeGroupInternal(
                 .all([nodeSecurityGroupId, extraNodeSecurityGroupIds])
                 .apply(([sg, extraSG]) => [sg, ...extraSG]),
             spotPrice: args.spotPrice,
-            rootBlockDevice: {
```
flostadler, line 844: I wish we would've just used `ebsBlockDevices` here. That would've simplified this a lot...
To stay backwards compatible we now have to do this if/else dance above :(

nodejs/eks/nodegroup.ts
```diff
@@ -829,6 +829,59 @@ function createNodeGroupInternal(
...
+
+    const { rootBlockDevice, ebsBlockDevices } = pulumi
+        .all([os, amiInfo])
+        .apply(([os, amiInfo]) => {
+            if (os !== OperatingSystem.Bottlerocket) {
+                return {
+                    rootBlockDevice: {
+                        encrypted: args.nodeRootVolumeEncrypted ?? false,
+                        volumeSize: args.nodeRootVolumeSize ?? 20, // GiB
+                        volumeType: args.nodeRootVolumeType ?? "gp2",
```
corymhall, line 855: Probably out of scope for this PR, but it would probably be good to update the default to `gp3`. I _think_ that the migration is non-breaking, but we would have to test
https://aws.amazon.com/blogs/storage/migrate-your-amazon-ebs-volumes-from-gp2-to-gp3-and-save-up-to-20-on-costs/

nodejs/eks/nodegroup.ts
```diff
@@ -829,6 +829,59 @@ function createNodeGroupInternal(
...
+                        iops: args.nodeRootVolumeIops,
+                        throughput: args.nodeRootVolumeThroughput,
+                        deleteOnTermination: args.nodeRootVolumeDeleteOnTermination ?? true,
+                    },
+                    ebsBlockDevices: [],
+                };
+            }
+
+            // Bottlerocket has two block devices, the root device stores the OS itself and the other is for data like images, logs, persistent storage
```
corymhall, line 864: I'm assuming that the root device is not something that can be configured by the user?

corymhall (approved):
Looks good!

nodejs/eks/nodegroup.ts
flostadler: Yeah in theory I agree with updating this here, the issue is that this will update the node groups and cause them to get cycled.
For self managed node groups this is an operation that will cause downtime when not done carefully (cordon & drain the nodes before running the update).

But this actually got me thinking. We shouldn't set any defaults if the OS is set to AL2023/Bottlerocket

But

nodejs/eks/nodegroup.ts
flostadler: No it's not. That's configured in the AMI

flostadler:
@corymhall I updated the code to not apply any defaults for the storage settings when the OS is AL2023/Bottlerocket. That way we'll be able to get all of Amazon's improvements OOTB when they push new recommended AMIs.

For now I left the hard coded defaults in for AL2, but we can think about removing them altogether with the next major release.

nodejs/eks/nodegroup.ts
```diff
@@ -1232,30 +1298,83 @@ function createNodeGroupV2Internal(
...
+        if (os === OperatingSystem.AL2) {
+            return [
                 {
-                    deviceName: device,
+                    deviceName: deviceName,
                     ebs: {
-                        encrypted: args.nodeRootVolumeEncrypted ?? false ? "true" : "false",
+                        encrypted: pulumi
+                            .output(args.nodeRootVolumeEncrypted)
+                            .apply((val) => (val === true ?? false ? "true" : "false")),
```
corymhall, line 1325:
```suggestion
                    .apply((val) => (val ? "true" : "false")),
```

I think this is what you mean?

corymhall (approved):

nodejs/eks/nodegroup.ts
flostadler: Ah yeah that's the same... good catch!

PR #1338, opened 2024-09-05
eks.NodeGroupV2 will now autoname the autoscaling group

### LINKED ISSUE #1335: eks.NodeGroupV2 does not randomize the suffix of the underlying Auto Scaling group
### What happened?

Creating an eks.NodeGroupV2 does not randomize the suffix of the underlying Auto Scaling group. This makes tests harder because one test failure that fails to clean up leads to subsequent test failures conflicting on the actual Auto Scaling group name.  I wonder if this is also something of value to end users outside of test scenarios, as by default Pulumi works hard to avoid name conflicts. 

The example is taken from our tests (examples/tags/index.ts). 

```
const ng2 = new eks.NodeGroupV2("example-ng-tags-ng2", {
    cluster: cluster2,
    instanceType: "t3.medium",
    desiredCapacity: 1,
    minSize: 1,
    maxSize: 2,
    instanceProfile: instanceProfile0,
    labels: {"preemptible": "true"},
    taints: {
        "special": {
            value: "true",
            effect: "NoSchedule",
        },
    },
    autoScalingGroupTags: cluster2.core.cluster.name.apply(clusterName => ({
        "myAutoScalingGroupTag3": "true",
        "k8s.io/cluster-autoscaler/enabled": "true",
        [`k8s.io/cluster-autoscaler/${clusterName}`]: "true",
    }))
})
```

### Example

The relevant code is this I think:

```
    const asGroup = new aws.autoscaling.Group(
        name,
        {
            name: name,
```

https://github.com/pulumi/pulumi-eks/blob/master/nodejs/eks/nodegroup.ts#L1302

I think specifying name: name opts out of the AWS auto-naming feature which would otherwise kick in. 

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

### LINKED ISSUE #1336: Workflow failure: cron
## Workflow Failure

[cron](https://github.com/pulumi/pulumi-eks/blob/master/.github/workflows/cron.yml) has failed. See the list of failures below:

- [2024-09-05T08:23:48.000Z](https://github.com/pulumi/pulumi-eks/actions/runs/10715046075)

t0yv0:
### Proposed changes

Before this change, eks.NodeGroupV2 would create an aws.autoscaling.Group resource with a fixed name, and the users
would need to make sure the name does not conflict with any unrelated resources in their AWS account. After the change,
the name automatically receives a randomized suffix, making eks.NodeGroupV2 easier to use correctly.

Users with existing eks.NodeGroupV2 instances will not be affected by this change as the new provider version will
respect the names allocated by the previous version.

### Related issues

Fixes #1335
Fixes #1336

### DIFF
--- a/examples/examples_nodejs_test.go
+++ b/examples/examples_nodejs_test.go
@@ -269,6 +269,12 @@ func TestAccTags(t *testing.T) {
 					info.Outputs["kubeconfig1"],
 					info.Outputs["kubeconfig2"],
 				)
+				for _, r := range info.Deployment.Resources {
+					if r.Type == "aws:autoscaling/group:Group" {
+						assert.NotEqualf(t, "example-ng-tags-ng2", string(r.ID),
+							"aws.autoscaling.Group should have an autonamed random suffix")
+					}
+				}
 			},
 		})
 
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -1298,7 +1298,6 @@ ${customUserData}
     const asGroup = new aws.autoscaling.Group(
         name,
         {
-            name: name,
             minSize: args?.minSize ?? 1,
             maxSize: args?.maxSize ?? 2,
             desiredCapacity: args?.desiredCapacity ?? 2,


flostadler (approved):
LGTM assuming the upgrade tests pass

PR #1339, opened 2024-09-05
Add missing configuration options for Managed Node Groups

flostadler:
The `ManagedNodeGroup` component was missing configuration the other node groups had.
In detail, that's `amiId`, `gpu` and `userData`.

Those will allow booting specific/custom AMIs, nodes with GPUs or setting custom user data.
The added E2E tests ensure this works as expected.

Relates to https://github.com/pulumi/pulumi-eks/issues/1224

### DIFF
--- a/examples/tests/managed-ng-os/index.ts
+++ b/examples/tests/managed-ng-os/index.ts
@@ -104,12 +104,12 @@ const managedNodeGroupBottlerocket = eks.createManagedNodeGroup("bottlerocket-mn
   nodeRole: role,
 });
 
-// Create a Bottlerocket node group with GPU support
+// Create a Bottlerocket node group with GPU support (it's the cheapest GPU instance type)
 const managedNodeGroupBottlerocketGpu = eks.createManagedNodeGroup("bottlerocket-mng-gpu", {
   ...scalingConfig,
   cluster: cluster,
   operatingSystem: eks.OperatingSystem.Bottlerocket,
-  instanceTypes: ["t3.medium"],
+  instanceTypes: ["g5g.xlarge"],
   nodeRole: role,
   gpu: true,
 });
--- a/examples/tests/managed-ng-os/index.ts
+++ b/examples/tests/managed-ng-os/index.ts
@@ -69,8 +69,10 @@ const managedNodeGroupAL2023UserData = eks.createManagedNodeGroup("al-2023-mng-u
   operatingSystem: eks.OperatingSystem.AL2023,
   instanceTypes: ["t3.medium"],
   nodeRole: role,
+  diskSize: 100,
   labels: {
     "increased-pod-capacity": "true",
+    "increased-storage-capacity": "true",
   },
   kubeletExtraArgs: `--max-pods=${increasedPodCapacity}`,
 });
@@ -82,8 +84,10 @@ const managedNodeGroupAL2023ArmUserData = eks.createManagedNodeGroup("al-2023-ar
   operatingSystem: eks.OperatingSystem.AL2023,
   instanceTypes: ["t4g.medium"],
   nodeRole: role,
+  diskSize: 100,
   labels: {
     "increased-pod-capacity": "true",
+    "increased-storage-capacity": "true",
   },
   kubeletExtraArgs: `--max-pods=${increasedPodCapacity}`,
 });
@@ -113,8 +117,10 @@ const managedNodeGroupBottlerocketUserData = eks.createManagedNodeGroup("bottler
   operatingSystem: eks.OperatingSystem.Bottlerocket,
   instanceTypes: ["t3.medium"],
   nodeRole: role,
+  diskSize: 100,
   labels: {
     "increased-pod-capacity": "true",
+    "increased-storage-capacity": "true",
   },
   bottlerocketSettings: {
     settings: {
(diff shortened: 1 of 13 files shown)


flostadler:
The diff got messed up when the base PR got rebased. Will need to untangle this..

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -905,6 +905,26 @@ func generateSchema() schema.PackageSpec {
...
 							"For an overview of the available settings, see https://bottlerocket.dev/en/os/1.20.x/api/settings/.",
 					},
+					"userData": {
+						TypeSpec: schema.TypeSpec{Type: "string"},
+						Description: "User specified code to run on node startup. This is expected to handle " +
+							"the full AWS EKS node bootstrapping. If omitted, the provider will configure the user data.\n\n" +
+							"See for more details: https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html#launch-template-user-data.",
+					},
+					"gpu": {
+						TypeSpec: schema.TypeSpec{Type: "boolean"},
```
t0yv0, line 915: Would this be any better with defining the explicit default in the schema?

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -905,6 +905,26 @@ func generateSchema() schema.PackageSpec {
...
+						Description: "Use the latest recommended EKS Optimized AMI with GPU support for the " +
+							"worker nodes.\nDefaults to false.\n\n" +
+							"Note: `gpu` and `amiId` are mutually exclusive.\n\n" +
+							"See for more details: https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-amis.html.",
+					},
+					"amiId": {
+						TypeSpec: schema.TypeSpec{Type: "string"},
+						Description: "The AMI ID to use for the worker nodes.\nDefaults to the latest recommended " +
```
t0yv0, line 923: Likewise, should schema define `Default: "latest"` or some such?

nodejs/eks/nodegroup.ts
```diff
@@ -2105,9 +2163,6 @@ function getRecommendedAMI(
     k8sVersion: pulumi.Output<string>,
     parent: pulumi.Resource | undefined,
 ): pulumi.Input<string> {
-    // TODO[pulumi/pulumi-eks#1195]: Add gpu argument to ManagedNodeGroupOptions
```
t0yv0, line 2108: Is this todo fixed by this PR?

flostadler:
> Does this cover #1224 GPU support?

Yes it does! I folded that into this PR because it's relevant here as well. (added this to the PR description)

provider/cmd/pulumi-gen-eks/main.go
flostadler: The default value here is not "latest". The provider performs actions in the background to fetch the latest recommended AMI ID.

nodejs/eks/nodegroup.ts
```diff
@@ -1815,6 +1843,19 @@ function createManagedNodeGroupInternal(
...
     }
 
+    const amiIdMutuallyExclusive: (keyof Omit<ManagedNodeGroupOptions, "cluster">)[] = [
+        "gpu",
+        "amiType",
+    ];
+    amiIdMutuallyExclusive.forEach((key) => {
+        if (args.amiId && args[key]) {
+            throw new pulumi.ResourceError(
+                `amiId and ${key} are mutually exclusive to create a managed node group`,
```
t0yv0, line 1853: The grammar here seems a little funny (but also I'm not a native speaker). Feed it to LLM?

examples/tests/managed-ng-os/index.ts
```diff
@@ -101,6 +104,16 @@ const managedNodeGroupBottlerocket = eks.createManagedNodeGroup("bottlerocket-mn
...
 });
 
+// Create a Bottlerocket node group with GPU support (it's the cheapest GPU instance type)
+const managedNodeGroupBottlerocketGpu = eks.createManagedNodeGroup("bottlerocket-mng-gpu", {
+  ...scalingConfig,
+  cluster: cluster,
+  operatingSystem: eks.OperatingSystem.Bottlerocket,
+  instanceTypes: ["g5g.xlarge"],
+  nodeRole: role,
+  gpu: true,
```
t0yv0, line 114: Neat.

nodejs/eks/nodegroup.ts
flostadler: Yes!

nodejs/eks/nodegroup.ts
flostadler: I actually copied the grammar from a similar error message from above. But now that you're saying this it also sounds weird to me (also not native speaker).

Let me reword those.

t0yv0 (approved):
🚢 

The settings with exclusivity constraints are getting a bit tricky to reason about but maybe fine for now. If this continues to grow we could try an internal TypeScript union model to parse into with several variants of settings.

provider/cmd/pulumi-gen-eks/main.go
t0yv0: So it does so when amiId is empty/undefined, the provider does the search. OK, I guess the current form is the best.

provider/cmd/pulumi-gen-eks/main.go
flostadler: We need to know whether it's false or not defined. There's certain combinations that do not work (e.g. you cannot choose an AMI with GPU and then set GPU to false).
Is my understanding wrong that we'd receive `false` in the provider in that case?

PR #1340, opened 2024-09-06
Add configuration parameter for nodeadm configuration to node groups

flostadler:
### Proposed changes

This change adds a new input property called `nodeadmExtraConfig` to the node group components. This property will allow injecting additional nodeadm sections into the user data.
This can be virtually anything. Some data, a shell script, or additional nodeadm [`NodeConfig`](https://awslabs.github.io/amazon-eks-ami/nodeadm/).

The nodeadm user data is a MIME multipart/mixed document. And every section has string based `content` and a MIME multipart `contentType`.

Right now there's no straight forward way to generate types for the nodeadm `NodeConfig` because it's not schematized. Work for enhancing this is tracked here: https://github.com/pulumi/pulumi-eks/issues/1341.


nodejs/eks/userdata.ts
```diff
@@ -300,6 +310,13 @@ function createNodeadmUserData(
         });
     }
 
+    // add extra nodeadm options if provided
+    if (args.nodeadmExtraOptions) {
+        for (const option of args.nodeadmExtraOptions) {
+            parts.push(option);
```
t0yv0, line 316: Will this override existing settings or how does the merging work?

flostadler: nodeadm will evaluate all `NodeConfig` sections and merge them together (deep merge) in the order they were specified.
So user settings overwrite base settings.

examples/tests/managed-ng-os-py/__main__.py
```diff
@@ -71,6 +71,27 @@
                                            labels={ "increased-pod-capacity": "true" },
                                            kubelet_extra_args="--max-pods=100")
 
+nodeadm_extra_config = """---
```
t0yv0, line 74: Ever makes sense to accept dicts here or the YAML literals are more natural for the users?

t0yv0:
Implementation looks good! I don't have a good handle on these use cases, would appreciate others to pipe in. It's a little imprecise but if this escape hatch alleviates user pain and makes things more straightforward to work around then perhaps it belongs.

sdk/python/pulumi_eks/outputs.py
```diff
@@ -398,6 +402,15 @@ def __init__(__self__, *,
         :param str node_user_data_override: User specified code to run on node startup. This code is expected to handle the full AWS EKS bootstrapping code and signal node readiness to the managing CloudFormation stack. This code must be a complete and executable user data script in bash (Linux) or powershell (Windows).
                
                See for more details: https://docs.aws.amazon.com/eks/latest/userguide/worker.html
+        :param Sequence['NodeadmOptions'] nodeadm_extra_options: Extra nodeadm configuration sections to be added to the nodeadm user data. This can be shell scripts, nodeadm NodeConfig or any other user data compatible script. When configuring additional nodeadm NodeConfig sections, they'll be merged with the base settings the provider sets.
```
t0yv0, line 405: This is already a generous doc comment but I would consider even more explicit, call out that these settings augment or override the base settings, and maybe give a quick motivating example (or a link to one).

examples/tests/managed-ng-os-py/__main__.py
flostadler: This section can be anything (bash scripts, yaml, etc.), so we'll need to be able to accept all of that.

What I thought of is that we add a function (like we have for IAM policies) for crafting the `NodeConfig` form a typed object.
But that has challenges associated with it. Right now AWS isn't publishing a typed schema for this (the settings are just `object`) and we'll need to come up with a versioning strategy. Tracked here: https://github.com/pulumi/pulumi-eks/issues/1341.

For now I think a plain old string is the most convenient for the users and will cause the least pain in the long run if we end up adding ways to compose the yaml from typed dicts.

sdk/python/pulumi_eks/outputs.py
flostadler: They do not necessarily overwrite. They can also be scripts for example. I can make this a bit clearer in the docs!

flostadler:
Example use cases for this are:
- Executing a bash script on node boot up. You might have to start some security agent on the node, load a kernel module or do some other startup handling that needs to be done.
- An example for adding to the `NodeConfig` is configuring advanced parameters like `kubeAPIQPS ` (to control how many QPS the kubelet is allowed to do when talking with the API server).

t0yv0 (approved):

PR #1342, opened 2024-09-06
Deprecate aws-auth ConfigMap

### LINKED ISSUE #1213: Deprecate aws-auth ConfigMap
## Hello!

## Issue details

[AWS recently deprecated](https://docs.aws.amazon.com/eks/latest/userguide/auth-configmap.html) the aws-auth ConfigMap. Currently this is still the default authentication mode for the Cluster component and we should replace that with the `API` (Access Entries) authentication mode.

We should also mark the usage of the aws-auth ConfigMap as deprecated. This includes deprecating the enum values for the authentication modes `CONFIG_MAP` and `API_AND_CONFIG_MAP`.

Additionally we should publish the [doc for the authentication mode migration](https://github.com/pulumi/pulumi-eks/blob/master/docs/authentication-mode-migration.md) as a How-To to the registry in order to surface it broader.
For reference, here's how the AWS How-To's are managed: https://github.com/pulumi/registry/tree/master/themes/default/content/registry/packages/aws/how-to-guides

### Affected area/feature
- pulumi-eks

flostadler:
Following up on AWS' deprecation of the aws-auth ConfigMap.
We're deprecating the enums for the `CONFIG_MAP` and `API_AND_CONFIG_MAP`
AuthenticationModes.
This will steer users into the right direction of using the `API` AuthenticationMode.

Details around the deprecation are here: https://docs.aws.amazon.com/eks/latest/userguide/auth-configmap.html
This is a good guide for how to migrate away from it: https://docs.aws.amazon.com/eks/latest/userguide/migrating-access-entries.html

We have a Pulumi migration guide for this in docs/authentication-mode-migration.md. As a follow up we want
to publish this to the Pulumi Docs as a How-To.

Closes #1213

### DIFF
--- a/provider/cmd/pulumi-gen-eks/main.go
+++ b/provider/cmd/pulumi-gen-eks/main.go
@@ -1624,6 +1624,8 @@ func generateSchema() schema.PackageSpec {
 						Name:        "ConfigMap",
 						Value:       "CONFIG_MAP",
 						Description: "Only aws-auth ConfigMap will be used for authenticating to the Kubernetes API.",
+						DeprecationMessage: "The aws-auth ConfigMap is deprecated. The recommended method to manage access to Kubernetes APIs is Access Entries with the AuthenticationMode API.\n" +
+							"For more information and instructions how to upgrade, see https://docs.aws.amazon.com/eks/latest/userguide/migrating-access-entries.html.",
 					},
 					{
 						Name:        "Api",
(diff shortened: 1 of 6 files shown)


t0yv0:
Let's track "As a follow up we want to publish this to the Pulumi Docs as a How-To." explicitly?

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1624,6 +1624,8 @@ func generateSchema() schema.PackageSpec {
 						Name:        "ConfigMap",
 						Value:       "CONFIG_MAP",
 						Description: "Only aws-auth ConfigMap will be used for authenticating to the Kubernetes API.",
+						DeprecationMessage: "The aws-auth ConfigMap is deprecated. The recommended method to manage access to Kubernetes APIs is Access Entries with the AuthenticationMode API.\n" +
+							"For more information and instructions how to upgrade, see https://docs.aws.amazon.com/eks/latest/userguide/migrating-access-entries.html.",
```
t0yv0, line 1628: It sounds like we had a Pulumi-specific upgrade document, should we consider linking that instead of the baseline aws.amazon.com migration document? Thanks!

t0yv0 (approved):

provider/cmd/pulumi-gen-eks/main.go
flostadler: I want to link this once we have it published in the docs. This is tracked in the ticket for the how-tos

flostadler:
> Let's track "As a follow up we want to publish this to the Pulumi Docs as a How-To." explicitly?

It's tracked here: https://github.com/pulumi/home/issues/3626

Issue #747, opened 2022-07-26
Option to skip creating node and cluster security groups


vinay-kosaraju:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently, it looks like [node security group](https://github.com/pulumi/pulumi-eks/blob/v0.41.2/nodejs/eks/cluster.ts#L1437) and [cluster security group](https://www.pulumi.com/registry/packages/eks/api-docs/cluster/#clustersecuritygroup_go) with default ingress and egress rules are created if they are not specified in the cluster args. 

Could we please get an option to skip creation of the security groups and let EKS create the security group with default rules? (https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html)

viveklak:
@roothorp could you take a look and prioritize accordingly?

t0yv0:
NOTE: skipDefaultNodeGroup should be implied by skipping security group.

t0yv0:
Quick update, apologies for the further delay but we decided to defer this item until the upcoming major release of the provider. It appears that the current signature implies that the eks.Cluster resource must return non-optional Node Security Group and Cluster Ingress rule. Making these optional is a technically breaking change, and we considered sending some sentinel values but decided against it for the moment.

IN the upcoming major release the default NodeGroup is going to be removed; the default behavior will use a ManagedNodeGroup but there will be a way to opt out of it. We will need to revisit these settings in the new context, likely will be either removed or moved optional for special cases and not used by default. 

We will track this item and follow up as part of the major release work.

cleverguy25:
Added to epic https://github.com/pulumi/home/issues/3558

flostadler:
We've addressed this as part of https://github.com/pulumi/pulumi-eks/pull/1416. This will be released with EKS v3

PR #1344, opened 2024-09-06
Add an eks.Cluster option to use an EKS-managed Security Group

t0yv0:
### Proposed changes

Currently by default eks.Cluster can utilize a user-provided Security Group or it can create one in the Pulumi layer. After this change users may opt into letting EKS create the security group. It still can be referenced in Pulumi via SecurityGroup.get machinery under the hood but this is transparent to users.

Example usage:

```typescript
const cluster = new eks.Cluster(`my-cluster`, {
    nodeAmiId: "ami-0384725f0d30527c7",
    clusterSecurityGroupDefault: true,
});
```

### Related issues (optional)

Relates: #747 

To fully close the above issue another PR is needed for the node group.

Design note: currently we think that using a Pulumi-provisioned SG is the right default, but the switch is added for users that require it. 

#### Official details on the default SG

https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html

### DIFF
--- a/examples/cluster/index.ts
+++ b/examples/cluster/index.ts
@@ -95,7 +95,7 @@ export const kubeconfig2: pulumi.Output<any> = cluster2.kubeconfig;
 export const kubeconfig3: pulumi.Output<any> = cluster3.kubeconfig;
 export const kubeconfig4: pulumi.Output<any> = cluster4.kubeconfig;
 export const kubeconfig5: pulumi.Output<any> = cluster5.kubeconfig;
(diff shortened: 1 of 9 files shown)


t0yv0:
The new option must be made to conflict with clusterSecurityGroup.

flostadler:
The default security group rules mentioned in the PR description aren't fully accurate.
The default cluster security group allows all egress traffic and all intra cluster traffic. That is managed node group to managed node group and managed node group to control plane

PR #1357, opened 2024-09-12
Use EKS Addon to manage `kube-proxy` and `coredns`

### LINKED ISSUE #1254: Auto update EKS Cluster Components when updating EKS cluster version
## Hello!

## Issue details

Right now, EKS cluster components (`coredns`, `kube-proxy`) do not take the clusters version into account after the initial deployment. If the cluster version is updated, the cluster components still stay at the old version. Right now this requires users to either adopt the cluster components into Pulumi as addons or manually trigger and update using the EKS API/AWS Console. 
Ideally the EKS Cluster Components should follow the cluster version by default.

This is how you'd adopt the cluster components into your Pulumi program as addons:
```
const kubeProxyVersion = aws.eks.getAddonVersionOutput({
  addonName: "kube-proxy",
  kubernetesVersion: cluster.eksCluster.version,
  mostRecent: true, // whether to return the default version or the most recent version for the specified kubernetes version
}).apply((addonVersion) => addonVersion.version);

const kubeProxyAddon = new aws.eks.Addon("example", {
  clusterName: cluster.eksCluster.name,
  addonName: "kube-proxy",
  addonVersion: kubeProxyVersion,
});
```

### LINKED ISSUE #1261: Use EKS Addon to manage kube-proxy and coredns
## Hello!

## Issue details

Right now kube-proxy and coredns are self-managed. This has the drawback that you cannot keep those cluster components updated via pulumi.

Additionally, this requires us to do a complex process of shelling out to kubectl to patch the coredns deployment. This is required so coredns can run on fargate:
https://github.com/pulumi/pulumi-eks/blob/986c528a058381be989c382d38a2e74238c9502b/nodejs/eks/cluster.ts#L967-L1024

Instead of doing that, we can manage those components as EKS addons and configure them with the pulumi resource `aws.eks.Addon`.

E.g. for enabling coredns to run on Fargate we should set:
```
const example = new aws.eks.Addon("example", {
    clusterName: "mycluster",
    addonName: "coredns",
    ...
    configurationValues: JSON.stringify({
        computeType: "Fargate"
    })
};
```

corymhall:
### Proposed changes

This PR switches the `coredns` and `kube-proxy` addons from self-managed to managed. By default the latest compatible version will be used.

This also introduces two new top level arguments to `ClusterOptions` for configuring these new addons.

- `corednsAddonOptions`
- `kubeProxyAddonOptions`

BREAKING CHANGE: creating an `eks.Cluster` will now also create the `coredns` and `kube-proxy` addons. If you are currently already managing these you will need to disable the creation of these through the new arguments `ClusterOptions.corednsAddonOptions.enabled = false` and `ClusterOptions.kubeProxyAddonOptions.enabled = false`

### Related issues (optional)

closes #1261, closes #1254


examples/cluster/index.ts
```diff
@@ -39,6 +39,9 @@ const cluster2 = new eks.Cluster(`${projectName}-2`, {
 const cluster3 = new eks.Cluster(`${projectName}-3`, {
     vpcId: vpc.vpcId,
     publicSubnetIds: vpc.publicSubnetIds,
+    corednsAddonOptions: {
```
corymhall, line 42: We need to add this to the migration guide.

nodejs/eks/cluster.ts
```diff
@@ -627,6 +624,74 @@ export function createCore(
         },
     );
 
+    pulumi.output(args.fargate).apply((fargate) => {
```
flostadler, line 627: What do you think about lifting the addons out of the apply and always creating them?
If fargate is enabled we set the compute type config.

nodejs/eks/cluster.ts
```diff
@@ -627,6 +624,74 @@ export function createCore(
...
+            const kubeProxyAddon = new aws.eks.Addon(
+                `${name}-kube-proxy`,
+                {
+                    clusterName: eksCluster.name,
+                    addonName: "kube-proxy",
+                    resolveConflictsOnCreate:
+                        args.kubeProxyAddonOptions?.resolveConflictsOnCreate ?? "OVERWRITE",
+                    resolveConflictsOnUpdate:
+                        args.kubeProxyAddonOptions?.resolveConflictsOnUpdate ?? "OVERWRITE",
+                    addonVersion: kubeProxyVersion,
```
flostadler, line 651: we should set the `preserve` property to `true`. Otherwise this would delete the resources from the cluster when users toggle the `enabled` setting

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1798,6 +1812,108 @@ func generateSchema() schema.PackageSpec {
 					Required: []string{"content", "contentType"},
 				},
 			},
+			"eks:index:ResolveConflictsOnUpdate": {
```
flostadler, line 1815: Nice!

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1798,6 +1812,108 @@ func generateSchema() schema.PackageSpec {
...
+					},
+					{
+						Name:  "Overwrite",
+						Value: "OVERWRITE",
+						Description: "If the self-managed version of the add-on is installed on your cluster and the Amazon EKS default value is different than the existing value, " +
+							"Amazon EKS changes the value to the Amazon EKS default value.",
+					},
+				},
+			},
+			"eks:index:CoreDnsAddonOptions": {
```
flostadler, line 1859: We should add the `serviceAccountRoleArn` prop as well.

I'm also wondering if we should allow users to pass configuration values and then merge those with the ones the provider sets.

Modifying the resource requests/limits or replica count of coredns is a common config for example.

flostadler: But we can exclude that from this PR here. I've added it to the VPC CNI and if we both start adding those settings we're just gonna step on each others toes

nodejs/eks/cluster.ts
corymhall: That brings up a question - Is enabled a thing that only works on initial deployment?

provider/cmd/pulumi-gen-eks/main.go
corymhall: > We should add the serviceAccountRoleArn prop as well.

I don't think either `coredns` or `kube-proxy` use a role. At least according to [this doc](https://docs.aws.amazon.com/eks/latest/userguide/workloads-add-ons-available-eks.html#add-ons-coredns)

flostadler: Ah nice! Didn't know that

nodejs/eks/cluster.ts
flostadler: The addons can be created/deleted at any time

nodejs/eks/cluster.ts
corymhall: Done

nodejs/eks/cluster.ts
corymhall: Updated

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -663,6 +663,20 @@ func generateSchema() schema.PackageSpec {
 						Description: "The authentication mode of the cluster. Valid values are `CONFIG_MAP`, `API` or `API_AND_CONFIG_MAP`.\n\n" +
 							"See for more details:\nhttps://docs.aws.amazon.com/eks/latest/userguide/grant-k8s-access.html#set-cam",
 					},
+					"corednsAddonOptions": {
+						TypeSpec: schema.TypeSpec{
+							Plain: true,
```
t0yv0, line 668: Do you need Plain: true? I think what it does is make it slightly less nice for the user but supposedly easier for the implementation. Wanted to make sure it's an informed decision here.

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1798,6 +1812,108 @@ func generateSchema() schema.PackageSpec {
...
+						Description: "If the self-managed version of the add-on is installed on your cluster and the Amazon EKS default value is different than the existing value, " +
+							"Amazon EKS changes the value to the Amazon EKS default value.",
+					},
+				},
+			},
+			"eks:index:CoreDnsAddonOptions": {
+				ObjectTypeSpec: schema.ObjectTypeSpec{
+					Type: "object",
+					Properties: map[string]schema.PropertySpec{
+						"enabled": {
```
t0yv0, line 1863: The Go design would be `disabled` with default: false, so missing and false are the same. But perhaps we intentionally prefer the extra "enabled: true" marker.

nodejs/eks/cluster.ts
```diff
@@ -627,6 +624,76 @@ export function createCore(
...
+                      {
+                          addonName: "kube-proxy",
+                          kubernetesVersion: eksCluster.version,
+                          mostRecent: true, // whether to return the default version or the most recent version for the specified kubernetes version
+                      },
+                      { parent, provider },
+                  )
+                  .apply((addonVersion) => addonVersion.version);
+
+        const kubeProxyAddon = new aws.eks.Addon(
```
t0yv0, line 641: Conditional resource, interesting. Good thing args.kubeProxyAddonOptions?.enabled are not outputty and cannot be unknown.

nodejs/eks/cluster.ts
```diff
@@ -627,6 +624,76 @@ export function createCore(
...
+                resolveConflictsOnUpdate:
+                    args.corednsAddonOptions?.resolveConflictsOnUpdate ?? "OVERWRITE",
+                configurationValues: pulumi.output(args.fargate).apply((fargate) => {
+                    if (fargate) {
+                        return JSON.stringify({
+                            computeType: "Fargate",
+                        });
+                    }
+                    return undefined;
+                }) as any,
```
t0yv0, line 691: Nit: disabling type-checking via any here avoids some pain perhaps?

t0yv0:
Left some nits but looking very healthy overall! Reviewing at the low-level mostly.  Trusting this is the right default we want: "creating an eks.Cluster will now also create the coredns and kube-proxy addons"

nodejs/eks/cluster.ts
flostadler: `any` is gonna get removed here when we expose the `configurationValues`. That's a follow up item.

flostadler: https://github.com/pulumi/pulumi-eks/issues/1369

provider/cmd/pulumi-gen-eks/main.go
t0yv0: Ah yes, unsolved problem of conditional creates.

provider/cmd/pulumi-gen-eks/main.go
flostadler: Those resources are deployed by default on EKS

provider/cmd/pulumi-gen-eks/main.go
corymhall: Why does it make it less nice for the user? I thought you use plain when it doesn't need to accept output values. Is that not the case?

provider/cmd/pulumi-gen-eks/main.go
corymhall: Most people will be using it in TypeScript 😀

nodejs/eks/cluster.ts
corymhall: The referenced issue is https://github.com/pulumi/pulumi/issues/6175

flostadler (approved):
LGTM!

provider/cmd/pulumi-gen-eks/main.go
t0yv0: I'm thinking Go has good style in here :) But deferring to y'all.

provider/cmd/pulumi-gen-eks/main.go
t0yv0: Generally if I'm a user, I want to be able to call resources by passing `Output<X>` values directly. If I have plain values, great, I can lift them to outputs. If I have however computed Output values, and the resource does not accept them, I'm really stuck and I can only provision inside an apply which defeats plans. So generally Plain: true is bad for users. However the fact that this property is used for conditional resource creation inside is a good reason for having Plain: true.

provider/cmd/pulumi-gen-eks/main.go
flostadler: I just had a look through other settings in the provider and more are called `enable...` then `disable...`. Many of those are rooted in AWS settings, so I'd vote for `enabled` here to stay more consistent with what the provider is already doing

t0yv0: Yeah if this is what AWS users expect it's perfect.

Issue #1132, opened 2024-04-24
Use EKS Addon to manage VPC CNI instead of being self-managed


rquitales:
## Issue details

Currently, the EKS provider maintains a hardcoded copy of v1.16.0 of the VPC CNI manifest that is installed on all EKS clusters created/managed by this provider. This could lead to a version compatibility mismatch between the CNI and EKS cluster version. We should migrate to using the underlying aws.eks.Addon resource to manage the VPC CNI manifests on cluster. This will also remove the dependency on shelling out to `kubectl`.

This will also help mitigate #669, as the error occurs when we attempt to communicate directly with the Kubernetes API server.

This is planned feature work for the next major version release.

### Affected area/feature

eks.Cluster and eks.VpcCni

flostadler:
Switching out the existing VPC CNI with the addon should be disruption free because the provider doesn't implement a delete operation for the VPC CNI.
EKS Addons adopt resources, so this should be transparent.

In terms of configuration options, almost all existing config options of the VPC CNI translate to configuration options of the addon. The exceptions are `securityContext` and `image` properties.

Those should work with the patch resources leveraging Kubernetes Server Side Apply. Currently building a prototype to confirm this.

flostadler:
This is merged into the release branch now

PR #1358, opened 2024-09-13
Move VPC CNI to EKS addon

flostadler:
### Proposed changes

Currently, the EKS provider maintains a hardcoded copy of v1.16.0 of the VPC CNI manifest that is installed on all EKS clusters created/managed by this provider. This could lead to a version compatibility mismatch between the CNI and EKS cluster version. We should migrate to using the underlying aws.eks.Addon resource to manage the VPC CNI manifests on cluster.

Switching out the existing `VpcCni` component with the addon is disruption free because the provider doesn't implement a delete operation for it. The EKS addon adopts the existing on cluster resources.

Almost all current configuration options exist on the addon. The exceptions are:
- `enableIpv6`: This is done automatically by the addon depending on whether the cluster is in IPv4 or IPv6 mode
- container images: The addon now automatically selects the right images depending on the AWS region. This was the reason why the images were made configurable. (see https://github.com/pulumi/pulumi-eks/issues/1078)
- `securityContextPrivileged`

`enableIpv6` and container images can be dropped now, but for `securityContextPrivileged` we can use a k8s patch resource. The patch resource will take ownership of the fields it's managing and the addon will manage the others.
We cannot take this same approach for the images because the addon will take ownership of those fields. This would lead to errors when updating the addon.

The dropped arguments will need to be documented in the migration guide.

### How to review?
1. have a look at the schema changes
2. review the provider code changes, the main changes are in `nodejs/eks/cni-addon.ts`
3. check the tests

### Related issues (optional)

Closes https://github.com/pulumi/pulumi-eks/issues/1132
Closes https://github.com/pulumi/pulumi-eks/issues/1254
Relates to https://github.com/pulumi/pulumi-eks/issues/1078


nodejs/eks/cni-addon.ts
```diff
@@ -0,0 +1,487 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
+import * as k8s from "@pulumi/kubernetes";
+
+export interface VpcCniAddonOptions {
```
flostadler, line 19: That's a 1:1 copy of the existing VpcCni addons except for the `enableIPv6` setting that is not automatically handled by the addon.

nodejs/eks/cni-addon.ts
```diff
@@ -0,0 +1,487 @@
...
+                        },
+                    },
+                },
+            },
+            { parent: this, dependsOn: [addon] },
+        );
+    }
+}
+
+function computeEnv(args: VpcCniAddonOptions): {
```
flostadler, line 328: Also a 1:1 copy minus the ipv6 setting

examples/examples_nodejs_test.go
```diff
@@ -674,52 +642,75 @@ func getJSBaseOptions(t *testing.T) integration.ProgramTestOptions {
 
 // TestAccCNIAcrossUpdates tests that the CNI manifest is reapplied when the EKS provider changes its base manifest.
 func TestAccCNIAcrossUpdates(t *testing.T) {
-	t.Log("Running `pulumi up` with v2.1.0 of the EKS provider")
```
flostadler, line 677: This test was broken before. It didn't actually test the upgrade, but just deployed two different clusters...

nodejs/eks/cni-addon.ts
```diff
@@ -0,0 +1,545 @@
...
+
+    /**
+     * The Kubernetes version of the cluster. This is used to determine the addon version to use if `addonVersion` is not specified.
+     */
+    clusterVersion?: pulumi.Input<string>;
+
+    /**
+     * How to resolve field value conflicts when migrating a self-managed add-on to an Amazon EKS add-on. Valid values are `NONE` and `OVERWRITE`. For more details see the [CreateAddon](https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateAddon.html) API Docs.
+     */
+    resolveConflictsOnCreate?: pulumi.Input<string>;
```
flostadler, line 31: Cory is adding enums for the conflict resolution in this PR: https://github.com/pulumi/pulumi-eks/pull/1357

I'm gonna update those params once that's merged in

nodejs/eks/cni-addon.ts
```diff
@@ -0,0 +1,545 @@
...
+     * Key-value map of resource tags. If configured with a provider `defaultTags` configuration block present, tags with matching keys will overwrite those defined at the provider-level.
+     */
+    tags?: pulumi.Input<{
+        [key: string]: pulumi.Input<string>;
+    }>;
+
+    /**
+     * Custom configuration values for the vpc-cni addon. This object must match the schema derived from [describe-addon-configuration](https://docs.aws.amazon.com/cli/latest/reference/eks/describe-addon-configuration.html).
+     */
+    configurationValues?: pulumi.Input<object>;
```
flostadler, line 63: AWS offers schemas (jsonschema) for the addon configuration. But those can differ per addon version (and can include breaking changes).

I'm thinking that we could generate the types for this, but that requires a versioning strategy (forwards & backwards compatibility) and automation. The generation itself  should be doable, we're already doing that in aws-native.

I'd propose that we punt this for a later time and gauge user interest in the meantime.

flostadler: And this would be better placed in aws classic IMHO. That way the whole user base profits from it and not only pu-eks users

nodejs/eks/tsconfig.json
```diff
@@ -15,32 +15,6 @@
...
-        "cmd/provider/randomSuffix.ts",
-        "cmd/provider/securitygroup.ts",
-
-        "ami.test.ts",
-        "authenticationMode.test.ts",
-        "dependencies.test.ts",
-        "nodegroup.test.ts",
-        "userdata.test.ts"
-    ]
+    "include": ["./**/*.ts"],
```
flostadler, line 18: I got tired of manually managing the files...

nodejs/eks/cni-addon.ts
flostadler: We had a chat about this, opening an issue in aws classic for it

nodejs/eks/cni-addon.ts
flostadler: Tracker for tackling the follow up: https://github.com/pulumi/pulumi-eks/issues/1369

nodejs/eks/cni-addon.ts
t0yv0: Cool idea but I'd defer in an issue to get this off the critical path.

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1063,17 +1063,13 @@ func generateSchema() schema.PackageSpec {
 					"eksCluster",
 				},
 			},
-			// The TypeScript library exports this at the top-level index.ts, so we expose it here. Although, it's not
-			// super useful on its own, so we could consider *not* exposing it to the other languages.
-			// Note that this is a _custom_ resource (not a component), implemented in the provider plugin.
-			// Previously it was implemented as a dynamic provider.
-			"eks:index:VpcCni": {
+			"eks:index:VpcCniAddon": {
```
t0yv0, line 1066: So this https://www.pulumi.com/registry/packages/eks/api-docs/vpccni/ is removed in favor of a new one it seems, ack.

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1568,7 +1564,7 @@ func generateSchema() schema.PackageSpec {
 				ObjectTypeSpec: schema.ObjectTypeSpec{
 					Type:        "object",
 					Description: "Describes the configuration options available for the Amazon VPC CNI plugin for Kubernetes.",
-					Properties:  vpcCniProperties(false /*kubeconfig*/),
+					Properties:  vpcCniProperties(true /*cluster*/),
```
t0yv0, line 1567: Nit: I have boolean blindness and prefer enums. But it's a very small matter.

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -2290,12 +2265,73 @@ func vpcCniProperties(kubeconfig bool) map[string]schema.PropertySpec {
...
+		"configurationValues": {
+			TypeSpec: schema.TypeSpec{
+				Type:                 "object",
+				AdditionalProperties: &schema.TypeSpec{Ref: "pulumi.json#/Any"},
+			},
+			Description: "Custom configuration values for the vpc-cni addon. This object must match the schema derived from " +
+				"[describe-addon-configuration]" +
+				"(https://docs.aws.amazon.com/cli/latest/reference/eks/describe-addon-configuration.html).",
+		},
+		"resolveConflictsOnCreate": {
```
t0yv0, line 2277: In a different PR I saw today, something had an enumeration for these, is this coming? Enumerations are nice. Thanks!

nodejs/eks/cmd/provider/cni-addon.ts
```diff
@@ -0,0 +1,43 @@
...
+import * as pulumi from "@pulumi/pulumi";
+import { VpcCniAddon } from "../../cni-addon";
+
+const cniAddonProvider: pulumi.provider.Provider = {
+    construct: (
+        name: string,
+        type: string,
+        inputs: pulumi.Inputs,
+        options: pulumi.ComponentResourceOptions,
+    ) => {
```
t0yv0, line 24: Nit: signature does not specify return type here, perhaps could help? Or use an async method?

provider/cmd/pulumi-gen-eks/main.go
flostadler: Yeah, I referenced this in a comment above: https://github.com/pulumi/pulumi-eks/pull/1358#discussion_r1764974168

There's some cleanup left to align Cory's and my PR once they're merged

nodejs/eks/cmd/provider/cni-addon.ts
```diff
@@ -0,0 +1,43 @@
...
+        try {
+            const addon = new VpcCniAddon(name, <any>inputs, options);
```
t0yv0, line 26: Hm curious, I wonder if `<any>inputs` cast is the recommended way to develop these, or is there any hydration step? Perhaps it's all been handled by the framework.

nodejs/eks/cni-addon.ts
```diff
@@ -0,0 +1,545 @@
...
+        if (args.securityContextPrivileged) {
+            containers.push({
+                name: "aws-node",
+                securityContext: {
+                    privileged: args.securityContextPrivileged,
+                },
+            });
+        }
+
+        return new k8s.apps.v1.DaemonSetPatch(
```
t0yv0, line 359: Interesting.

nodejs/eks/cmd/provider/cni-addon.ts
flostadler: The inputs are untyped, so we have to accept a cast somewhere I think.

nodejs/eks/cmd/provider/cni-addon.ts
flostadler: In the long run we should do what awsx does here with generated ts types IMO

examples/utils/utils.go
```diff
@@ -802,3 +882,92 @@ func EnsureKubeconfigFails(t *testing.T, kubeconfig any) {
...
+	}
+
+	if addon == nil {
+		return "", fmt.Errorf("unable to find most recent version of addon %s", addonName)
+	}
+
+	return addon.AddonVersion, nil
+}
+
+func findAddonVersion(eksClient *eks.Client, addonName string, selector func (currentAddon *addonInfo, versionInfo eksTypes.AddonVersionInfo, compatibility eksTypes.Compatibility) *addonInfo) (*addonInfo, error) {
```
t0yv0, line 936: Nice!

examples/examples_nodejs_test.go
```diff
@@ -22,23 +22,26 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"testing"
 	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
```
t0yv0, line 31: Huh

examples/tests/cluster-addons/step2/index.ts
```diff
@@ -0,0 +1,55 @@
...
+      }
+    },
+  },
+});
+
+// Export the cluster's kubeconfig.
+export const kubeconfig = cluster.kubeconfig;
+export const clusterName = cluster.core.cluster.name;
+
+const ng = eks.createManagedNodeGroup("cluster-addons", {
```
t0yv0, line 44: Out of scope of this PR - but remind me do we like this API that creates an aws.eks.NodeGroup instance vs some form of API that's a wrapper Component resource e.g. `new eks.ManagedNodeGroup()`

t0yv0 (approved):
Looking thorough.

nodejs/eks/cmd/provider/cni-addon.ts
t0yv0: Indeed, indeed out of scope of the PR.

nodejs/eks/cni-addon.ts
```diff
@@ -0,0 +1,545 @@
...
+     */
+    securityContextPrivileged?: pulumi.Input<boolean>;
+}
+
+/**
+ * Addon manages an EKS add-on.
+ * For more information about supported add-ons, see: https://docs.aws.amazon.com/eks/latest/userguide/eks-add-ons.html
+ */
+export class VpcCniAddon extends pulumi.ComponentResource {
+    public readonly addon: pulumi.Input<aws.eks.Addon>;
```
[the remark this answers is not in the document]
flostadler: Oh good catch, that was definitely not intended!
It should've been an output

examples/tests/cluster-addons/step2/index.ts
flostadler: I'm indifferent about those methods tbh. The one shortcoming of them is that they're not documented in the registry (which we could do if we really wanted to).

examples/examples_nodejs_test.go
flostadler: helps passing pointers of constants like `true`

examples/tests/cluster-addons/step2/index.ts
t0yv0: Yeah definitely more of a style/consistency issue than anything.

Issue #563, opened 2021-04-18
IAM roles not getting tags propagated


jaxxstorm:
Since we updated to multi lang, it seems all IAM roles and instance profiles are not getting the tags propagated down to them.

I found this when defining an EKS cluster, and overnight the cleanup lambda function deleted the instance profile because no `Owner` tag was set. As a result, the cluster became unhealthy. 

## Expected behaviour
Tags set on the EKS component should be pushed down to the IAM roles, instance profile etc

## Current behavior
It seems all IAM sub-components are not getting tags set on them.,

flostadler:
Seems like we're not setting the tags on the IAM roles the component creates: https://github.com/pulumi/pulumi-eks/blob/a5c6e7a18c067b5bb7d823634fb18515718f46f8/nodejs/eks/servicerole.ts#L77-L84

PR #1359, opened 2024-09-13
Pass tags to IAM roles

flostadler:
### Proposed changes

Tags weren't propagated to the IAM roles the provider creates. This passes them down to the role

### Related issues (optional)

Fixes https://github.com/pulumi/pulumi-eks/issues/563


t0yv0 (approved):
LGTM if we believe this does not cause replacement.

flostadler:
> LGTM if we believe this does not cause replacement.

Nope, it doesn't (and the upgrade tests would flag that)

Issue #535, opened 2021-02-08
Pulumi force initiates RollingUpdatewhile updating ASG's launch configurations


abaskar-tableau:
### Problem description
Whenever ASG's launch configurations are updated by pulumi, as part of that CFN nodes stacks update, Pulumi’s generated CFN template includes a rolling update configuration, without any explicit configuration from our end. Check “AutoScalingRollingUpdate” in this doc for CFN’s update policy : [AutoScalingRollingUpdate](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html)
Below is the CFN template Pulumi generates as part of the above list of CFN updates. As we can see, there is a “AutoScalingRollingUpdate” section in the template that drives CFN to initiate termination of nodes with older AMI. We would like this to be configurable, since we don't want the old nodes to be terminated before the pods are drained from them.

Here is the hard coded Rolling Update CFN template on Pulumi's end : [CFN template](https://github.com/pulumi/pulumi-eks/blob/b78a28d455b866d2d671d6153774bcb5ca70bb32/nodejs/eks/nodegroup.ts#L541-L568)

### Errors & Logs

              AWSTemplateFormatVersion: '2010-09-09'
                Outputs:
                    NodeGroup:
                        Value: !Ref NodeGroup
                Resources:
                    NodeGroup:
                        Type: AWS::AutoScaling::AutoScalingGroup
                        Properties:
                          DesiredCapacity: 5
                          LaunchConfigurationName: online-cd-main-default-worker-node-group-0-nodeLaunchConfiguration-5e7f9ff
                          MinSize: 1
                          MaxSize: 25
                          VPCZoneIdentifier: ["subnet-02f0bf13256db0423"]
                          Tags:
                          - Key: Name
                            Value: online-cd-main-eksCluster-68f6e20-worker
                            PropagateAtLaunch: 'true'
                          - Key: kubernetes.io/cluster/online-cd-main-eksCluster-68f6e20
                            Value: owned
                            PropagateAtLaunch: 'true'
                          - Key: k8s.io/cluster-autoscaler/enabled
                            Value: true
                            PropagateAtLaunch: 'true'
                          - Key: k8s.io/cluster-autoscaler/online-cd-main-eksCluster-68f6e20
                            Value: true
                            PropagateAtLaunch: 'true'
                        UpdatePolicy:
                          AutoScalingRollingUpdate:
                            MinInstancesInService: '1'
[...]

PR #1373, opened 2024-09-20
Deprecate NodeGroup component

### LINKED ISSUE #1352: Deprecate `NodeGroup`
The `NodeGroup` component uses the deprecated AWS Launch Configuration ([see](https://docs.aws.amazon.com/autoscaling/ec2/userguide/launch-configurations.html))

Mark the legacy CloudFormation based self-managed NodeGroup (also referred to as NodeGroupV1) as deprecated. The Pulumi native NodeGroupV2 is functionally equivalent (same inputs) but doesn't suffer from problems like [pulumi-eks#535](https://github.com/pulumi/pulumi-eks/issues/535). Users will need to replace their self managed node groups anyway to migrate away from AL2 in a safe way (see [What does a node group update look like for users?](https://docs.google.com/document/d/1XyLq_EyAziCp3f6rQ_8qfcUk0RMl8AgUq1mqpqsQcHM/edit#bookmark=id.l6qozya46ay3)).

### LINKED ISSUE #1353: Update cluster default node group to use a ManagedNodeGroup
Right now the clusters still use the outdated `NodeGroup`(aka NodegroupV1). Align the `ManagedNodeGroup` with `NodeGroup` and make it functionally equivalent. The two mostly provide the same configuration parameters already, just different names. What's currently missing is the ability to specify node volume configuration.

flostadler:
The `NodeGroup` component uses the deprecated AWS Launch Configuration ([see](https://docs.aws.amazon.com/autoscaling/ec2/userguide/launch-configurations.html))

This marks the legacy CloudFormation based self-managed NodeGroup (also referred to as NodeGroupV1) as deprecated. The Pulumi native NodeGroupV2 is functionally equivalent (same inputs) but doesn't suffer from problems like [pulumi-eks#535](https://github.com/pulumi/pulumi-eks/issues/535). Users will need to replace their self managed node groups anyway to migrate away from AL2 in a safe way (see [What does a node group update look like for users?](https://docs.google.com/document/d/1XyLq_EyAziCp3f6rQ_8qfcUk0RMl8AgUq1mqpqsQcHM/edit#bookmark=id.l6qozya46ay3)).

This also switches the default node group of the cluster component to use NodeGroupV2 instead.

Closes #1353 
Closes #1352

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -35,9 +35,11 @@ import { assertCompatibleAWSCLIExists } from "./dependencies";
 import {
     computeWorkerSubnets,
     createNodeGroup,
+    createNodeGroupV2,
     NodeGroup,
     NodeGroupBaseOptions,
     NodeGroupData,
+    NodeGroupV2Data,
 } from "./nodegroup";
 import { createNodeGroupSecurityGroup } from "./securitygroup";
 import { ServiceRole } from "./servicerole";
@@ -1896,7 +1898,7 @@ export class Cluster extends pulumi.ComponentResource {
     /**
      * The default Node Group configuration, or undefined if `skipDefaultNodeGroup` was specified.
      */
-    public readonly defaultNodeGroup: NodeGroupData | undefined;
+    public readonly defaultNodeGroup: NodeGroupV2Data | undefined;
 
     /**
      * The EKS cluster.
@@ -2009,7 +2011,7 @@ export interface ClusterResult {
     instanceRoles: pulumi.Output<aws.iam.Role[]>;
     nodeSecurityGroup: aws.ec2.SecurityGroup;
     eksClusterIngressRule: aws.ec2.SecurityGroupRule;
-    defaultNodeGroup: NodeGroupData | undefined;
+    defaultNodeGroup: NodeGroupV2Data | undefined;
     eksCluster: aws.eks.Cluster;
     core: CoreData;
 }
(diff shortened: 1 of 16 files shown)


nodejs/eks/cluster.ts
```diff
@@ -2069,26 +2071,19 @@ export function createCluster(
...
             },
             self,
         );
-        if (defaultNodeGroup.cfnStack) {
-            configDeps.push(defaultNodeGroup.cfnStack.id);
-        }
     }
 
-    // Export the cluster's kubeconfig with a dependency upon the cluster's autoscaling group. This will help
-    // ensure that the cluster's consumers do not attempt to use the cluster until its workers are attached.
```
flostadler, line 2089: This was only needed for the nodejs SDK. MLCs handle that correctly

t0yv0:
Trusting migration guide covers how the users that are currently on this can move forward? Could link to it in the deprecation notice.

provider/cmd/pulumi-resource-eks/schema.json
```diff
@@ -292,10 +292,21 @@
                     "plain": true,
                     "description": "Custom k8s node labels to be attached to each worker node. Adds the given key/value pairs to the `--node-labels` kubelet argument."
                 },
+                "launchTemplateTagSpecifications": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "/aws/v6.18.2/schema.json#/types/aws:ec2%2FLaunchTemplateTagSpecification:LaunchTemplateTagSpecification"
```
t0yv0, line 298: Do we have tooling that updates this reference over time? v6.18.2 I mean.

t0yv0, line 298: Do we have tooling that updates this reference over time? v6.18.2 I mean.

t0yv0 (approved):

provider/cmd/pulumi-resource-eks/schema.json
flostadler: Right now the aws provider version is centrally managed here: https://github.com/pulumi/pulumi-eks/blob/d4f539f99700f3cd084da85f3dad59ec018a2dc0/provider/cmd/pulumi-gen-eks/main.go#L82

But updating the aws version is a manual process. I'm gonna create an issue for automating that.

flostadler:
> Trusting migration guide covers how the users that are currently on this can move forward? Could link to it in the deprecation notice. 

Yeah there will be a whole section on this. Right now there's a bit of a chicken&egg problem, we'll first need to publish the migration guide before we can reference it here

t0yv0:
So long as there's a reminder somewhere in the epic we're good.

PR #1379, opened 2024-09-23
Make ManagedNodeGroups track the cluster version

### LINKED ISSUE #1253: Auto update Managed Node Groups when updating EKS cluster version
## Hello!

## Issue details

Right now, the Managed Node Groups do not take the clusters version into account. Instead, they're configured without a version unless users provide it.

When configuring Managed Node Groups without a version, they default to using the current cluster version at deploy time. Changes to the clusters version do not propagate to the node group in this case.
This forces users to manually manage the versions of their Managed Node Groups.

Ideally the Managed Node Groups should always follow the cluster version when no fixed version is specified by users.

As a workaround you can set the Managed Node Group version to the version output of the EKS cluster like so:
```ts
const cluster = new eks.Cluster("example-managed-nodegroups", {
  skipDefaultNodeGroup: true,
  deployDashboard: false,
  vpcId: eksVpc.vpcId,
  publicSubnetIds: eksVpc.publicSubnetIds,
  privateSubnetIds: eksVpc.privateSubnetIds,
  instanceRoles: [role],
  version: "1.29",
});

const managedNodeGroup0 = eks.createManagedNodeGroup("example-managed-ng0", {
  cluster: cluster,
  nodeRole: role,
  enableIMDSv2: true,
  version: cluster.eksCluster.version, // <- this makes the version propagate to the node group
});
```

flostadler:
When configuring Managed Node Groups without a version, they default to using the
current cluster version at deploy time. Changes to the clusters version do not
propagate to the node group in this case.
EKS managed node groups only support 1 minor version skew between the control plane and data plane.
Otherwise cluster upgrades will fail.

This change makes the ManagedNodeGroups track the cluster version unless a fixed version is provided by users.

Fixes #1253


examples/tests/cluster-upgrade/step2/index.ts
```diff
@@ -0,0 +1,48 @@
+
+import * as pulumi from "@pulumi/pulumi";
+import * as awsx from "@pulumi/awsx";
+import * as eks from "@pulumi/eks";
+import * as iam from "./iam";
+
+const config = new pulumi.Config();
+const clusterVersion = config.require("upgradeVersion");
```
t0yv0, line 8: I wish ProgramTest steps could just indicate different config values to avoid dup'ing source.

examples/examples_nodejs_test.go
```diff
@@ -995,3 +995,61 @@ func TestAccClusterAddons(t *testing.T) {
...
+		t.Skip("skipping test in short mode.")
+	}
+
+	eksClient := createEksClient(t)
+	k8sVersions, err := utils.ListKubernetesVersions(eksClient)
+	require.NoError(t, err)
+	require.GreaterOrEqual(t, len(k8sVersions), 2, "expected at least 2 k8s versions to be available")
+
+	initialVersion := k8sVersions[1]
+	upgradeVersion := k8sVersions[0]
```
t0yv0, line 1010: nit: log this to `t.Logf`

t0yv0 (approved):
LGTM

Issue #1293, opened 2024-08-09
Allow ignoring changes to `desiredSize` of Node Groups for Cluster Autoscaler support


JustASquid:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

When using [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) for automatically scaling node groups based on cluster requirements pulumi permanently shows a diff.

The problem is because the cluster autoscaler takes control of the `desiredSize` of the `scalingConfig`, putting it out of sync with Pulumi state. 
In pulumi, the recommended way to handle a case like this is to have Pulumi give up control over the specified property with the `ignoreChanges` option.

However, currently in Pulumi, `ignoreChanges` does nothing when given to a `ComponentResource`. See [docs](https://www.pulumi.com/docs/concepts/options/ignorechanges/).

My suggestion for a simple and easy fix would be to pass `ignoreChanges` passed to the resource down to the options of the internally created `aws.eks.NodeGroup` or ASG when using the `NodeGroup`/`NodeGroupV2` component.

flostadler:
Hey @JustASquid,
this sounds like a good enhancement to the node groups! I can see us adding a new parameter (e.g. `ignoreDesiredSize`) to the `scalingConfig`.

In the meantime you can achieve this using transforms as a workaround. For example using typescript:
```
const mng = new eks.ManagedNodeGroup("example-ng", {
    cluster: cluster,
    nodeGroupName: "aws-managed-ng",
    nodeRoleArn: role.arn,
    scalingConfig: {
        desiredSize: 4,
        maxSize: 10,
        minSize: 1,
    }
}, { transforms: [args => {
    if (args.type === "aws:eks/nodeGroup:NodeGroup") {
        return {
            props: args.props,
            opts: pulumi.mergeOptions(args.opts, { ignoreChanges: ["scalingConfig.desiredSize"] })
        }
    }
    return undefined;
}]});
```

flostadler:
This is done now and will be released as part of EKS v3

PR #1380, opened 2024-09-23
Add the ability to ignore changes to desired size of node groups

### LINKED ISSUE #985: Ability to Ignore Changes on `desiredSize` in eks.ManagedNodeGroup
## Hello!

### Context
Currently, any change to the `desiredSize` of a node group leads to updates within the EKS ManagedNodeGroup and the child NodeGroup, which can be disruptive and undesired in certain scenarios. For example, when scaling operations are managed by external tools or autoscaling policies, the Pulumi should not interfere with these dynamic changes.

### Affected area/feature

eks.ManagedNodeGroup in AWS EKS

### Expected Behavior
Introduce a configuration option, possibly a boolean flag, that allows the `desiredSize` of a node group to be ignored during updates. This would ensure that changes to this specific property do not trigger updates in the EKS ManagedNodeGroup, providing greater flexibility in managing node group sizes.

### Impact
This feature would greatly enhance the usability of eks.ManagedNodeGroup for scenarios involving dynamic scaling, making it more adaptable to various operational requirements.

flostadler:
When using [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) for automatically scaling node groups based on cluster requirements pulumi permanently shows a diff.

The problem is because the cluster autoscaler takes control of the desiredSize of the scalingConfig, putting it out of sync with Pulumi state.

MLCs don't support ignore changes for their children resources, because of this we're adding a new input that allows you to selectively ignore scaling changes.
This is not added to the deprecated `NodeGroup` (aka `NodeGroupV1`) because the desired size is part of the cloudformation stack that is deployed as part of that component. We cannot selectively ignore changes to the cloudformation template because its a string.

Closes #985 #1293


provider/test-programs/ignore-scaling-changes/index.ts
```diff
@@ -0,0 +1,64 @@
...
+export const kubeconfig = cluster.kubeconfig;
+
+// Create a simple AL 2023 node group with x64 instances
+const mng = eks.createManagedNodeGroup("ignore-scaling-config-mng", {
+  scalingConfig: {
+    minSize: 1,
+    maxSize: 3,
+    desiredSize,
+  },
+  ignoreScalingChanges: true,
```
t0yv0, line 48: From which parameters here does it follow that this resource is using "cluster autoscaler"? If so could ignoreScalingChanges: true be inferred automatically?

provider/test-programs/ignore-scaling-changes/index.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const ngV2 = new eks.NodeGroupV2("ignore-scaling-config-ngv2", {
+  minSize: 1,
+  maxSize: 3,
+  desiredCapacity: desiredSize,
+  cluster: cluster,
+  instanceType: "t3.medium",
+  operatingSystem: eks.OperatingSystem.AL2023,
+  instanceProfile: new aws.iam.InstanceProfile("ignore-scaling-config-ngv2", {role: ngV2Role}),
+  ignoreScalingChanges: true,
```
t0yv0, line 63: Same q here.

t0yv0 (approved):
Could also support this by directing users to the transforms resource option, but it's a little more involved so this seems nice.   Left some questions on opt-in vs auto-infer.

provider/test-programs/ignore-scaling-changes/index.ts
flostadler: Cluster Autoscaler is not an EKS feature, it's software users can deploy in their clusters.

flostadler:
> Could also support this by directing users to the transforms resource option, but it's a little more involved so this seems nice.   Left some questions on opt-in vs auto-infer. 

I think we shouldn't require users to meddle with transforms for common use-cases. Cluster autoscaling is one of them

PR #1388, opened 2024-09-25
Skip upgrade tests until the first v3 alpha is released

flostadler:
The release branch for EKS v3 contains breaking changes now. This causes the upgrade tests to fail. Because of that we disabled the upgrade tests.
Once the first alpha version was released we can re-record and re-enable them.

### DIFF
--- a/provider/provider_test.go
+++ b/provider/provider_test.go
@@ -26,7 +26,7 @@ const (
 )
 
 func TestExamplesUpgrades(t *testing.T) {
-	t.Skip("Skipping for now until EKS v3 alpha is released. Otherwise we cannot re-record.")
+	t.Skip("Temporarily skipping upgrade tests. Needs to be addressed in pulumi/pulumi-eks#1387")
 	t.Run("cluster", func(t *testing.T) {
 		testProviderUpgrade(t, "cluster")
 	})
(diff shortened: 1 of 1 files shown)


corymhall (approved):

t0yv0:
I think in @pulumi/providers went stricter on the recent Azure Classic major release and actually tried to test some of the upgrade scenarios (involving code changes) by actual tests. We're still finding what's the right balance here but getting those tests in is a significant lift we didn't plan into the EKS release upfront. I think perhaps reasonable to just ride on alpha testing here since eks.Cluster is a highly custom and stateful resource and upgrading is a high-touch activity for end-users that is likely not going to be meaningfully covered by our upgrade tests anyway. Copying @mjeffryes  for second opinion.

t0yv0 (approved):

t0yv0:
Please merge we can revisit ex-post-facto if there is pushback.

mjeffryes:
I do want us to test the upgrade path. Whether we do that manually or with upgrade tests can be a pragmatic decision based on what gives us confidence with least effort. The test provide a nice specification for how we actually expect it to work for users, but they are a bit of a pain to maintain. At any rate, no objections to turning them off for now in this branch.

PR #1393, opened 2024-09-25
Allow testing release.yml changes on pre-releases before merging

t0yv0:
### Proposed changes

Similar to how master.yml works, this change allows to trigger release.yml manually and point it to a WIP branch. This will help us out to test alpha releases and changes to the release workflow for the 3.x.x series.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,7 @@ env:
   DOTNETVERSION: "6.x"
   PYTHONVERSION: "3.8"
   JAVAVERSION: "11"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -807,3 +807,4 @@ name: release
   push:
     tags:
       - v*.*.*
+  workflow_dispatch: {}
```
flostadler, line 810: Wouldn't this actually publish docs now and release notes?
`IS_PRERELEASE: ${{ contains(github.ref_name,'-') }}` I'm not sure what this resolves to for an manual trigger.

corymhall (approved):

PR #1399, opened 2024-09-26
Update go sdk v3

t0yv0:
Go SDK must be versioned as v3 otherwise Go refuses to use this.


sdk/go.mod
```diff
@@ -1,10 +1,11 @@
-module github.com/pulumi/pulumi-eks/sdk/v2
+module github.com/pulumi/pulumi-eks/sdk/v3
 
 go 1.21
 
 require (
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/pulumi/pulumi-aws/sdk/v6 v6.29.0
+	github.com/pulumi/pulumi-eks/sdk/v2 v2.8.1
```
t0yv0, line 8: This looketh wrong.

flostadler, line 8: why are we importing the old sdk here?

t0yv0: because we generate code incorrectly

flostadler (approved):

t0yv0:
Version computation is not right per  https://github.com/pulumi/provider-version-action/issues/13

t0yv0:
Tagging this PR as needs-release/major has resulted in `PROVIDER_VERSION: 3.0.0-alpha.1727387122+135269b` being computed which should pass the previously failing check I think.

t0yv0:
A tad better but tests need to be updated to point to V3 now.

examples/cluster-go/step1/go.mod
```diff
@@ -1,94 +1,99 @@
...
 
 require (
 	github.com/pulumi/pulumi-aws/sdk/v6 v6.50.1
-	github.com/pulumi/pulumi-eks/sdk/v2 v2.0.0
-	github.com/pulumi/pulumi/sdk/v3 v3.93.0
+	github.com/pulumi/pulumi-eks/sdk/v3 v3.0.0-alpha.6
+	github.com/pulumi/pulumi/sdk/v3 v3.129.0
 )
 
+replace github.com/pulumi/pulumi-eks/sdk/v3 => ../../../sdk
```
t0yv0, line 13: It would be good to avoid keeping the replaces but I can't seem to be able to run `go mod tidy` since there is no valid v3 SDK published yet. Trying replaces temporarily.

t0yv0:
Go tests actually passed against the V3 SDK, that's good.

t0yv0:
I'm going to merge and cut another alpha.

Issue #1402, opened 2024-09-27
java.lang.IllegalArgumentException: '@Import com.pulumi.aws.eks.ClusterArgs.roleArn' is required but was not given a value


t0yv0:
### What happened?

Getting this problem on pulumi up on the 3.x testing:

```
java.lang.IllegalArgumentException: '@Import com.pulumi.aws.eks.ClusterArgs.roleArn' is required but was not given a value 
```

With the latest alpha:

```
    <dependencies>
        <dependency>
            <groupId>com.pulumi</groupId>
            <artifactId>pulumi</artifactId>
            <version>(,1.0]</version>
        </dependency>
        <dependency>
            <groupId>com.pulumi</groupId>
            <artifactId>aws</artifactId>
            <version>(6.0.2,6.99]</version>
        </dependency>
        <dependency>
            <groupId>com.pulumi</groupId>
            <artifactId>eks</artifactId>
            <version>3.0.0-alpha.7</version>
        </dependency>
    </dependencies>
```

This is unexpected, ClusterArgs does not even have roleArn. Wondering if something in Java SDK generation is going awry here. 

### Example

```java

package myproject;

import com.pulumi.Pulumi;
import com.pulumi.core.Output;
import com.pulumi.aws.s3.Bucket;
import com.pulumi.aws.eks.Cluster;

public class App {
    public static void main(String[] args) {
        Pulumi.run(ctx -> {
            var c = new Cluster("my-cluster");
            ctx.export("cid", c.id());
        });
    }
}
```

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
javagen version is behind, v0.9.8 vs 0.16.1, we need to automate having this up-to-date, likely fixing the issue here.

flostadler:
`rolearn` is required for the underlying `aws.eks.Cluster`. The error is about that. See: https://www.pulumi.com/registry/packages/aws/api-docs/eks/cluster/#rolearn_java

flostadler:
Fixed it with [Fix output of default node group referring to autoscalingGroupName #1410](https://github.com/pulumi/pulumi-eks/pull/1410). Release is currently running

PR #1408, opened 2024-09-30
Upgrade javagen to v0.16.1

flostadler:
Upgrades javagen to v0.16.1. This might solve the problems in #1402


t0yv0 (approved):
Interesting that .NET got edited, but LGTM

flostadler:
> Interesting that .NET got edited, but LGTM

That's the version string, not quite sure why that one didn't get updated when preparing for the alpha release. But it's expected that this changes to v3

PR #1410, opened 2024-09-30
Fix output of default node group referring to autoscalingGroupName

flostadler:
In https://github.com/pulumi/pulumi-eks/pull/1373 the default node group was updated to use the `NodeGroupV2` component. We missed changing the `NodeGroupData` type to reflect this. It was still referring to a property called `autoScalingGroupName`, but it should've been changed to expose an `autoScalingGroup`.

Fixes https://github.com/pulumi/pulumi-eks/issues/1402


t0yv0 (approved):

t0yv0:
Tests unhappy for some reason.

flostadler:
> Tests unhappy for some reason.

Yeah, I miss-typed an output in the example. It's fixed now!

Issue #1176, opened 2024-05-31
`instanceRole` still created when `skipDefaultNodeGroup` is enabled.


ringods:
### What happened?

Customer reports he doesn't pass an `instanceRole` or `instanceRoles` because he enabled `skipDefaultNodeGroup` (actually the `skip_default_node_group` in the Python SDK). They don't want our code to still create an instance role.

### Example

Role created:

https://github.com/pulumi/pulumi-eks/blob/161e7af6803b58fea43ffdc3c6f9ee85cf658134/nodejs/eks/cluster.ts#L782-L802

Profile skipped due to `skipDefaultNodeGroup`:

https://github.com/pulumi/pulumi-eks/blob/161e7af6803b58fea43ffdc3c6f9ee85cf658134/nodejs/eks/cluster.ts#L822-L830

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

flostadler:
One workaround is setting the `instance_roles` parameter to an empty array:
```
from pulumi_aws import s3
import pulumi_eks as eks

cluster2 = eks.Cluster('eks-cluster', instance_roles=[])
```

I'd argue that this is not really intuitive, but removing that role now when `skipDefaultNodeGroup` is true would be a breaking change because some users might use the role through the `instanceRoles` output.

flostadler:
We could add an additional flag `skipInstanceRole`, but this would add clutter to the API surface of the cluster. I'd rather take this small enhancement with the next major release

flostadler:
We've addressed this as part of https://github.com/pulumi/pulumi-eks/pull/1411. This will be released with EKS v3

cleverguy25:
Added to epic https://github.com/pulumi/home/issues/3558

flostadler:
This is fixed in the latest major release https://github.com/pulumi/pulumi-eks/releases/tag/v3.0.0

PR #1411, opened 2024-09-30
Do not create instance role when `skipDefaultNodeGroup` is enabled

flostadler:
When enabling `skipDefaultNodeGroup` we should not be creating resources related to the default node group. This includes the default instance role.

With this change the `Cluster` component will no longer create this default role when `skipDefaultNodeGroup` is enabled.

One of the tests needs re-recording because we're introducing a breaking change (i.e. not creating the extraneous instance role), but in doing so we're aligning the provider with user expectations. Once this is merged in I'll release the next version and re-record.

Fixes https://github.com/pulumi/pulumi-eks/issues/1176


flostadler:
The change is failing the `TestEksAuthModeUpgrade` test that has a recorded baseline. This is expected because we're changing the default behavior of the cluster here, but we're more closely aligning it with customer expectations.

flostadler:
This is a bit more complex than initially expected. The node groups use the cluster's default instanceProfile when none is passed in. Therefore they also use the default instance role.

The better course of action here is to add a flag like `skipDefaultInstanceRole`.

sdk/python/pulumi_eks/_inputs.py
```diff
@@ -289,7 +289,7 @@ class ClusterNodeGroupOptionsArgsDict(TypedDict):
 
         Note: Given the inheritance of auto-generated CF tags and `cloudFormationTags`, you should either supply the tag in `autoScalingGroupTags` or `cloudFormationTags`, but not both.
         """
-        bootstrap_extra_args: NotRequired[str]
+        bootstrap_extra_args: NotRequired[pulumi.Input[str]]
```
flostadler, line 292: Merging https://github.com/pulumi/pulumi-eks/pull/1414 & https://github.com/pulumi/pulumi-eks/pull/1415 ontop of each other dropped those changes.
This adds is back now, work tree is clean again

docs/eks-v3-migration.md
```diff
@@ -160,3 +160,7 @@ The `NodeGroup` and `NodeGroupV2` components now accept inputs for the following
 - `nodeAssociatePublicIpAddress`
 
 If you're using Go you'll need to adjust your program to handle those types being inputs.
+
+### Cluster does not create extraneous instance IAM role if `skipDefaultNodeGroup` is set to `true`
+
+Previously the Cluster component created a default instance IAM role even if `skipDefaultNodeGroup` was set to `true`. This role gets correctly omitted now if you're specifying `skipDefaultNodeGroup`.
```
t0yv0, line 166: "if you are"

sdk/python/pulumi_eks/_inputs.py
t0yv0: Are we missing CI check guarantees that the worktree should be clean possibly?

t0yv0 (approved):

sdk/python/pulumi_eks/_inputs.py
flostadler: We're merging those into a feature branch that's only built when cutting an alpha release or when triggered manually. It's checking it as part of those workflows

Issue #1274, opened 2024-07-23
Allow all arguments in NodeGroupV2 to be `pulumi.Input`


Makeshift:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

I ran into an issue today when trying to create a NodeGroup whose labels and taints were not known initially and were therefore `pulumi.Output`'s. Unfortunately, the `labels` and `taints` args (and a few more) in the `NodeGroupV2` component are treated and typed as plain values, so it isn't possible to use potentially-unresolved values when creating them.

Please add support for `labels` and `taints` to be resolved from `pulumi.Output`.

### Affected area/feature

This component.

t0yv0:
Thank you for trying to send a PR here! Unfortunately it's a little more involved. Our team will have a look at this when planning the next major release, possibly combined with  https://github.com/pulumi/pulumi-eks/issues/603

t0yv0:
CC @flostadler

flostadler:
We've addressed this as part of https://github.com/pulumi/pulumi-eks/pull/1415. This will be released with EKS v3

cleverguy25:
Added to epic https://github.com/pulumi/home/issues/3558

PR #1415, opened 2024-10-02
Change arguments of NodeGroup & NodeGroupV2 to accept inputs

flostadler:
Historically the following `NodeGroup` & `NodeGroupV2` input properties have been plain:
- `kubeletExtraArgs`
- `bootstrapExtraArgs`
- `labels`
- `taints`
- `nodeAssociatePublicIpAddress`

Those should instead be inputs so users can pass outputs into them.

fixes https://github.com/pulumi/pulumi-eks/issues/1274


provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -2206,8 +2198,7 @@ func nodeGroupProperties(cluster, v2 bool) map[string]schema.PropertySpec {
 		},
 		"nodeAssociatePublicIpAddress": {
 			TypeSpec: schema.TypeSpec{
-				Type:  "boolean",
```
t0yv0, line 2209: 👍

t0yv0 (approved):

t0yv0:
As discussed the expected breaking changes are only for Go (and those are slight) so this is very welcome overall.

PR #1416, opened 2024-10-02
Add parameter for skipping default security groups

flostadler:
Pulumi EKS currently always creates a cluster security group and node security group.
- The cluster security group gets assigned to the control plane ENIs in addition to the security group EKS creates (see [AWS Docs](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html)). This security group gets an ingress rule from the node security group.
- The node security group gets assigned to `NodeGroup` and `NodeGroupV2` components that do not specify a custom security group.

Users that either manage the node security themselves or use the `ManagedNodeGroup` component (uses the EKS created SG) do not need those default security groups.

This change adds a flag on the cluster (`skipDefaultSecurityGroups`) that will skip creating those default security groups. Instead.

This introduces a small breaking change, the `clusterSecurityGroup`, `nodeSecurityGroup` and `clusterIngressRule` outputs are now optional. The impact of this should be minimal because users that create custom node groups usually do not use the security groups of the cluster for that. If they do, they need to add a null check.

Fixes https://github.com/pulumi/pulumi-eks/issues/747


corymhall (approved):

nodejs/eks/nodegroup.ts
```diff
@@ -1093,19 +1093,19 @@ function createNodeGroupV2Internal(
...
-                throw new pulumi.ResourceError(
-                    `The NodeGroup's nodeSecurityGroup and the cluster option nodeSecurityGroupTags are mutually exclusive. Choose a single approach`,
-                    parent,
-                );
+    const coreSecurityGroupId = core.nodeGroupOptions.nodeSecurityGroup?.apply((sg) => sg?.id);
+    pulumi
+        .all([coreSecurityGroupId, args.nodeSecurityGroup?.id, core.nodeSecurityGroupTags])
+        .apply(([coreSecurityGroup, nodeSecurityGroup, sgTags]) => {
+            if (coreSecurityGroup && nodeSecurityGroup) {
+                if (sgTags && coreSecurityGroup !== nodeSecurityGroup) {
```
t0yv0, line 1101: Always suspicious of equality but this look like it's over string right? Might be OK. Perhaps call the variables "coreSecurityGroupId" or some such.

nodejs/eks/cluster.ts
```diff
@@ -620,6 +622,28 @@ export function createCore(
         },
     );
 
+    // apply the tags to the cluster security group if it was created by EKS
+    if (!eksClusterSecurityGroup) {
+        const eksCreatedSgId = eksCluster.vpcConfig.clusterSecurityGroupId;
+        eksClusterSecurityGroup = aws.ec2.SecurityGroup.get(
```
t0yv0, line 628: Does this .get work OK for you inlight of today's discussion about a possible bug in MLC R.Get ?

nodejs/eks/cluster.ts
```diff
@@ -620,6 +622,28 @@ export function createCore(
...
+            `${name}-eksClusterSecurityGroup`,
+            eksCreatedSgId,
+            {},
+            { parent, provider },
+        );
+        pulumi.output(args.clusterTags).apply((tags) => {
```
t0yv0, line 634: This is a new feature that I was not expecting to find must have missed something from the PR description. It looks important and a nice feature.

sdk/dotnet/Cluster.cs
```diff
@@ -592,6 +592,15 @@ public InputList<Inputs.RoleMappingArgs> RoleMappings
...
         public bool? SkipDefaultNodeGroup { get; set; }
 
+        /// <summary>
+        /// If this toggle is set to true, the EKS cluster will be created without the default node and cluster security groups. Instead it will use the security group created by the EKS service.
+        /// Defaults to false.
+        /// 
+        /// See for more details: https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html.
+        /// </summary>
+        [Input("skipDefaultSecurityGroups")]
+        public bool? SkipDefaultSecurityGroups { get; set; }
```
t0yv0, line 602: Weird three-state boolean. true, false, and unknown. Alas, not worth fixing probably.

nodejs/eks/nodegroup.ts
flostadler: Yeah it's over string. Noticed that it was comparing `Output<string>` before..

nodejs/eks/cluster.ts
flostadler: Had to remove now that we're not able to re-use the cluster sg

nodejs/eks/cluster.ts
flostadler: Yeah, this was before our call. I just pushed a new version

flostadler:
@t0yv0 @corymhall can you please have another look? I had to change the approach due to the issues with `.get` in MLCs.

t0yv0 (approved):

t0yv0:
> I had to change the approach due to the issues with .get in MLCs.

Very much worth spending time on to extract a repro and link here 🙏 If you are out of capacity sent it over to me as I might have some bandwidth.

flostadler:
> > I had to change the approach due to the issues with .get in MLCs.
> 
> Very much worth spending time on to extract a repro and link here 🙏 If you are out of capacity sent it over to me as I might have some bandwidth.

Fully agreed! I already had a chat with will about this. Will try to create a minimal repro tomorrow to further investigate

PR #1429, opened 2024-10-08
Add example for AWS feature: Security Groups for Pods

flostadler:
This adds an example (and acceptance test) for the AWS feature: Security Groups for Pods.

The configuration is derived from this AWS example: https://docs.aws.amazon.com/eks/latest/userguide/security-groups-pods-deployment.html

### DIFF
--- a/examples/pod-security-groups/README.md
+++ b/examples/pod-security-groups/README.md
@@ -1,8 +1,8 @@
-# EKS Pod Security Groups
+# Pod Security Groups
 
-Demonstrates how to configure your Pulumi EKS Cluster to use Pod Security Groups.
+This example demonstrates how to set up Security Groups for Pods in an Amazon EKS cluster using Pulumi.
 
-The following shows which pods get deployed by this example and how communication is enforced by the security groups:
+## Topology
 
 ```mermaid
 graph TD
@@ -10,10 +10,24 @@ graph TD
     wrongCaller[wrong-caller Job]
     nginx[nginx]
     coredns[corends]
-
     caller -- Allow --> nginx
     caller -- Allow --> coredns
     wrongCaller -. Deny .-> nginx
     wrongCaller -- Allow --> coredns
     nginx -- Allow --> coredns
 ```
+
+## Components
+
+1. **caller Job**: A Kubernetes Job that is allowed to communicate with both the nginx pod and CoreDNS.
+2. **wrong-caller Job**: A Kubernetes Job that is allowed to communicate with CoreDNS but denied access to the nginx pod.
+3. **nginx**: A web server pod that can communicate with CoreDNS.
+4. **CoreDNS**: The cluster's DNS service, accessible by all pods.
+
+## Security Group Rules
+
+This example implements the following Security Group rules:
+
+1. Allow the `caller` Job to access both nginx and CoreDNS.
+2. Deny the `wrong-caller` Job access to nginx, while allowing access to CoreDNS.
+3. Allow nginx to access CoreDNS.
--- a/examples/pod-security-groups/README.md
+++ b/examples/pod-security-groups/README.md
@@ -1,0 +1,3 @@
+# EKS Pod Security Groups
+
+Demonstrates how to configure your Pulumi EKS Cluster to use Pod Security Groups.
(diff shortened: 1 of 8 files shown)


examples/pod-security-groups/iam.ts
```diff
@@ -0,0 +1,28 @@
...
+// Creates a role and attches the EKS worker node IAM managed policies
+export function createRole(name: string): aws.iam.Role {
+    const role = new aws.iam.Role(name, {
+        assumeRolePolicy: aws.iam.assumeRolePolicyForPrincipal({
+            Service: "ec2.amazonaws.com",
+        }),
+    });
+
+    let counter = 0;
+    for (const policy of managedPolicyArns) {
```
t0yv0, line 20: Not sure if you find functional constructs cute or distracting but fwiw:

```
> ([1,2,3]).forEach((x, i) => console.log(x, i));
1 0
2 1
3 2

```

corymhall (approved):

examples/pod-security-groups/iam.ts
flostadler: I like them! But to be very honest, I just copied this from another test

examples/pod-security-groups/index.ts
```diff
@@ -0,0 +1,196 @@
...
+          ],
+        }],
+        restartPolicy: "Never",
+      },
+    },
+    backoffLimit: 3,
+  },
+}, { provider: kube, dependsOn: [nginx, callerSgp] });
+
+function configureClusterAccess(name: string, cluster: eks.Cluster, sg: aws.ec2.SecurityGroup) {
```
t0yv0, line 178: Should this live in a helper file?

examples/pod-security-groups/index.ts
```diff
@@ -0,0 +1,196 @@
...
+  },
+  spec: {
+    podSelector: { matchLabels: { app: "wrong-caller" } },
+    securityGroups: {
+      groupIds: [wrongCallerSg.id],
+    }
+  },
+}, { provider: kube });
+
+// Create a job that is not allowed to curl the nginx service. The job will fail if it can reach the service.
```
t0yv0, line 152: This feels like a counter-example to test something instead of a best-practice example of how to GTD, or am I misunderstanding? Does it belong under ./examples?

examples/pod-security-groups/index.ts
flostadler: It's an important part of this example (allowing the pods to reach things like coredns). Putting it into another file would hide that aspect IMO

examples/pod-security-groups/index.ts
```diff
@@ -0,0 +1,196 @@
...
+
+/*
+ * Verify that the SecurityGroupPolicy is working as expected.
+ */
+
+// Create a SecurityGroup for the caller job
+const callerSg = new aws.ec2.SecurityGroup("caller", {
+  vpcId: eksVpc.vpcId,
+});
+// Allow all traffic between the cluster and the caller SecurityGroup
```
t0yv0, line 81: Consider consistent spacing. My eye not trained on TypeScript style but should this have an empty line before //?

examples/pod-security-groups/index.ts
flostadler: I'd say showing that traffic can be restricted is an important part of a firewall example.

t0yv0 (approved):

examples/pod-security-groups/index.ts
t0yv0: Hmm ok. I'm surprised it works while written at the bottom of the file but I guess that's how Node works.

examples/pod-security-groups/index.ts
t0yv0: Hmm, yeah I concede the point. A few comments to that end then perhaps? Maybe in the readme? I still get confused by mixing tests and examples in one.

examples/pod-security-groups/index.ts
flostadler: Uh, that was by accident. Somehow format on save was turned off

examples/pod-security-groups/README.md
```diff
@@ -0,0 +1,3 @@
+# EKS Pod Security Groups
+
+Demonstrates how to configure your Pulumi EKS Cluster to use Pod Security Groups.
```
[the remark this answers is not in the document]
flostadler: Good idea! I added one

PR #1432, opened 2024-10-09
Add example for EKS feature: Network Policies

flostadler:
This adds an example (and acceptance test) for EKS Network Policies.

The configuration is derived from this AWS example: https://docs.aws.amazon.com/eks/latest/userguide/cni-network-policy-configure.html

### DIFF
--- a/examples/network-policies/README.md
+++ b/examples/network-policies/README.md
@@ -1,3 +1,33 @@
 # Network Policies
 
-Demonstrates how to use Kubernetes Network Policies with EKS.
+This example demonstrates how to set up Kubernetes Network Policies in an Amazon EKS cluster using Pulumi.
+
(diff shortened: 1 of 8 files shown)


corymhall (approved):

examples/network-policies/README.md
```diff
@@ -0,0 +1,3 @@
+# Network Policies
+
+Demonstrates how to use Kubernetes Network Policies with EKS.
```
t0yv0, line 3: This is very laconic.

t0yv0 (approved):

examples/network-policies/README.md
flostadler: Added more details and a mermaid diagram

Issue #1425, opened 2024-10-07
Pulumi EKS v3.0.0 Release


flostadler:
AWS recently announced the deprecation of two features used by default in Pulumi EKS: the aws-auth ConfigMap and the AL2 operating system. Pulumi EKS v3 addresses these deprecations, enhances the maintainability of the provider, and aligns it with EKS best practices.
This release delivers significant improvements in flexibility and security, introducing new features to enhance your Kubernetes experience on AWS.

### Support for Amazon Linux 2023 and Bottlerocket

We've expanded the operating system options for node groups in EKS v3 to address the upcoming deprecation of Amazon Linux 2 (AL2). You can now choose between Amazon Linux 2 (deprecated), Amazon Linux 2023 and Bottlerocket for your EKS nodes. This flexibility allows you to select the OS that best fits your workloads, security requirements, and compliance needs, while ensuring you're using a supported and actively maintained operating system. We've introduced a new `operatingSystem` property for node groups to facilitate this choice.

### Access Entries for IAM Integration

AWS has introduced Access Entries as a new method for granting IAM principals access to Kubernetes resources. This approach relies solely on AWS resources for managing Kubernetes auth, replacing the deprecated `aws-auth` ConfigMap. You can now leverage Access Entries by setting the `authenticationMode` to `API` in your cluster configuration.

### EKS Managed Addons

The EKS cluster components `vpc-cni`, `coredns`, and `kube-proxy` are now configured as EKS managed addons. This change simplifies management, especially for clusters with private API endpoints, and ensures that these critical components stay up to date automatically. Additionally it removes the dependency on `kubectl`, allowing pulumi-native management of clusters.

### Cluster Autoscaler Integration

Pulumi EKS v3 introduces better support for the Kubernetes Cluster Autoscaler. A new `ignoreScalingChanges` parameter for node groups allows Pulumi to ignore external scaling changes, facilitating seamless integration with dynamic scaling solutions.

### EKS Security Groups for Pods and Network Policies

We've added support for EKS security groups for pods and EKS Network Policies, providing more granular control over pod-to-pod and pod-to-external network communication within your EKS clusters.

### `NodeGroup` component deprecation
[...]

PR #1441, opened 2024-10-14
Fix ManagedNodeGroup taints being wrongly set in userdata

flostadler:
The taints for the `ManagedNodeGroup` component were being wrongly calculated when using custom userdata.
That was the case because the EKS service uses different capitalization for the taint effect enum than the kubernetes API(e.g. `NO_SCHEDULE` vs `NoSchedule`).
When building the custom userdata we need to map the EKS style enums to kubernetes style enums, otherwise it doesn't work.

Fixing this also revealed that taint values being absent aren't correctly handled either. The change fixes that as well.

While this was found during the v3 beta testing ([see](https://github.com/pulumi/pulumi-eks/issues/1425#issuecomment-2408954568)), it's a bug on v2 as well. The bug fix is targeting v3 anyways because the node group and user data logic was heavily rewritten during the development of v3. Back porting that to v2 would be too big of an effort given that v3 is right around the corner.


rquitales (approved):

examples/cluster-py/__main__.py
```diff
@@ -84,11 +84,15 @@
...
+coredns_version = aws.eks.get_addon_version(addon_name="coredns",
+    kubernetes_version=cluster3["eks_cluster"]["version"],
+    most_recent=True)
+
 coredns = eks.Addon(
     f"{project_name}-cluster3-coredns",
     cluster=cluster3,
     addon_name="coredns",
-    addon_version="v1.11.1-eksbuild.9",
+    addon_version=coredns_version.version,
```
t0yv0, line 95: Unrelated fix looks like.

t0yv0: But great! I was fixing this again as port of ops, your solution is more robust.

t0yv0 (approved):

PR #1443, opened 2024-10-14
Upgrade coredns addon version

t0yv0:
### Proposed changes

Fixes https://github.com/pulumi/pulumi-eks/issues/1437

CoreDNS addon version used in tests is bumped to the latest available in to compensate for a "version not supported" error discovered by nightly tests:

 https://docs.aws.amazon.com/eks/latest/userguide/managing-coredns.html

### DIFF
--- a/examples/cluster/index.ts
+++ b/examples/cluster/index.ts
@@ -66,7 +66,7 @@ const cluster4 = new eks.Cluster(`${projectName}-4`, {
 const coredns = new eks.Addon("coredns", {
   cluster: cluster3,
   addonName: "coredns",
-  addonVersion: "v1.11.1-eksbuild.9",
+  addonVersion: "v1.11.3-eksbuild.1",
   resolveConflictsOnUpdate: "PRESERVE",
   configurationValues: {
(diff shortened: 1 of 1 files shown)


corymhall (approved):

examples/cluster/index.ts
```diff
@@ -66,7 +66,7 @@ const cluster4 = new eks.Cluster(`${projectName}-4`, {
 const coredns = new eks.Addon("coredns", {
   cluster: cluster3,
   addonName: "coredns",
-  addonVersion: "v1.11.1-eksbuild.9",
```
flostadler, line 69: We should rather dynamically fetch the correct version using https://www.pulumi.com/registry/packages/aws/api-docs/eks/getaddonversion/. That way we'll not have to keep updating this

You'll have to do the same for the `cluster-py` example.

t0yv0: Closing in favor of https://github.com/pulumi/pulumi-eks/pull/1441 I think that's fixed there with dynamic fetching.

Issue #1041, opened 2024-02-22
Fargate tests are flakey


rquitales:
### What happened?

Surfaced in #1039. Our daily cron job fails about 33% of the time due to our Fargate tests.

The error from the test is:
```
 utils.go:290: 
          	Error Trace:	/home/runner/work/pulumi-eks/pulumi-eks/examples/utils/utils.go:290
          	            				/home/runner/work/pulumi-eks/pulumi-eks/examples/utils/utils.go:225
          	            				/home/runner/work/pulumi-eks/pulumi-eks/examples/utils/utils.go:84
          	            				/home/runner/work/pulumi-eks/pulumi-eks/examples/utils/utils.go:90
          	            				/opt/hostedtoolcache/go/1.21.6/x64/src/runtime/asm_amd64.s:1650
          	Error:      	Not equal: 
          	            	expected: 3
          	            	actual  : 2
          	Test:       	TestAccFargate
          	Messages:   	2 out of 3 Pods are ready
```

It is unclear why only 2 of the 3 pods can become live and requires further debugging.

### Example

https://github.com/pulumi/pulumi-eks/issues/1039

### Output of `pulumi about`

-

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

thomas11:
@rquitales I think we also need to disable Fargate tests until this is resolved, otherwise we'll constantly get new P1s like this #1043.

PR #1445, opened 2024-10-16
Add scalar types for most commonly used resource outputs

flostadler:
To ease the impact of the breaking API changes caused by generating the node SDK, we decided to add additional scalar inputs that simplify UX across all SDKs (for more details [see internal doc](https://docs.google.com/document/d/1f97nmDUG_nrZSllYxu_XSeI7ON8vhZzfVrdBTQQmZzw/edit#heading=h.fbweiu8gc5bw)).

This change adds the scalar properties mentioned in the doc and adds acceptance tests for them.
While adding the acceptance tests I noticed that running pods on Fargate doesn't work deterministically. In some cases the cluster fails to get healthy (coredns stuck in pending).
This was caused by a race-condition between coredns starting and the fargate profile being created. If the fargate profile deployed after coredns, the pods got stuck in pending because they got assigned to the `default-scheduler` instead of the `fargate-scheduler`.
The fix is relatively easy; making coredns depend on the fargate profile.

I'll separately update the migration guide.

### New properties

| Existing Resource |  | New Top Level Property | Description |
| :---- | :---- | :---- | :---- |
| `clusterSecurityGroup: Output<aws.ec2.SecurityGroup \| undefined>` |  | `clusterSecurityGroupId: Output<string>` | Only really useful property of a security group. Used to add additional ingress/egress rules. Default to `the EKS created security group id` |
| `nodeSecurityGroup: Output<aws.ec2.SecurityGroup \| undefined>` |  | `nodeSecurityGroupId: Output<string>` |  |
| `eksClusterIngressRule: Output<aws.ec2.SecurityGroupRule \| undefined>` |  | `clusterIngressRuleId: Output<string>` | Only really useful property of a rule. Default to `””` |
| `defaultNodeGroup: Output<eks.NodeGroupData \| undefined>` |  | `defaultNodeGroupAsgName: Output<string>` | The only useful property of the default node group is the auto scaling group. Exposing its name allows users to reference it in IAM roles, tags, etc. Default to `””` |
| `core` | `fargateProfile: Output<aws.eks.FargateProfile \| undefined>` | `fargateProfileId: Output<string>` | The id of the fargate profile. Can be used to reference it. Default to `””` |
|  |  | `fargateProfileStatus: Output<string>` | The status of the fargate profile. Default to `””` |
|  | `oidcProvider: Output<aws.iam.OpenIdConnectProvider \| undefined>` | `oidcProviderArn: Output<string>` & `oidcProviderUrl: Output<string>` & `oidcIssuer: Output<string` | Arn and Url are properties needed to set up IAM identities for pods (required for the assume role policy of the IAM role). Users currently need to trim the `https://` part of the url to actually use it. We should expose `oidcProvider` with that already done to ease usage. |

Fixes https://github.com/pulumi/pulumi-eks/issues/1041

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -656,61 +656,6 @@ export function createCore(
         );
     }
 
-    const corednsExplicitlyEnabled =
-        args.corednsAddonOptions?.enabled === true || args.corednsAddonOptions?.configurationValues;
-    // We can only enable the coredns addon if we have a node group to place it on
-    // This means we are either using the default node group or the cluster is a fargate cluster
-    // Also, if the user explicitly enables it then do what they want
-    pulumi.output(args.fargate).apply((fargate) => {
-        if (
-            corednsExplicitlyEnabled ||
-            ((fargate || !args.skipDefaultNodeGroup) && (args.corednsAddonOptions?.enabled ?? true))
-        ) {
-            const corednsVersion: pulumi.Output<string> = args.corednsAddonOptions?.version
-                ? pulumi.output(args.corednsAddonOptions.version)
-                : aws.eks
-                      .getAddonVersionOutput(
-                          {
-                              addonName: "coredns",
-                              kubernetesVersion: eksCluster.version,
-                              mostRecent: true, // whether to return the default version or the most recent version for the specified kubernetes version
-                          },
-                          { parent, provider },
-                      )
-                      .apply((addonVersion) => addonVersion.version);
-
-            const configurationValues = pulumi
-                .all([args.fargate, args.corednsAddonOptions?.configurationValues])
-                .apply(([fargate, configurationValues]) => {
-                    if (fargate) {
-                        return {
-                            computeType: "Fargate",
-                            ...configurationValues,
-                        };
-                    } else {
-                        return configurationValues;
-                    }
-                });
-
-            const corednsAddon = new aws.eks.Addon(
-                `${name}-coredns`,
-                {
-                    clusterName: eksCluster.name,
-                    addonName: "coredns",
-                    tags: args.tags,
-                    preserve: true,
-                    addonVersion: corednsVersion,
-                    resolveConflictsOnCreate:
-                        args.corednsAddonOptions?.resolveConflictsOnCreate ?? "OVERWRITE",
-                    resolveConflictsOnUpdate:
-                        args.corednsAddonOptions?.resolveConflictsOnUpdate ?? "OVERWRITE",
-                    configurationValues: stringifyAddonConfiguration(configurationValues),
-                },
-                { parent, provider },
-            );
-        }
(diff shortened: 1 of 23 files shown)


nodejs/eks/cluster.ts
```diff
@@ -1064,6 +1009,75 @@ export function createCore(
...
+                    resolveConflictsOnCreate:
+                        args.corednsAddonOptions?.resolveConflictsOnCreate ?? "OVERWRITE",
+                    resolveConflictsOnUpdate:
+                        args.corednsAddonOptions?.resolveConflictsOnUpdate ?? "OVERWRITE",
+                    configurationValues: stringifyAddonConfiguration(configurationValues),
+                },
+                {
+                    parent,
+                    provider,
+                    dependsOn: fargateProfile.apply((profile) => {
```
flostadler, line 1065: That's the fix for the Fargate issues. I needed to include it here otherwise I couldn't test the changes in this PR.

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -1103,6 +1145,7 @@ func generateSchema(version semver.Version) schema.PackageSpec {
 				},
 			},
 			"eks:index:VpcCniAddon": {
+				IsComponent: true,
```
flostadler, line 1148: Unrelated to this PR, but I noticed this was incorrectly set and we had `.get` functions for this component.

provider/cmd/pulumi-resource-eks/schema.json
```diff
@@ -1077,10 +1077,18 @@
                     "$ref": "/aws/v6.18.2/schema.json#/provider",
                     "description": "The AWS resource provider."
                 },
+                "clusterIngressRuleId": {
```
t0yv0, line 1080: The additions here can be copied into PR description to make searching easier. Referring to an internal doc is not that helpful for posterity. Thanks!

sdk/dotnet/VpcCniAddon.cs
```diff
@@ -14,7 +14,7 @@ namespace Pulumi.Eks
     /// For more information see: https://docs.aws.amazon.com/eks/latest/userguide/eks-add-ons.html
     /// </summary>
     [EksResourceType("eks:index:VpcCniAddon")]
-    public partial class VpcCniAddon : global::Pulumi.CustomResource
+    public partial class VpcCniAddon : global::Pulumi.ComponentResource
```
t0yv0, line 17: Will this break things? Should we call it out? Or it is a new resource.

t0yv0 (approved):
LGTM

provider/cmd/pulumi-resource-eks/schema.json
flostadler: Added it to the PR description

Issue #1436, opened 2024-10-11
`public_access_cidrs` is ignored if `endpoint_public_access` is set to `false` - could we mention this in the docs?


stooj:
### What happened?

In an effort to lock down access to an EKS cluster, we set `public_access_cidrs` to a list of [pulumi's IP addresses](https://www.pulumi.com/docs/pulumi-cloud/faq/#which-domains-and-ips-should-i-allowlist).

Unfortunately, `endpoint_public_access` was set to `false`, so the allowed cidrs option was ignored.

Would it be possible to include a note in the documentation to call out that setting this to false will mean you can't use the allowlist?

### Example

```python
import pulumi
import pulumi_eks as eks

cluster = eks.Cluster(
    "cluster",
    vpc_config=eks.ClusterVpcConfigArgs(
        public_access_cidrs=["0.0.0.0/0"],  # Will be ignored
        endpoint_public_access=False,
    ),
)
```

### Output of `pulumi about`

---

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #1452, opened 2024-10-17
Add validation for public access CIDR

flostadler:
Setting public access CIDRs with public access disabled does not work, but the EKS service doesn't validate this case.
This can lead (and has) to very confusing debugging sessions à la "why can my IP not access the cluster endpoint, it's included in the public access CIDR range!".

This change adds validation for the public access CIDR.

Fixes https://github.com/pulumi/pulumi-eks/issues/1436

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -587,7 +587,19 @@ export function createCore(
                 subnetIds: clusterSubnetIds,
                 endpointPrivateAccess: args.endpointPrivateAccess,
                 endpointPublicAccess: args.endpointPublicAccess,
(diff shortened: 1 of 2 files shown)


t0yv0 (approved):
Nice! I wish unit tests with outputs worked better so we could dispatch of this sort of problem in the context of Node unit tests.

flostadler:
> Nice! I wish unit tests with outputs worked better so we could dispatch of this sort of problem in the context of Node unit tests.

Yeah... :( we'd have to do a refactor to make it unit-testable because of the issues in https://github.com/pulumi/pulumi-aws/issues/4585

PR #1459, opened 2024-10-22
Upgrade golangci-lint to v1.61.0

flostadler:
golangci-lint is failing without any output in CI right now. Upgrading it to the latest version fixes it

Fixes https://github.com/pulumi/pulumi-eks/issues/1456


t0yv0 (approved):
Sure, good to keep this up to date.

PR #1464, opened 2024-10-28
Fix ManagedNodeGroup with custom launch template and AMI type

### LINKED ISSUE #1463: InvalidParameterException: You cannot specify an AMI Type other than CUSTOM, when specifying an image id in your launch template
### What happened?

Encountered an error stating that an AMI Type other than CUSTOM cannot be specified when providing an image ID. However, no launch template has been set in the configuration.

Error message: ` InvalidParameterException: You cannot specify an AMI Type other than CUSTOM, when specifying an image id in your launch template` 

### Example

```
  const managedNodeGroup =  new eks.ManagedNodeGroup(
    "nodegroup-28102024",
    {
      cluster: cluster,
      nodeGroupName:  "managed-ng-28102024",
      amiType: "AL2023_ARM_64_STANDARD",
      instanceTypes: ["t3.large"],
      diskSize: 50,
      kubeletExtraArgs: "--max-pods=110", 
      scalingConfig: {
        desiredSize: 2,
        minSize: 1,
        maxSize: 3,
      },
    },
  );
```

### Output of `pulumi about`

```
CLI          
Version      3.137.0
Go Version   go1.23.2
Go Compiler  gc

Plugins
KIND      NAME        VERSION
resource  aws         6.45.0
resource  awsx        2.13.0
resource  eks         3.0.0
resource  kubernetes  4.17.1
language  nodejs      unknown
resource  random      4.16.3

Host     
OS       darwin
Version  15.0.1
Arch     arm64

This project is written in nodejs: executable='/opt/homebrew/bin/node' version='v20.18.0'

Backend        
Name           myhostname
URL            s3://mys3bcuket
User           myuser
Organizations  
Token type     personal

```

### Additional context

I encountered [this issue - #1779](https://github.com/pulumi/pulumi-eks/issues/1179) when using pulumi/eks version 2.7.8. I attempted to update to version 3.0 to resolve it, but then encountered the issue described above

### Contributing

flostadler:
When creating an EKS managed NodeGroup with a custom launch template, the AMI type must be removed from the nodegroup itself and instead specified in the launch template.
This did not work if the user specified the AMI type instead of the provider determining it.

This change fixes that by correctly omitting the user defined AMI type as well in that case.

Fixes #1463

### DIFF
--- a/examples/tests/managed-ng-os/index.ts
+++ b/examples/tests/managed-ng-os/index.ts
@@ -54,6 +54,16 @@ const managedNodeGroupAL2023 = eks.createManagedNodeGroup("al-2023-mng", {
   nodeRole: role,
 });
 
+// Create a node group with a launch template and custom AMI
+const managedNodeGroup = new eks.ManagedNodeGroup("al-2023-mng-amitype-launch-template", {
+  ...scalingConfig,
+  cluster: cluster,
+  nodeRole: role,
+  amiType: "AL2023_ARM_64_STANDARD",
+  instanceTypes: ["t4g.medium"],
+  kubeletExtraArgs: "--max-pods=110",
+});
+
 const managedNodeGroupAL2023Taints = eks.createManagedNodeGroup("al-2023-mng-taints", {
   ...scalingConfig,
   cluster: cluster,
--- a/nodejs/eks/nodegroup.ts
+++ b/nodejs/eks/nodegroup.ts
@@ -2089,6 +2089,7 @@ export function createManagedNodeGroupInternal(
         delete nodeGroupArgs.version;
         delete nodeGroupArgs.releaseVersion;
         delete nodeGroupArgs.amiType;
+        amiType = undefined;
     } else if (amiType === undefined && args.operatingSystem !== undefined) {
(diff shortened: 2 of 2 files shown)


examples/tests/managed-ng-os/index.ts
```diff
@@ -54,6 +54,16 @@ const managedNodeGroupAL2023 = eks.createManagedNodeGroup("al-2023-mng", {
   nodeRole: role,
 });
 
+// Create a node group with a launch template and custom AMI
+const managedNodeGroup = new eks.ManagedNodeGroup("al-2023-mng-amitype-launch-template", {
+  ...scalingConfig,
+  cluster: cluster,
+  nodeRole: role,
+  amiType: "AL2023_ARM_64_STANDARD",
```
t0yv0, line 62: Should this instead warn the user to not specify amiType here since it is a no-op, or it actually takes effect in other parts of the system?

t0yv0 (approved):

examples/tests/managed-ng-os/index.ts
flostadler: It actually takes effect. This causes the ami id to be set in the launch template

PR #1484, opened 2024-11-13
Use --profile instead of AWS_PROFILE for kubeconfig

blampe:
This PR changes our kubeconfig logic to use a `--profile` arg instead of an `AWS_PROFILE` environment variable so it will always use the expected profile. It also parallelizes the relevant tests and simplifies workflows slightly.

As a user, if I generate a kubeconfig for a particular profile I would expect that configuration to _always_ use the profile I specified. However, because we rely on `AWS_PROFILE` it is possible for our generated kubeconfig to be inadvertently overridden by the presence of `AWS_ACCESS_KEY_ID`.

> Credentials from environment variables have precedence over credentials from the shared credentials and AWS CLI config file. Credentials specified in the shared credentials file have precedence over credentials in the AWS CLI config file. If AWS_PROFILE environment variable is set and the AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables are set, then the credentials provided by AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY will override the credentials located in the profile provided by AWS_PROFILE.

https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#id1

_**If I got this wrong and the overriding behavior is a "feature not a bug" please let me know!**_

I'll note that I discovered this as part of the prep work for https://github.com/pulumi/ci-mgmt/issues/1142. In particular, our tests currently do a few things to exercise profile switching behavior:
1. During CI setup, we set some fixed access keys for the `default` profile. (This is unnecessary.)
2. During CI setup, we set some fixed access keys for an `alt` profile. This is the profile we expect to use in `TestAccAwsProfile*` tests.
3. During `TestAccAwsProfile*` tests we unset `AWS_SECRET_ACCESS_KEY`, `AWS_ACCESS_KEY_ID`, and `AWS_SESSION_TOKEN` for our process.

Importantly, (3) is currently implemented such that (a) it prevents parallelization, and (b) subsequent queries to the k8s API server also lack ambient credentials.

After I refactored (3) to allow parallelization the tests started failing. Eventually I realized this was because I was unsetting credentials for the `pulumi` subprocess and our test's k8s client now had ambient credentials taking priority over the expected profile.


nodejs/eks/cluster.ts
```diff
@@ -219,7 +219,10 @@ export function generateKubeconfig(
     }
 
     if (includeProfile && opts?.profileName) {
-        env.push({ name: "AWS_PROFILE", value: opts.profileName });
+        // Use --profile instead of AWS_PROFILE because the latter can be
+        // overridden by ambient credentials:
+        // https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#id1
+        args = [...args, "--profile", opts.profileName];
```
blampe, line 225: The only potentially user-facing change.

.github/workflows/cron.yml
```diff
@@ -1,7 +1,6 @@
 env:
   ALT_AWS_ACCESS_KEY_ID: ${{ secrets.ALT_AWS_ACCESS_KEY_ID }}
   ALT_AWS_SECRET_ACCESS_KEY: ${{ secrets.ALT_AWS_SECRET_ACCESS_KEY }}
-  ALT_AWS_PROFILE: ${{ secrets.ALT_AWS_PROFILE }}
```
blampe, line 4: This and "Configure AWS CLI" are now handled directly by the tests.

examples/aws-profile-py/__main__.py
```diff
@@ -9,14 +9,17 @@
 if not os.getenv("ALT_AWS_PROFILE"):
     raise Exception("ALT_AWS_PROFILE must be set")
 
+if not os.getenv("AWS_REGION"):
+    raise Exception("AWS_REGION must be set")
+
```
blampe, line 14: Mirroring the ts test.

examples/aws-profile-py/__main__.py
```diff
@@ -25,6 +28,7 @@
 # Create the cluster using the AWS provider and credential opts.
 cluster = eks.Cluster(project_name,
                       provider_credential_opts=kubeconfig_opts,
+                      # TODO(#1475): bootstrap_self_managed_addons=false, # To speed up the test.
```
blampe, line 31: Just a suggestion -- I noticed the coredns addon adds ~900s to the test, and it would be really great to disable!

examples/aws-profile-py/requirements.txt
```diff
@@ -1,3 +1,3 @@
 pulumi>=3.0.0,<4.0.0
 pulumi-aws>=6.0.0,<7.0.0
-pulumi-eks>=2.0.0,<3.0.0
\ No newline at end of file
+pulumi-eks>=3.0.0,<4.0.0
```
blampe, line 3: Makes local debugging easier.

examples/examples_nodejs_test.go
```diff
@@ -300,11 +300,23 @@ func TestAccScopedKubeconfig(t *testing.T) {
...
+				{Key: "pulumi:disable-default-providers[0]", Value: "aws", Path: true},
+			},
+			RetryFailedSteps: false,
+			Env: []string{
+				"ALT_AWS_PROFILE=" + profile,
+				"AWS_PROFILE=",           // unset
+				"AWS_SECRET_ACCESS_KEY=", // unset
+				"AWS_ACCESS_KEY_ID=",     // unset
+				"AWS_SESSION_TOKEN=",     // unset
+			},
```
blampe, line 319: This unsets the credentials for the pulumi subprocess instead of globally.

examples/aws-profile-py/__main__.py
flostadler: The culprit here is not aws adding the coredns addon, they do that as a self-managed addon (just a helm chart) and it doesn't wait for coredns to be healthy.

We're now using the managed coredns addon by default. If you want to disable it for this test, you can do so by setting `corednsAddonOptions.enabled` to false. I don't think we need it for this test because we're already testing that with others.

nodejs/eks/cluster.ts
flostadler: This change sounds good to me!

Tbh, I'm a bit surprised that even `aws eks update-kubeconfig` uses the env variable for specifying the profile.

I think it would make sense to extend the `aws-profile` test to actually deploy something to the cluster. That way we also cover the change with upgrade tests to ensure this doesn't cause any cascading replacements

blampe: Yeah, based on https://github.com/aws/aws-cli/issues/7794 and https://github.com/aws/aws-cli/issues/4337it I'm even more convinced `--profile` is the correct behavior here.

I'll look into the upgrade test to make sure this doesn't trigger replacements. I don't think it will since the k8s provider primarily looks at the cluster/context (not user, which this impacts), but if it does we should be able to leverage the new `clusterIdentifier` config to pin the provider to the cluster.

blampe: > I think it would make sense to extend the aws-profile test to actually deploy something to the cluster. That way we also cover the change with upgrade tests to ensure this doesn't cause any cascading replacements

@flostadler I re-enabled this upgrade test and recorded a baseline from master with a ConfigMap deployed into the cluster. It passes locally for me and I confirmed manually that no replacements were triggered.

examples/aws-profile-py/__main__.py
blampe: This was a huge time saver, thank you!

flostadler:
LGTM, thanks for this improvement!

PR #1474, opened 2024-11-06
[WIP] ci-mgmt onboarding

blampe:
Notes

Template drops the following options because they can be overridden by makefile targets:
- setup script
- pretests
- generate_coverage_data (replaced with codecov)
- extra tests
- ssh private key
- gotestfmt
- p/examples tests

Example tests are all sharded together without language distinction.

To do:
- [ ] make test_nodejs_upgrade
- [ ] free disk space needed?
- [x] shard test_nodejs (15)
- [x] shard test_python (6) + go + dotnet
- install kubectl

Issue #1387, opened 2024-09-25
Re-Enable Upgrade tests


flostadler:
The release branch for EKS v3 contains breaking changes now. This causes the upgrade tests to fail. Because of that we disabled the upgrade tests.
Once the first alpha version was released we can re-record and re-enable them.

cleverguy25:
Added to epic https://github.com/pulumi/home/issues/3558

PR #1490, opened 2024-11-14
Move integration tests to `./tests`

blampe:
This consolidates integration tests (currently located in `./examples` and `./provider`) under `./tests`.

This is prep work for https://github.com/pulumi/pulumi-eks/pull/1474, and the net result is a clearer distinction between "fast" and "slow" tests meant to be run during the "prerequisites" and "tests" steps in CI. Everything under `./provider` can be considered "fast" and everything under `./tests` can be considered "slow" and worthy of sharding.

Specifically:
* Everything under `./examples/tests` is moved to `./tests/testdata/programs` as these were assumed to not be user-facing.
* Upgrade tests under `./provider` are also moved to `./tests` but tagged in a way that they continue to run as their own job. (These are the new `upgrade_test.go` and `validation_test.go` files.)
* `examples/utils/utils.go` moves to `tests/internal/utils`.
* `TestEksAuthModeUpgrade` wasn't re-enabled as part of #1387, so I re-enabled and re-recorded it.
* `go.mod` is consolidated at the repo root to keep dependencies consistent across the provider and tests. (There are a ton of additional DX advantages to structuring things this way.) This project is not meant to be consumed as a library so this is an acceptable breaking change.


Makefile
```diff
@@ -167,36 +167,33 @@ dist:: dist/${GZIP_PREFIX}-darwin-amd64.tar.gz
 dist:: dist/${GZIP_PREFIX}-darwin-arm64.tar.gz
 dist:: dist/${GZIP_PREFIX}-windows-amd64.tar.gz
 
-test_build::
-	cd examples/utils/testvpc && yarn install && yarn run tsc
-
```
blampe, line 172: This directory doesn't exist.

flostadler (approved):
Sweat. Looks good to me!
And good catch with the upgrade test that was missed

PR #1496, opened 2024-11-18
adds instanceProfileName field to nodegroups

jdavredbeard:
### Proposed changes
Add `instanceProfileName` as an `Input<string>` field to `NodeGroup` and `NodeGroupV2`. This would unblock https://github.com/pulumi/pulumi-self-hosted-installers/issues/181 by creating a path for users to create an `InstanceProfile` in a separate stack from the `NodeGroup` and pass the instance profile's name by stack reference. Currently https://github.com/pulumi/pulumi/issues/17515, prevents `NodeGroup` components from accepting `InstanceProfile` values created in another stack and provided to the `NodeGroup`'s stack by `aws.iam.InstanceProfile.get()`, and the `instanceProfile` parameter cannot accept an `InstanceProfile` resource passed by stack reference because the `instanceProfile` parameter is a plain `InstanceProfile` type, not an `Input<InstanceProfile>`.

### Related issues (optional)
Closes https://github.com/pulumi/pulumi-eks/issues/1497


nodejs/eks/cluster.ts
```diff
@@ -802,7 +802,7 @@ export function createCore(
                 name,
                 parent,
                 args.instanceRole,
-                args.instanceProfileName,
```
flostadler, line 805: We should do validation to ensure only one of them is set.

nodejs/eks/nodegroup.ts
```diff
@@ -652,19 +653,54 @@ export function createNodeGroup(
     return createNodeGroupInternal(name, args, pulumi.output(core), parent, provider);
 }
 
+export function resolveOrGetInstanceProfile(
```
flostadler, line 656: Could you please add function docs explaining what this function is exactly doing and when it's throwing errors. Thanks!

nodejs/eks/nodegroup.ts
```diff
@@ -652,19 +653,54 @@ export function createNodeGroup(
...
+        !c.nodeGroupOptions.instanceProfileName
+    ) {
+        throw new pulumi.ResourceError(
+            `an instanceProfile or instanceProfileName is required`,
+            parent,
+        );
+    }
+    if (
+        (args.instanceProfile || c.nodeGroupOptions.instanceProfile) &&
+        (args.instanceProfileName || c.nodeGroupOptions.instanceProfileName)
```
flostadler, line 676: You could simplify this with null coalescing:
```suggestion
        (args.instanceProfile?.instanceProfile && c.nodeGroupOptions?.instanceProfileName)
```

nodejs/eks/yarn.lock
```diff
@@ -4,28 +4,28 @@
 
 "@ampproject/remapping@^2.2.0":
   version "2.3.0"
-  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.3.0.tgz#ed441b6fa600072520ce18b43d2c8cc8caecc7f4"
```
flostadler, line 7: These seem to be unrelated changes

nodejs/eks/nodegroup.ts
```diff
@@ -652,19 +653,54 @@ export function createNodeGroup(
...
+    ) {
+        throw new pulumi.ResourceError(
+            `invalid args for node group ${name}, instanceProfile and instanceProfileName are mutually exclusive`,
+            parent,
+        );
+    }
+    if (args.instanceProfile || c.nodeGroupOptions.instanceProfile) {
+        return args.instanceProfile ?? c.nodeGroupOptions.instanceProfile!;
+    }
+    return aws.iam.InstanceProfile.get(
```
flostadler, line 686: I think pulling in the resource with the `.get` function would incur RUM costs. We circumvent this by using the data source instead.

It seems like we're only interested in the ARN, right? Instance Profile ARNs follow this format: `arn:aws:iam::<account-id>:instance-profile/<profile-name>`.
If we know the name we can construct it without having to load the resource.

In that case we need other props, we could use [`iam.getInstanceProfileOutput`](https://www.pulumi.com/registry/packages/aws/api-docs/iam/getinstanceprofile/) instead.

provider/cmd/pulumi-gen-eks/main.go
```diff
@@ -2308,7 +2308,11 @@ func nodeGroupProperties(cluster, v2 bool) map[string]schema.PropertySpec {
 				Ref:   awsRef("#/resources/aws:iam%2FinstanceProfile:InstanceProfile"),
 				Plain: true,
 			},
-			Description: "The ingress rule that gives node group access.",
+			Description: "The IAM InstanceProfile to use on the NodeGroup.",
+		},
+		"instanceProfileName": {
+			TypeSpec:    schema.TypeSpec{Type: "string"},
+			Description: "The name of the IAM InstanceProfile to use on the NodeGroup.",
```
flostadler, line 2315: We should point out here that `instanceProfile` and `instanceProfileName` is mutually exclusive.

flostadler (commented):
Left some comments. Love the unit tests! But it would be great if we could add an integration test for this as well.

nodejs/eks/cluster.ts
flostadler: Ah saw you're doing that further down.

nodejs/eks/nodegroup.ts
flostadler: Mhm wait, my suggestion is not correct.

Aren't we susceptible to null pointer issues though? Is `c.nodeGroupOptions` guaranteed to be set?

flostadler: Just checked, we're good. `c.nodeGroupOptions` is set even when we're not creating the default node group. All good!

nodejs/eks/nodegroup.ts
```diff
@@ -653,42 +654,185 @@ export function createNodeGroup(
...
+ * Returns either the name of a passed InstanceProfile resource or a passed instanceProfileName.
+ * If an instanceProfile or instanceProfileName is passed in both the args and c.nodeGroupOptions,
+ * uses the input from args.
+ * Errors if no instanceProfile and no instanceProfileName is passed.
+ * Errors if both an instanceProfile and an instanceProfileName is passed.
+ */
+export function resolveInstanceProfileName(
     name: string,
     args: Omit<NodeGroupOptions, "cluster">,
-    c: pulumi.UnwrappedObject<CoreData>,
```
flostadler: Why did you have to make this change?

nodejs/eks/nodegroup.ts
```diff
@@ -698,10 +842,8 @@ function createNodeGroupInternal(
     parent: pulumi.ComponentResource,
     provider?: pulumi.ProviderResource,
 ): NodeGroupData {
-    const instanceProfile = core.apply((c) =>
-        resolveOrGetInstanceProfile(name, args, c, parent, provider),
-    );
-
+    const instanceProfileName = resolveInstanceProfileName(name, args, core, parent)
```
flostadler, line 845: Why not do `core.apply(c => resolveInstanceProfileName(...))`  instead like before?
That way resolveInstanceProfileName can stay out of the output-space and operate on promptly available data instead. That would simplify testing

nodejs/eks/nodegroup.ts
```diff
@@ -653,42 +654,185 @@ export function createNodeGroup(
...
-        throw new pulumi.ResourceError(
-            `an instanceProfile or instanceProfileName is required`,
-            parent,
-        );
-    }
-    if (
-        (args.instanceProfile || c.nodeGroupOptions.instanceProfile) &&
-        (args.instanceProfileName || c.nodeGroupOptions.instanceProfileName)
-    ) {
+): pulumi.Output<string> | undefined {
```
flostadler, line 811: Shouldn't this always return a value?

nodejs/eks/nodegroup.ts
flostadler: The function would change to this in that case:

```ts
export function resolveInstanceProfileOrName(
    name: string,
    args: Omit<NodeGroupOptions, "cluster">,
    c: pulumi.UnwrappedObject<CoreData>,
    parent: pulumi.ComponentResource,
): pulumi.Output<string> {
    if (
        !args.instanceProfile &&
        !args.instanceProfileName &&
        !c.nodeGroupOptions.instanceProfile &&
        !c.nodeGroupOptions.instanceProfileName
    ) {
        throw new pulumi.ResourceError(
            `an instanceProfile or instanceProfileName is required`,
            parent,
        );
    }
    if (
        (args.instanceProfile || c.nodeGroupOptions.instanceProfile) &&
        (args.instanceProfileName || c.nodeGroupOptions.instanceProfileName)
    ) {
        throw new pulumi.ResourceError(
            `invalid args for node group ${name}, instanceProfile and instanceProfileName are mutually exclusive`,
            parent,
        );
    }

    return args.instanceProfile?.name ?? pulumi.output(c.nodeGroupOptions.instanceProfileName!) 
}
```

flostadler: I'd also slightly change the definedness checks to check for all possible cases:
```ts
export function resolveInstanceProfileName(
    name: string,
    args: Omit<NodeGroupOptions, "cluster">,
    c: pulumi.UnwrappedObject<CoreData>,
    parent: pulumi.ComponentResource,
): pulumi.Output<string> {
    if (
        !args.instanceProfile || !c.nodeGroupOptions.instanceProfile ||
        !args.instanceProfileName || !c.nodeGroupOptions.instanceProfileName
    ) {
        // todo correct message that nothing's set
        throw new pulumi.ResourceError(
            `invalid args for node group ${name}, instanceProfile and instanceProfileName are mutually exclusive`,
            parent,
        );
    }

    if (args.instanceProfileName) {
        return pulumi.output(args.instanceProfileName);
    } else if (c.nodeGroupOptions.instanceProfileName) {
        return pulumi.output(c.nodeGroupOptions.instanceProfileName);
    } else if (args.instanceProfile !== undefined) {
        return args.instanceProfile.name;
    } else if (c.nodeGroupOptions.instanceProfile) {
        return c.nodeGroupOptions.instanceProfile.name;
    } else {
        throw new pulumi.ResourceError(
            `an instanceProfile or instanceProfileName is required`,
            parent,
        );
    }
}
```

tests/testdata/programs/self-managed-ng-instanceProfile/index.ts
```diff
@@ -0,0 +1,162 @@
...
+
+const ngv2_instanceProfileNameFromClusterNgOpts = new eks.NodeGroupV2("ng-instanceProfileNameFromClusterNgOpts", {
+  ...capacity,
+  cluster: cluster_instanceProfileNameFromNgOpts,
+  instanceType: "t4g.medium",
+  operatingSystem: eks.OperatingSystem.AL2023,
+});
+
+export const passedInstanceProfileName = instanceProfile.name
+export const kubeconfig = cluster_NoIp_NoIpn.kubeconfig;
```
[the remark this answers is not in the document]
jdavredbeard: Whoops missed exporting the other kubeconfigs... sure I can reduce the scope of the tests, I was going for a more completionist approach

flostadler (approved):
Nice! Thanks a lot

PR #1506, opened 2024-11-28
Add a postUpgradeTask to regenerate schema

blampe:
Similar to https://github.com/pulumi/pulumi-cdk/pull/267, although I'm less confident `go run` will work inside Renovate. Worth a shot! If this does

Re-generate the schema whenever we bump codegen, aws, or k8s.

Edit: SDK steps will still need at least gradle, etc. Moving this to WIP.


t0yv0:
Very interested in something like this. If we run it a few times extra that may be OK (i.e. filtering is overly permissive). We can optimize the target being run.

t0yv0:
What's on the master branch is a bit more aggressive with filters but it does seem to work! Closing this I think this should be good enough for now.

Issue #1510, opened 2024-12-03
instanceRole is no longer automatically created when `skipDefaultNodeGroup: true`


bob-bins:
### What happened?

After I upgraded to v3 of this package, `pulumi preview` looks to delete the instanceRole for my EKS clusters. This is a regression caused by the resolution of https://github.com/pulumi/pulumi-eks/issues/1176, in this PR https://github.com/pulumi/pulumi-eks/pull/1411.

My workaround is currently to set min/desired/max replicas for `nodeGroupOptions` to `0`, but since my org requires specific tags to exist for launch configs and ASGs, and since seeing the frequent AMI changes ruins one of our workflows, this regression is forcing me to add 33 lines of code that were not previously necessary.

Switching our cluster to use a new instanceRole managed outside this module requires a lot of work given that all node groups need to be completely recreated.

Can a boolean option like `createInstanceRole` be exposed so we can maintain backward-compatible behavior?

### Example

N/A

### Output of `pulumi about`

```
$ pulumi about
CLI          
Version      3.128.0
Go Version   go1.22.5
Go Compiler  gc

Plugins
KIND      NAME    VERSION
language  nodejs  unknown

Host     
OS       darwin
Version  14.5
Arch     arm64
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

flostadler:
Hey @bob-bins, sorry this is causing you trouble!
As part of https://github.com/pulumi/pulumi-eks/issues/1176 we've took the decision to not create the default instance role if users disable the default node groups. Our assumption back then was that there's not a substantial need for this role when users opt out of the default node groups.

But given your inputs I can see this making an upgrade more painful than necessary! I'll track this as an enhancement and will try to get to it as soon as possible!

flostadler:
@bob-bins I just merged the PR to introduce a flag for creating the instance role. The change will be released shortly

PR #1511, opened 2024-12-04
Allow creating default instance role even when no default node group is created

flostadler:
As part of https://github.com/pulumi/pulumi-eks/issues/1176 we took the decision to not create the default instance role if users disable the default node groups. Our assumption back then was that there's not a substantial need for this role when users opt out of the default node groups.

https://github.com/pulumi/pulumi-eks/issues/1510 has shown that there is indeed users that were relying on this default instance role that was created "by accident". This original change made the upgrade to v3 more painful than necessary for this set of users!

This change introduces a new plain flag `createInstanceRole` that can be used to opt-into creating the default instance role even when no default node group is needed.

Resolves https://github.com/pulumi/pulumi-eks/issues/1510

### DIFF
--- a/nodejs/eks/cluster.ts
+++ b/nodejs/eks/cluster.ts
@@ -824,7 +824,10 @@ export function createCore(
         }
         instanceRoles = pulumi.output([args.instanceRole]);
         defaultInstanceRole = pulumi.output(args.instanceRole);
-    } else if (!args.skipDefaultNodeGroup) {
+    } else if (
+        args.createInstanceRole ||
+        (!args.skipDefaultNodeGroup && args.createInstanceRole === undefined)
+    ) {
         const instanceRole = new ServiceRole(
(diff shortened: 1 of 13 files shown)


corymhall (approved):

flostadler:
> Thanks a bunch @flostadler ! One nitpick is that `docs/eks-v3-migration.md` can be upgraded to mention `createInstanceRole`. Something like:
> 
> > Previously the Cluster component created a default node IAM role even if `skipDefaultNodeGroup` was set to `true`. This role gets correctly omitted now if you are specifying `skipDefaultNodeGroup`. If you have `skipDefaultNodeGroup` set to `true` and would like to keep the node IAM role through the migration, you may set `createInstanceRole` to `true`.

Absolutely, I'm already working on the changes to the official docs that get published to https://www.pulumi.com/docs/ and will back-port those to here afterwards!

Issue #1509, opened 2024-12-02
Support for EKS Auto Mode


jhamman:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

AWS EKS has recently added a new "Auto Mode" feature: https://aws.amazon.com/blogs/aws/streamline-kubernetes-cluster-management-with-new-amazon-eks-auto-mode/. We would like to try this feature in our Pulumi-managed EKS cluster. 

Specific questions:
1. How can I create a new cluster with this mode enabled?
2. can I upgrade an existing cluster to use this mode?

flostadler:
Hey @jhamman, thanks a lot for this feature request!

In order to support this new feature we'll first need it to be supported in the upstream terraform provider. Good news is that the [PR](https://github.com/hashicorp/terraform-provider-aws/issues/40373) for this has already been merged and seems to be queued up for the next release. Once that's released, it'll get pulled into `pulumi-aws` shortly after.
After that we can start the work on adding support for this in `pulumi-eks`.


To answer your second question, according to AWS docs it seems like you can enable auto mode on existing clusters and migrate workloads safely: https://docs.aws.amazon.com/eks/latest/userguide/migrate-auto.html.

flostadler:
Depends on: https://github.com/pulumi/pulumi-aws/issues/4873

flostadler:
Support for this landed in `pulumi-aws` just now. I'm working on adding support for it to pulumi-eks now

flostadler:
While implementing this I found a bug in the implementation of `aws.eks.Cluster` that prevents updating an existing cluster to use EKS Auto Mode: https://github.com/pulumi/pulumi-aws/issues/4885

This bug is in turn caused by the following upstream bug: https://github.com/hashicorp/terraform-provider-aws/issues/40411
I'm investigating what options we have to work around this

Issue #1475, opened 2024-11-08
Allow to pass `bootstrapSelfManagedAddons` argument to cluster


ahanoff:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently EKS cluster sets by default `bootstrapSelfManagedAddons` as `true`. That settings creates 3 addons: `vpc-cni`, `coredns`, and `kube-proxy`

### Affected area/feature

Prevents from creating clean cluster without any addon which is beneficial if you use custom CNI like Cilium

t0yv0:
Hi @ahanoff thanks for the issue. Have you tried the defaultAddonsToRemove property?

 https://www.pulumi.com/registry/packages/eks/api-docs/cluster/#defaultaddonstoremove_nodejs 

I believe if you populate that property with a list of "vpc-cni", "coredns", "kube-proxy" then they should not be auto-created anymore.

ahanoff:
> Have you tried the defaultAddonsToRemove property?

hi @t0yv0, thank! Yes, that's what I currently set while waiting for this issue resolution

> List of addons to remove upon creation. Any addon listed will be "adopted" and then removed

Actually I'm confused with two things:

-  `defaultAddonsToRemove` comment above as I'm not sure what `adopted` means in this case. If it means it will be installed first and then removed, would it be a case when AWS VPC CNI leaves some iptables rules?
- having two options (bootstrapSelfManagedAddons and defaultAddonsToRemove) to do same things?

Anyway I'm going test your suggestion this week and get back with feedback

t0yv0:
Great feedback, let me double-check and see if we can clarify the docstrings.

flostadler:
Hey @ahanoff, `defaultAddonsToRemove` wouldn't leave any ip tables rules behind because the adopt->delete procedure is done before there's any nodes in the cluster.
In detail, we're creating an EKS Addon for every addon you mention in `defaultAddonsToRemove`. This adopts the existing resources on the cluster into Pulumi's management. Finally, we're deleting those addons again, also cleaning up the underlying resources.

[...]

PR #1519, opened 2024-12-06
Add support for EKS Auto Mode

flostadler:
### Proposed changes

This change adds support for EKS Auto Mode (see [AWS docs](https://docs.aws.amazon.com/eks/latest/userguide/automode.html)). It's a new feature that fully automates compute, storage, and networking management for Kubernetes clusters.

To support EKS Auto Mode, `pulumi-aws` was upgraded to `6.63.0` and the required options are exposed via the Cluster component's input properties.
Additionally, it also exposes the `bootstrapSelfManagedAddons` property because it is required to be set to false when creating new clusters with EKS Auto Mode. For existing clusters it's not required that this is set to false.

### Related issues (optional)

Fixes https://github.com/pulumi/pulumi-eks/issues/1475
Closes https://github.com/pulumi/pulumi-eks/issues/1509


nodejs/eks/cluster.ts
```diff
@@ -496,25 +510,38 @@ export function createCore(
...
+                "AmazonEKSLoadBalancingPolicy",
+                "AmazonEKSNetworkingPolicy",
+            );
+        }
+
+        eksServiceRole = new ServiceRole(
             `${name}-eksRole`,
             {
-                service: "eks.amazonaws.com",
+                service: pulumi.interpolate`eks.${dnsSuffix}`,
```
flostadler, line 531: Noticed this was not partition independent

nodejs/eks/cluster.ts
```diff
@@ -1904,6 +2002,10 @@ export const AccessEntryType = {
      * For IAM roles associated with self-managed Windows node groups. Allows the nodes to join the cluster.
      */
     EC2_WINDOWS: "EC2_WINDOWS",
+    /**
+     * For IAM roles associated with EC2 instances that need access policies. Allows the nodes to join the cluster.
+     */
+    EC2: "EC2",
```
flostadler, line 2008: This was a huge PITA to discover. This new Access Entry type is not documented and required when using a custom node role with EKS Auto Mode.

nodejs/eks/servicerole.ts
```diff
@@ -45,15 +45,22 @@ export interface ServiceRoleArgs {
...
+    readonly assumeRoleActions?: pulumi.Input<string[]>;
 }
 
 /**
  * The ServiceRole component creates an IAM role for a particular service and attaches to it a list of well-known
  * managed policies.
  */
 export class ServiceRole extends pulumi.ComponentResource {
-    // The service role.
-    public readonly role: pulumi.Output<aws.iam.Role>;
```
flostadler, line 56: Previously this internal component exported a single output for the role, that depended on all policy attachments.
Because of that it becomes unknown when attaching new policies.

In order to upgrade an existing cluster to Auto Mode, you need to attach new policies. The Cluster's service role property triggers replacements, so the preview would've shown that the cluster gets replaced.
This new approach is a bit more careful. We expose am output for the fully resolved role (with all policies attached) and the direct role resource itself.
Now the consumers can pick and choose what they want.

tests/testdata/programs/replace-cluster-add-subnets/index.ts
```diff
@@ -22,7 +22,16 @@ const popped = pulumi.output(publicSubnetIds).apply(subnets => {
 const cluster = new eks.Cluster(`${projectName}`, {
     vpcId: vpc.vpcId,
     publicSubnetIds: popped,
-});
+}, { transforms: [args => {
```
flostadler, line 25: This test was not working correctly anymore. Previously changes to the subnets triggered replacements. this is not the case anymore. This test wants to test a cluster replacement so I forced it.

Background why this wasn't failing hard in the past:
EKS picks two of the provided subnets for its control plane ENIs. When updating them, they need to be in the same AZs as the original ones.
We have a 1/3 chance to pick the right ones (the VPC has 3 AZs) and we have 3 retries configured for the test. So we got lucky most of the times.

bryantbiggs:
> Additionally, it also exposes the bootstrapSelfManagedAddons property because it is required to be set to false when creating new clusters with EKS Auto Mode. For existing clusters it's not required that this is set to false.

Not quite. You can enable this, it just doesn't make sense. And in general, the better approach is to move to have this hard codeded as `false` and users will need to use the EKS Addon API to provision EKS addons when needed. This is considered a legacy behavior - but if EKS were re-designed today, it would not exist and users would only use the EKS Addon API for creating/modifying addons

flostadler:
> > Additionally, it also exposes the bootstrapSelfManagedAddons property because it is required to be set to false when creating new clusters with EKS Auto Mode. For existing clusters it's not required that this is set to false.
> 
> Not quite. You can enable this, it just doesn't make sense. And in general, the better approach is to move to have this hard codeded as `false` and users will need to use the EKS Addon API to provision EKS addons when needed. This is considered a legacy behavior - but if EKS were re-designed today, it would not exist and users would only use the EKS Addon API for creating/modifying addons

Thanks a lot for the feedback @bryantbiggs! Generally I agree with your sentiment, but hardcoding it to false would be a breaking change that we cannot take until the next major version because users expect the addons to be installed ATM.

I was thinking about an alternative until we can take the breaking change. We could just not expose that property and ignore any changes to it. When users enable EKS Auto Mode we set it to false. This would only have an effect during cluster creation.

bryantbiggs:
oh completely agree - this is what we did in the EKS module since the default value for the boostrap managed addons is `null` https://github.com/terraform-aws-modules/terraform-aws-eks/blob/c63a57908d7b4d51895d3f8502d60daae758b761/main.tf#L37

with a TODO reminder to set this to a hardcoded value of `false` on the next breaking change.

flostadler:
`TestAccClusterGo` is failing because of:
```
aws:eks:Addon (example-cluster-4-coredns):
      error:   sdk-v2/provider2.go:520: sdk.helper_schema: waiting for EKS Add-On (example-cluster-4-eksCluster-d1ccc7b:coredns) create: timeout while waiting for state to become 'ACTIVE' (last state: 'DEGRADED', timeout: 20m0s): provider=aws@6.63.0
```

While looking into this, I noticed that this test was never working. IPv6 cannot work with the current setup because the node role is missing `ec2:AssignIpv6Addresses`. Right now on master it's also failing ([see](https://github.com/pulumi/pulumi-eks/actions/runs/12219403658/job/34086025947#step:22:1832)), but the retry "works". It seems to update the tainted addon and not do any health checking for the addon anymore.
After updating `pulumi-aws` to `v6.63.0` it doesn't taint the addon anymore after a failed deploy. Instead it just doesn't save it to state at all. The retries now try to create the addon instead of updating it, and that fails because it already exists:
```
EKS Add-On (example-cluster-4-eksCluster-d1ccc7b:coredns): operation error EKS: CreateAddon, https response error StatusCode: 409, RequestID: 13a5e935-b8d2-458c-840b-fa65290c1228, ResourceInUseException: Addon already exists.: provider=aws@6.63.0
```

This seems to be an instance of https://github.com/pulumi/pulumi-aws/issues/4759 which was believed to be a bug, but this behavior might actually make a regression. I'll look into it!

nodejs/eks/cluster.ts
```diff
@@ -447,6 +454,13 @@ export function createCore(
         );
     }
 
+    if (args.autoMode?.enabled && !supportsAccessEntries(args.authenticationMode)) {
+        throw new pulumi.ResourceError(
```
corymhall, line 458: TIL!

examples/eks-auto-mode/index.ts
```diff
@@ -0,0 +1,139 @@
...
+        {type: SubnetType.Private, tags: {[`kubernetes.io/cluster/${clusterName}`]: "shared", "kubernetes.io/role/internal-elb": "1"}},
+    ],
+    subnetStrategy: "Auto"
+});
+
+const cluster = new eks.Cluster("eks-auto-mode", {
+    name: clusterName,
+    // EKS Auto Mode requires Access Entries, use either the `Api` or `ApiAndConfigMap` authentication mode.
+    authenticationMode: eks.AuthenticationMode.Api,
+    skipDefaultNodeGroup: true,
```
corymhall, line 26: All these defaults that we have to flip, are they the only option when using autoMode or can you still set `skipDefaultNodeGroup` to `false`?

Can we make different defaults when autoMode is enabled so the configuration is less verbose? As it is not it looks like enabling "autoMode" required _more_ configuration from the user.

e.g. can using autoMode look like this?

```ts
const cluster = new eks.Cluster("eks-auto-mode", {
    name: clusterName,
    vpcId: eksVpc.vpcId,
    publicSubnetIds: eksVpc.publicSubnetIds,
    privateSubnetIds: eksVpc.privateSubnetIds,
    autoMode: {
        enabled: true,
    }
});
```

corymhall (changes requested):
Looks good! Just some questions around the API and default values.

examples/eks-auto-mode/index.ts
flostadler: I updated the defaults for `skipDefaultNodeGroup` and `skipDefaultSecurityGroups`. I intentionally didn't update the defaults for the addons. While kube-proxy and vpc-cni are not necessarily required for Auto Mode, there's no downside in creating them and users need them if they add other node groups.

bryantbiggs: While it is possible to have both Auto Mode and EKS managed node groups in the same cluster - the reality is that its highly unlikely. I don't know if it breaks your current user contract but it would be better to not deploy any addons by default when Auto Mode is enabled

bryantbiggs: The downside to this approach is that users who are not fully versed on Auto Mode may view this as a bug or an EKS issue when the VPC CNI daemonst fails to create any pods, nor do any of their VPC CNI configurations show up

flostadler: It does break the user contract to some extent. Users of the provider expect it to manage the addons for them unless they opt out of it.
But the VPC CNI configuration not applying here is a good point, I didn't know about that! I think in that case it's better to opt for defaulting to false and letting users decide.

flostadler: Thanks @bryantbiggs for all the insights!

bryantbiggs: I work with the EKS service team and manage the EKS Terraform module - I'm happy to have a quick chat sometime to answer any questions and provide some insights on what we commonly see folks doing within their EKS provisioning. Just send me a message brybiggs@amazon.com and we can set something up if interested

nodejs/eks/cluster.ts
```diff
@@ -496,32 +510,48 @@ export function createCore(
...
-                ],
+                managedPolicyArns: managedPolicies.map((policy) => ({
+                    id: `arn:aws:iam::aws:policy/${policy}`,
+                    arn: pulumi.interpolate`arn:${partition}:iam::aws:policy/${policy}`,
+                })),
                 tags: args.tags,
+                // EKS Auto Mode needs "sts:TagSession" in addition to the default "sts:AssumeRole"
+                assumeRoleActions: args.autoMode?.enabled
+                    ? ["sts:AssumeRole", "sts:TagSession"]
+                    : undefined,
```
t0yv0, line 541: The comment makes it look like `undefined` should be `["sts:AssumeRole"]` instead?

tests/testdata/programs/auto-mode-upgrade/README.md
```diff
@@ -0,0 +1,3 @@
+# EKS Auto Mode
+
+TODO
```
t0yv0, line 3: Sketch a few sentences here?

nodejs/eks/cluster.ts
bryantbiggs: just add both regardless of auto mode - more policies are moving to use session tags

flostadler: `["sts:AssumeRole"]` is the default of `assumeRoleActions`

tests/testdata/programs/auto-mode-upgrade/README.md
flostadler: Whoops, this should actually be removed. Was a left over from copying the test setup. Good catch!

nodejs/eks/cluster.ts
flostadler: Good to know @bryantbiggs, thanks!

flostadler: Changed it to add `sts:TagSession` independent of EKS Auto Mode now

corymhall (approved):

PR #1529, opened 2024-12-16
chore: Remove gotestfmt from CI

t0yv0:
Instead use standard go test logging without -v and -json options. This should help reduce confusion as to which test is failing and which log belongs to which test.

Follows: https://github.com/pulumi/pulumi-gcp/pull/2712
Re: https://github.com/pulumi/pulumi-eks/issues/1527

### DIFF
--- a/.github/workflows/eks-cron.yml
+++ b/.github/workflows/eks-cron.yml
@@ -241,11 +241,6 @@ jobs:
         run: tar -zxf ${{ github.workspace}}/nodejs.tar.gz -C ${{github.workspace}}/nodejs
       - name: Install dependencies
         run: make install_nodejs_sdk
-      - name: Install gotestfmt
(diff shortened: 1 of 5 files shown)


blampe (approved):
This will stay dead after you're onboarded to ci-mgmt.

t0yv0:
Looks like there's still a little bit of references to it in ci-mght:

https://github.com/pulumi/ci-mgmt/blob/532904f665741a9d7e959af066ec51a6a7ab67f6/native-provider-ci/src/workflows.ts#L606

LMK if I can help rip out.

flostadler:
@t0yv0 the upgrade tests are still referencing gotestfmt and failing due to that

t0yv0:
Looking. Sorry about that. Do they only run on PRs and not on master?

PR #1541, opened 2024-12-18
Accept renovate suggestions

t0yv0:
This change accepts suggestions from https://github.com/pulumi/pulumi-eks/pull/1540 and has two things done on top:

1. update Go test code to the newest version of providertest

Let us see if the build and test passes.


t0yv0:
Looks like some minor breaks in pulumi test are a bit of a problem, looking.

nodejs/eks/package.json
```diff
@@ -23,8 +23,8 @@
     "bugs": "https://github.com/pulumi/pulumi-eks/issues",
     "dependencies": {
         "@pulumi/aws": "6.65.0",
-        "@pulumi/kubernetes": "4.18.3",
-        "@pulumi/pulumi": "^3.143.0",
```
t0yv0, line 27: Wondering if this will introduce issues similar to https://github.com/pulumi/pulumi-aws-apigateway/pull/178 that necessitates keeping a range or using a special YARN syntax?

blampe (approved):

PR #1549, opened 2024-12-23
Clean up end-to-end tests

flostadler:
This removes an obsolete test (managed-ng-os-py), it was added before nodejs and the rest of the SDKs shared the same implementation. This functionality is already covered by the nodejs based test (managed-ng-os).

Additionally it changes the e2e tests to place instances in private subnets in order to reduce the EIP usage (reduces contention and saves money).


tests/testdata/programs/managed-ng-os/index.ts
```diff
@@ -36,19 +36,20 @@ const cluster = new eks.Cluster("managed-ng-os", {
...
 export const kubeconfig = cluster.kubeconfig;
 
-const scalingConfig = {
+const baseConfig = {
   scalingConfig: {
     minSize: 1,
     maxSize: 1,
     desiredSize: 1,
   },
+  subnetIds: eksVpc.privateSubnetIds,
```
flostadler, line 45: This one alone should free up 14 EIPs...

t0yv0 (approved):

PR #1552, opened 2024-12-23
Automtically install pulumictl

t0yv0:
This will automatically install `pulumictl` via `go install` if one is not found on PATH.

This should help `make renovate` command to succeed in an environment where `pulumictl` is not installed in PATH.

If pulumictl is found in PATH the ambient version is used as before.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,9 @@ PROJECT_NAME := Pulumi Amazon Web Services (AWS) EKS Components
 # Override during CI using `make [TARGET] PROVIDER_VERSION=""` or by setting a PROVIDER_VERSION environment variable
 # Local & branch builds will just used this fixed default version unless specified
 PROVIDER_VERSION ?= 3.0.0-alpha.0+dev
-# Use this normalised version everywhere rather than the raw input to ensure consistency.
-VERSION_GENERIC = $(shell pulumictl convert-version --language generic --version "$(PROVIDER_VERSION)")
+
+# Use this normalized version everywhere rather than the raw input to ensure consistency.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -3,8 +3,9 @@ PROJECT_NAME := Pulumi Amazon Web Services (AWS) EKS Components
 # Override during CI using `make [TARGET] PROVIDER_VERSION=""` or by setting a PROVIDER_VERSION environment variable
 # Local & branch builds will just used this fixed default version unless specified
 PROVIDER_VERSION ?= 3.0.0-alpha.0+dev
-# Use this normalised version everywhere rather than the raw input to ensure consistency.
-VERSION_GENERIC = $(shell pulumictl convert-version --language generic --version "$(PROVIDER_VERSION)")
+
+# Use this normalized version everywhere rather than the raw input to ensure consistency.
+VERSION_GENERIC ?= $(shell make VERSION_GENERIC=$(PROVIDER_VERSION) version_generic) # recur to auto-install pulumictl
```
t0yv0, line 8: Without the recursive call to make, it seems a bigger refactor is required as `VERSION_GENERIC` expansion is run very early in the cycle by Make and is not preceded by any target execution. The recursive call avoids the larger refactor. Note we need to seed the VERSION_GENERIC variable with an incorrect value to avoid an infinite recursion.

flostadler (approved):
The change looks good, but I'm a bit worried that this could break in subtle ways in CI leading to broken releases. Let's do a pre-release right away to confirm the version is correctly set

PR #1559, opened 2024-12-23
chore: ensure nodedjs/eks dependencies are deduplicated

t0yv0:
In particular yarn may link two copies of @pulumi/pulumi into the provider causing difficult to diagnose failures. This change applies the approach from @pulumi/pulumi-awsx where the dependencies are proactively deduplicated at build time.

Also note that `yarn install` should receive a --frozen-lockfile option to actually respect the lockfile during build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,7 +147,8 @@ install_java_sdk::
 	#Intentionally empty for CI / CD templating
 
 nodejs/eks/node_modules: nodejs/eks/package.json nodejs/eks/yarn.lock
-	yarn install --cwd nodejs/eks --no-progress
+	yarn install --frozen-lockfile --cwd nodejs/eks --no-progress
+	cd nodejs/eks && yarn run check-duplicate-deps
 	@touch nodejs/eks/node_modules
 
 nodejs/eks/bin: nodejs/eks/node_modules ${EKS_SRC}
--- a/nodejs/eks/package.json
+++ b/nodejs/eks/package.json
@@ -13,7 +13,9 @@
         "format": "prettier --list-different --write .",
         "lint-check": "yarn run format-check && tslint -c ../tslint.json -p tsconfig.json",
         "lint": "yarn run format && tslint -c ../tslint.json -p tsconfig.json --fix",
-        "test": "jest"
+        "test": "jest",
+        "dedupe-deps": "yarn-deduplicate",
+        "check-duplicate-deps": "yarn-deduplicate --fail"
     },
     "homepage": "https://pulumi.com",
     "repository": {
@@ -47,7 +49,8 @@
         "prettier": "^2.8.1",
         "tslint": "^6.1.3",
         "tslint-config-prettier": "^1.18.0",
-        "typescript": "^4.6.2"
+        "typescript": "^4.6.2",
+        "yarn-deduplicate": "^6.0.2"
     },
     "pulumi": {
         "resource": true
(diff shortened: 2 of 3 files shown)


iwahbe (commented):
Looks like a sensible fix.

iwahbe (approved):

Makefile
```diff
@@ -147,7 +147,8 @@ install_java_sdk::
 	#Intentionally empty for CI / CD templating
 
 nodejs/eks/node_modules: nodejs/eks/package.json nodejs/eks/yarn.lock
-	yarn install --cwd nodejs/eks --no-progress
+	yarn install --frozen-lockfile --cwd nodejs/eks --no-progress
```
flostadler, line 150: Won't this break auto-update PRs?

The behavior of `yarn install` without `--forzen-lockfile` is to do the following:

> If yarn.lock is present and is enough to satisfy all the dependencies listed in package.json, the exact versions recorded in yarn.lock are installed, and yarn.lock will be unchanged. Yarn will not check for newer versions. If yarn.lock is absent, or is not enough to satisfy all the dependencies listed in package.json (for example, if you manually add a dependency to package.json), Yarn looks for the newest versions available that satisfy the constraints in package.json. The results are written to yarn.lock.

I'd say we want the lockfile to be updated when it cannot satisfy what's defined in `package.json`

flostadler: But how would the dependency PRs update the lockfile? with `--frozen-lockfile` this wouldn't happen.

I don't quite understand how we could reach a point where `package.json` and `yarn.lock` would drift apart. Shouldn't the worktree clean check catch those cases?
The lock file would only be regenerate if the package.json was updated without running `yarn install` and if that's done, the worktree clean checks fail.

t0yv0: In a Renovate PR the bot is updating yarn.lock by calling `yarn` without `--frozen-lockfile`, that is the update procedure is bypassing this Make target AFAIK. 

Perhaps relevant, some places where we use frozen-lockfile:

https://github.com/search?q=org%3Apulumi%20frozen-lockfile&type=code

t0yv0: You raise a good point about yarn.lock changes being detected by the worktree check. Unfortunately that does not seem to be trustworthy. First we do not do worktree checks on provider builds AFIAK but only on the SDK builds where code is expected to be generated. Then also it seems that we allow some files to drift:

```
      - name: Check worktree clean
        uses: pulumi/git-status-check-action@v1
        with:
          allowed-changes: |
            sdk/**/pulumi-plugin.json
            sdk/dotnet/Pulumi.*.csproj
            sdk/go/**/pulumiUtilities.go
            sdk/nodejs/package.json
            sdk/python/pyproject.toml
```

flostadler: Shouldn't this check everything except those few allowed changes in sdk? (I.e. include provider changes like the lockfile)

Don't get me wrong, overall I agree that `--frozen-lockfile` is the right thing to do. I'm was just concerned that it will break renovate.
I didn't know that renovate automatically updates the lockfile ([docs](https://docs.renovatebot.com/faq/)). I thought that this happens through one of our make targets.. Good to know!

t0yv0: Yeah it would still catch lockfile changes it seems. But I am not sure if it is running in the same job that builds the provider.

flostadler: It is (or at least was), but the approach with the frozen lock file is definitely better! :)

PR #1562, opened 2024-12-24
Fix ManagedNodeGroup ignoring custom AMI if no user data is configured

### LINKED ISSUE #1550: Managed node group ignores provided AMI id
### What happened?

When I create a managed node group with a amiID, worker node instance do not have same amiID.

### Example

```
const EKS_VERSION = '1.30';
const ami = pulumi.interpolate`/aws/service/eks/optimized-ami/${EKS_VERSION}/amazon-linux-2023/x86_64/nvidia/recommended/image_id`.apply((name) =>
      aws.ssm.getParameter({ name,  }, { async: true, provider: ... }),
    ).apply((result) => result.value);

const cluster = new eks.Cluster(
      `${regionalNamespace}-cluster`,
      {
        name: `${regionalNamespace}`,
        version: EKS_VERSION,
        vpcId: ...,
        privateSubnetIds: ...
        publicSubnetIds: ...,
        enabledClusterLogTypes: ['api', 'audit', 'authenticator'],
        tags: projectTags,
        endpointPrivateAccess: true,
        endpointPublicAccess: true,
        nodeAssociatePublicIpAddress: false,
        providerCredentialOpts: {
          profileName: aws.config.profile,
        },
        roleMappings: [
          ...
        ],
        instanceType: 'g5.2xlarge',
        // gpu: true,
       nodeAmiId: ami,
        nodeRootVolumeSize: 200,
        ...,
      },
      { provider: ... },
    );

    new eks.ManagedNodeGroup(
      `${regionalNamespace}-ng`,
      {
        cluster,
        amiId:  ami,
        gpu: true,
        instanceTypes: ['g5.xlarge'],
        ignoreScalingChanges: true,
        scalingConfig: {
          minSize: 1,
          maxSize: 2,
          desiredSize: DESIRED_RESOURCE_COUNT[config.env].node.desiredCapacity,
        },
        diskSize: 200,
        nodeRole: cluster.instanceRoles[0],
        enableIMDSv2: true,
        labels: {
          'transcend.io/k8s_version': '1.31'
        },
        tags: {
          ...projectTags,
          'k8s.io/cluster-autoscaler/enabled': 'true',
          [`k8s.io/cluster-autoscaler/${regionalNamespace}`]: 'owned',
        },
      },
      { provider: MAIN_REGION_PROVIDERS[mainRegion] },
    );
```

### Output of `pulumi about`

CLI          
Version      3.142.0
Go Version   go1.23.3
Go Compiler  gc

Host     
OS       debian
Version  11.7
Arch     x86_64

Backend        
Name           fv-az1490-728
URL            s3://staging-pulumi-state-io
User           root
Organizations  
Token type     personal

### Additional context

Even when I tried to hardcode amiID in managed node group, it was ignored. 

### Contributing

flostadler:
When users specified a custom AMI for an EKS Managed Node Group, the setting was only taking effect if their configuration already triggered custom user data generation, otherwise it used the default AMI. To fix this, I modified the logic to generate custom user data whenever a custom AMI is configured. This ensures the node group launches successfully with proper bootstrap settings regardless of other configuration options (AWS EKS does not provide default user data if an AMI ID is specified in the launch template).

The previous E2E tests worked because they coincidentally used the same AMI ID as the default AMI. The updated test intentionally chooses an AMI for the previous minor version and verifies the EC2 instances are using this exact AMI.

Fixes #1550

### DIFF
--- a/tests/testdata/programs/managed-ng-os/index.ts
+++ b/tests/testdata/programs/managed-ng-os/index.ts
@@ -306,8 +306,15 @@ const managedNodeGroupBottlerocketTaints = eks.createManagedNodeGroup("bottleroc
   },
 });
 
-// Create a simple AL2023 node group with a custom user data and a custom AMI
-const amiId = pulumi.interpolate`/aws/service/eks/optimized-ami/${cluster.core.cluster.version}/amazon-linux-2023/x86_64/standard/recommended/image_id`.apply(name =>
+// Fetch the previous minor version. This way we test that the user specified AMI is used instead of the default one.
+// MNGs tolerate a skew of 1 minor version
+export const customAmiId = pulumi.all([cluster.eksCluster.version]).apply(([version]) => {
+  const versionParts = version.split('.');
+  const majorVersion = parseInt(versionParts[0], 10);
+  const minorVersion = parseInt(versionParts[1], 10) - 1;
+  const versionString = `${majorVersion}.${minorVersion}`;
+  return pulumi.interpolate`/aws/service/eks/optimized-ami/${versionString}/amazon-linux-2023/x86_64/standard/recommended/image_id`;
+}).apply(name =>
   aws.ssm.getParameter({ name }, { async: true })
 ).apply(result => result.value);
 
@@ -318,6 +325,7 @@ const customUserData = userdata.createUserData({
   serviceCidr: cluster.core.cluster.kubernetesNetworkConfig.serviceIpv4Cidr,
 }, `--max-pods=${increasedPodCapacity} --node-labels=increased-pod-capacity=true`);
 
+// Create a simple AL2023 node group with a custom user data and a custom AMI
 const managedNodeGroupAL2023CustomUserData = eks.createManagedNodeGroup("al-2023-mng-custom-userdata", {
   ...baseConfig,
   operatingSystem: eks.OperatingSystem.AL2023,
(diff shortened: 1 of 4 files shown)


tests/testdata/programs/managed-ng-os/index.ts
```diff
@@ -306,8 +306,15 @@ const managedNodeGroupBottlerocketTaints = eks.createManagedNodeGroup("bottleroc
...
 });
 
-// Create a simple AL2023 node group with a custom user data and a custom AMI
-const amiId = pulumi.interpolate`/aws/service/eks/optimized-ami/${cluster.core.cluster.version}/amazon-linux-2023/x86_64/standard/recommended/image_id`.apply(name =>
+// Fetch the previous minor version. This way we test that the user specified AMI is used instead of the default one.
+// MNGs tolerate a skew of 1 minor version
+export const customAmiId = pulumi.all([cluster.eksCluster.version]).apply(([version]) => {
+  const versionParts = version.split('.');
+  const majorVersion = parseInt(versionParts[0], 10);
+  const minorVersion = parseInt(versionParts[1], 10) - 1;
```
flostadler, line 314: This would break if k8s rolled out a new major version, but there's not indications this will happen in the next years and would make this computation unnecessarily complex.
If they ever do release one we can fix up the test (it'll break and not silently succeed).

corymhall (approved):

Issue #1583, opened 2025-01-12
NodeJS SKD uses dev dependency as normal dependency


maxemann96:
`nodejs/eks/package.json` uses `jest` as a normal dependency instead of being a dev dependency. Is this intentional?

flostadler:
Hey @maxemann96, thanks a lot for bringing this to our attention! No, this is not intentional.

It's also added as a dependency in the Node.js SDK: https://github.com/pulumi/pulumi-eks/blame/6fd71265ae49a66b8389066c621e25bf77f14fd1/sdk/nodejs/package.json#L20

I'll work on fixing this.

PR #1587, opened 2025-01-15
Remove jest from nodejs dependencies

flostadler:
`jest` was wrongly added as a dependency instead of a devDependency in https://github.com/pulumi/pulumi-eks/pull/1017.
This change removes it again.

Fixes https://github.com/pulumi/pulumi-eks/issues/1583


nodejs/eks/package.json
```diff
@@ -45,6 +44,7 @@
         "@types/node": "^18.11.13",
         "@types/semver": "^7.3.10",
         "@types/which": "^1.3.1",
+        "jest": "^29.7.0",
```
flostadler, line 47: It's a dev dependency for the provider impl now. It's fully removed from the SDK.

corymhall (approved):

t0yv0 (approved):

PR #1589, opened 2025-01-16
Unpin dependencies of `TestAccOidcIam` and `TestAccCluster`

flostadler:
Renovate pins our test dependencies (`@pulumi/...`), which ultimately leads to having multiple versions of dependencies like `@pulumi/pulumi` in node_modules. This is a known situation that can cause issues (e.g. https://github.com/pulumi/pulumi/issues/13505).

Unless that core issue is resolved, I cannot see how we can pin `@pulumi/...` dependencies in tests. For now I'm gonna revert the failing tests to use version ranges, but we'll need to fix renovate.

### DIFF
--- a/examples/cluster/package.json
+++ b/examples/cluster/package.json
@@ -5,9 +5,9 @@
         "@types/node": "latest"
     },
     "dependencies": {
-        "@pulumi/pulumi": "3.144.1",
-        "@pulumi/awsx": "2.19.0",
-        "@pulumi/aws": "6.66.1",
+        "@pulumi/pulumi": "^3.144.1",
+        "@pulumi/awsx": "^2.19.0",
+        "@pulumi/aws": "^6.66.1",
         "@pulumi/eks": "latest"
     },
     "scripts": {
--- a/examples/oidc-iam-sa/package.json
+++ b/examples/oidc-iam-sa/package.json
@@ -5,9 +5,9 @@
         "@types/node": "latest"
     },
     "dependencies": {
-        "@pulumi/aws": "6.66.1",
+        "@pulumi/aws": "^6.66.1",
         "@pulumi/eks": "latest",
-        "@pulumi/kubernetes": "4.19.0",
-        "@pulumi/pulumi": "3.144.1"
+        "@pulumi/kubernetes": "^4.19.0",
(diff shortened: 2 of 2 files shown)


flostadler:
Huh, that's weird. We're still getting:
```
  pulumi:pulumi:Stack (example-cluster-p-it-fv-az1758--cluster-6f3e505c):
    warning: using pulumi-resource-eks from $PATH at /home/runner/work/pulumi-eks/pulumi-eks/bin/pulumi-resource-eks
    warning: using pulumi-resource-eks from $PATH at /home/runner/work/pulumi-eks/pulumi-eks/bin/pulumi-resource-eks
    error: Running program '/tmp/p-it-fv-az1758--cluster-6f3e505c-4089464324/' failed with an unhandled exception:
    <ref *1> TypeError: Cannot read properties of undefined (reading 'isKnown')
        at /tmp/p-it-fv-az1758--cluster-6f3e505c-4089464324/node_modules/@pulumi/runtime/invoke.ts:270:44
        at Generator.next (<anonymous>)
        at fulfilled (/tmp/p-it-fv-az1758--cluster-6f3e505c-4089464324/node_modules/@pulumi/pulumi/runtime/invoke.js:18:58)
        at processTicksAndRejections (node:internal/process/task_queues:95:5) {
      promise: Promise { <rejected> [Circular *1] }
    }
```

t0yv0 (approved):
Reverting is fine!

I think we might be able to do this https://github.com/pulumi/pulumi-awsx/pull/1493 which instructs `make renovate` to deduplicate the provider, but scale it to examples? That's unfortunate I guess. Adding example lockfiles could be ok.

flostadler:
> Reverting is fine!
> 
> I think we might be able to do this [pulumi/pulumi-awsx#1493](https://github.com/pulumi/pulumi-awsx/pull/1493) which instructs `make renovate` to deduplicate the provider, but scale it to examples? That's unfortunate I guess. Adding example lockfiles could be ok.

In this case the problem was actually a pu/pu regression that sadly manifested in the same way. That being said, while digging into this I noticed that the examples really end up having duplicate pu/pu versions. Luckily that didn't manifest in a problem yet. But It's something worth keeping in mind if it ever does

Issue #1564, opened 2024-12-26
Support Elastic Fabric Adapter in managed node groups


pierskarsenbarg:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

AWS added ability to support elastic fabric adapter in EKS: https://aws.amazon.com/about-aws/whats-new/2021/03/amazon-eks-supports-elastic-fabric-adapter/

Can we add this in to the managed node groups? Is there currently a workaround to enable this (I couldn't see anything in the base AWS provider)

flostadler:
@pierskarsenbarg The managed node groups already support EFA. This is not a node group or cluster setting (i.e. nothing configurable via AWS APIs).

You'll need to choose an instance type that has EFA support (e.g. `p5.48xlarge`), add the EFA interfaces to the launch template and then install the [`aws-efa-k8s-device-plugin` helm chart](https://artifacthub.io/packages/helm/aws/aws-efa-k8s-device-plugin).

What we could do provider side is expose a flag (e.g. `enabledEfaSupport`) or auto detect instances with EFA support and add the EFA interfaces to the launch template.

PR #1591, opened 2025-01-22
Clean up provider internals

flostadler:
This change cleans up the provider internals in preparation for more involved features like adding EFA support (https://github.com/pulumi/pulumi-eks/issues/1564). This was necessary because there was still a substantial amount of legacy cruft in the provider that wasn't even exposed to users (seem to be leftovers from moving the nodejs implementation onto MLCs).
Additionally it slightly reorganizes the internal module structure in order to have a more manageable code structure (compared to the current flat structure and HUGE files).

This change should be fully transparent to users.


flostadler:

- #1591 ◀
    - #1592

corymhall (approved):
Nice!

PR #1592, opened 2025-01-22
Use InputPropertyError for validations in the node group components

flostadler:
pu/pu introduced new error types for component authors to use. The `InputPropertyError` and `InputPropertiesError` errors are pretty printed by the engine and don't confuse users with a load of GRPC stack traces (see [docs](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/classes/InputPropertyError.html)). Additionally those errors accumulate multiple validation errors and allow presenting them all to users. Previously we exited on the first error.

For adding EFA support (https://github.com/pulumi/pulumi-eks/issues/1564) we'd like to make use of these new errors. In preparation for that I've updated the node group components to make use of this new error type.

**UX**:

Old (missing the second error):
Screenshot 2025-01-23 at 11 01 38
New:
Screenshot 2025-01-23 at 10 59 25

### DIFF
--- a/nodejs/eks/nodes/ami.ts
+++ b/nodejs/eks/nodes/ami.ts
@@ -200,7 +200,6 @@ export function toAmiType(str: string): AmiType | undefined {
 export function getOperatingSystem(
     amiType: string | undefined,
     operatingSystem: OperatingSystem | undefined,
-    parent: pulumi.Resource | undefined,
 ): OperatingSystem {
     if (operatingSystem && !amiType) {
         return operatingSystem;
@@ -210,20 +209,20 @@ export function getOperatingSystem(
 
     const resolvedAmiType = toAmiType(amiType);
     if (!resolvedAmiType) {
-        throw new pulumi.ResourceError(
-            `Cannot determine OS of unknown AMI type: ${amiType}`,
-            parent,
-        );
+        throw new pulumi.InputPropertyError({
+            propertyPath: "amiType",
+            reason: `Cannot determine OS of unknown AMI type: ${amiType}`,
+        });
     }
 
     const resolvedOs = getAmiMetadata(resolvedAmiType).os;
 
     // if users provided both OS and AMI type, we should check if they match
     if (operatingSystem && operatingSystem !== resolvedOs) {
-        throw new pulumi.ResourceError(
-            `Operating system '${operatingSystem}' does not match the detected operating system '${resolvedOs}' of AMI type '${amiType}'.`,
-            parent,
-        );
+        throw new pulumi.InputPropertyError({
+            propertyPath: "operatingSystem",
+            reason: `Operating system '${operatingSystem}' does not match the detected operating system '${resolvedOs}' of AMI type '${amiType}'.`,
+        });
     }
 
     return resolvedOs;
--- a/nodejs/eks/nodes/nodegroup.ts
+++ b/nodejs/eks/nodes/nodegroup.ts
@@ -587,9 +587,7 @@ function createNodeGroup(
 ): NodeGroupData {
     const validationErrors: pulumi.InputPropertyErrorDetails[] = [];
 
-    const instanceProfileName = core.apply((c) =>
-        resolveInstanceProfileName(name, args, c),
-    );
+    const instanceProfileName = core.apply((c) => resolveInstanceProfileName(name, args, c));
 
     if (args.clusterIngressRule && args.clusterIngressRuleId) {
         validationErrors.push({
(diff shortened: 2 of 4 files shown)


flostadler:

- #1591
    - #1592 ◀

nodejs/eks/nodes/ami.ts
```diff
@@ -210,20 +209,20 @@ export function getOperatingSystem(
 
     const resolvedAmiType = toAmiType(amiType);
     if (!resolvedAmiType) {
-        throw new pulumi.ResourceError(
-            `Cannot determine OS of unknown AMI type: ${amiType}`,
-            parent,
```
corymhall, line 215: The new `InputPropertyError` doesn't allow you to specify a `parent`, how is it able to attribute it to the correct resource? Does it not care or is it just somehow smarter than the `ResourceError`?

nodejs/eks/nodes/nodegroup.ts
```diff
@@ -572,10 +571,10 @@ export function resolveInstanceProfileName(
     } else if (c.nodeGroupOptions.instanceProfileName) {
         return pulumi.output(c.nodeGroupOptions.instanceProfileName!);
     } else {
-        throw new pulumi.ResourceError(
-            `an instanceProfile or instanceProfileName is required`,
-            parent,
-        );
+        throw new pulumi.InputPropertyError({
```
corymhall, line 574: In these cases where the error could be attributed to multiple properties should we add two property errors or should it stay as a resource error?

corymhall (commented):
Can you add screenshots of before and after error messages to the PR? I'm especially interested in what it looks like for errors that are due to the interaction between multiple properties.

nodejs/eks/nodes/ami.ts
flostadler: The errors are attached to the resource for which construct/call was invoked. So the parent gets resolved implicitly now.

nodejs/eks/nodes/nodegroup.ts
flostadler: I feel like a single input property error is enough here.
The benefit of the input property error is that a user will be presented with all errors and not only the first validation that failed.

**For example compare those two errors for the same program:**

Old:
Screenshot 2025-01-23 at 11 01 38
New:
Screenshot 2025-01-23 at 10 59 25

IMO we shouldn't create multiple errors for the same issue when there's multiple properties involved. That's just gonna increase the cognitive complexity of understanding the issue.

nodejs/eks/nodes/ami.ts
corymhall: Nice!

nodejs/eks/nodes/nodegroup.ts
corymhall: ah I see. I wasn't sure how the property errors would be shown, but since they are just listed out under the resource then that makes total sense.

It looks like that value `{{urn:pulumi:..}` is something that the engine adds? I'm assuming it only works with primitive types or something?

flostadler: That's how dependency resources are visualized here. I guess the URN is the only sensible value that can be shown here 🤔

nodejs/eks/nodes/nodegroup.ts
```diff
@@ -586,22 +584,22 @@ function createNodeGroup(
     parent: pulumi.ComponentResource,
     provider?: pulumi.ProviderResource,
 ): NodeGroupData {
-    const instanceProfileName = core.apply((c) =>
-        resolveInstanceProfileName(name, args, c, parent),
-    );
+    const validationErrors: pulumi.InputPropertyErrorDetails[] = [];
+
+    const instanceProfileName = core.apply((c) => resolveInstanceProfileName(args, c));
```
corymhall, line 589: What happens when you throw inside of an apply? Do we still get the rest of the validation errors unless they depend on `instanceProfileName`? And if we eventually do throw because of other errors do these get surfaced then?

nodejs/eks/nodes/nodegroup.ts
```diff
@@ -586,22 +584,22 @@ function createNodeGroup(
...
+            reason: `clusterIngressRule and clusterIngressRuleId are mutually exclusive`,
+        });
     }
 
     if (args.nodeSecurityGroup && args.nodeSecurityGroupId) {
-        throw new pulumi.ResourceError(
-            `invalid args for node group ${name}, nodeSecurityGroup and nodeSecurityGroupId are mutually exclusive`,
-            parent,
-        );
+        validationErrors.push({
```
corymhall, line 599: If we no longer throw here and keep going, can that lead to us being in a bad place? When we use validation errors like this, do we need to make sure we throw validation errors _before_ _any_ resources are created?

In this case I think we do throw before creating resources, but I'm just curious for future reference.

nodejs/eks/nodes/nodegroup.ts
corymhall: Just curious, does it resolve `Output` values correctly? Wondering if we ever need to be careful about what kind of values we can include here.

corymhall (approved):
Looks good, just a couple of questions for my own interest. 

Do you know if there is a guide anywhere on these new error types? Looking at the linked docs it doesn't look like there is any info on how to use (i.e. my questions on this PR)

nodejs/eks/nodes/nodegroup.ts
flostadler: You mean the message the engine adds, right? No, this one shows either that it's an output (e.g. during preview) or the resolved value if it's already resolved.

PR #1593, opened 2025-01-23
Add EFA support for Managed Node Groups

flostadler:
### Proposed changes

This change adds support for configuring [EFA](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa.html) for Managed Node Groups. EFA is a high-speed network device for instance to instance communication, usually used in AI or HPC scenarios.

The change involves adding two new input properties:
- `enableEfaSupport`
- `placementGroupAvailabilityZone`

`placementGroupAvailabilityZone` is necessary because EFA does not work cross-AZ. We also cannot pick a random AZ based on the subnets because intersection of subnets and supported AZs (by the instance) is not stable.
When AWS adds support for certain instance types to new AZs this selection would change and cause a replacement of the whole node group.

Right now I made `placementGroupAvailabilityZone` a required input when `enableEfaSupport` is set to true. This way users will not be susceptible to this foot gun. This is in line with what other tools in the eco system are doing/planning to do (e.g. https://github.com/terraform-aws-modules/terraform-aws-eks/blob/50cb230c8a1793f5ef4cc52c4c789b656b141168/modules/eks-managed-node-group/main.tf#L645-L648).

### Related issues (optional)

Closes https://github.com/pulumi/pulumi-eks/issues/1564

### DIFF
--- a/nodejs/eks/nodes/instances.test.ts
+++ b/nodejs/eks/nodes/instances.test.ts
@@ -1,0 +1,190 @@
+// Copyright 2016-2025, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as pulumi from "@pulumi/pulumi";
+
+describe("filterEfaSubnets", () => {
+    const callReturnValues = new Map<string, pulumi.runtime.MockCallResult>();
+
+    beforeAll(() => {
+        pulumi.runtime.setMocks(
+            {
+                newResource: function (args: pulumi.runtime.MockResourceArgs): {
+                    id: string;
+                    state: any;
+                } {
+                    return {
+                        id: args.inputs.name + "_id",
+                        state: args.inputs,
+                    };
+                },
+                call: function (args: pulumi.runtime.MockCallArgs): pulumi.runtime.MockCallResult {
+                    const callReturnValue = callReturnValues.get(args.token);
+                    if (callReturnValue) {
+                        return callReturnValue;
+                    }
+
+                    return args.inputs;
+                },
+            },
+            "project",
+            "stack",
+            false, // Sets the flag `dryRun`, which indicates if pulumi is running in preview mode.
+        );
+    });
+
+    let instances: typeof import("./instances");
+    beforeEach(async function () {
+        instances = await import("./instances");
+    });
+
+    it("should return filtered subnet IDs when the availability zone is supported", async () => {
+        callReturnValues.set("aws:ec2/getInstanceTypeOfferings:getInstanceTypeOfferings", {
+            locations: pulumi.output(["us-west-2a", "us-west-2b"]),
+        });
+
+        const subnetIds = ["subnet-12345", "subnet-67890", "subnet-54321"];
+
+        callReturnValues.set("aws:ec2/getSubnets:getSubnets", {
+            ids: pulumi.output([subnetIds[0]]),
+        });
+
+        const result = await promisify(
+            instances.filterEfaSubnets(subnetIds, "us-west-2a", ["t3.medium"], {}),
+        );
+
+        expect(result).toEqual([subnetIds[0]]);
+    });
+});
+
+describe("getEfaNetworkInterfaces", () => {
+    const callReturnValues = new Map<string, pulumi.runtime.MockCallResult>();
+
+    beforeAll(() => {
+        pulumi.runtime.setMocks(
+            {
+                newResource: function (args: pulumi.runtime.MockResourceArgs): {
+                    id: string;
+                    state: any;
+                } {
+                    return {
+                        id: args.inputs.name + "_id",
+                        state: args.inputs,
+                    };
+                },
+                call: function (args: pulumi.runtime.MockCallArgs): pulumi.runtime.MockCallResult {
+                    const callReturnValue = callReturnValues.get(args.token);
+                    if (callReturnValue) {
+                        return callReturnValue;
+                    }
+
+                    return args.inputs;
+                },
+            },
+            "project",
+            "stack",
+            false, // Sets the flag `dryRun`, which indicates if pulumi is running in preview mode.
+        );
+    });
+
+    let instances: typeof import("./instances");
+    beforeEach(async function () {
+        instances = await import("./instances");
+    });
+
+    it("should return network interfaces for EFA-enabled instance types with a single network card", async () => {
+        callReturnValues.set("aws:ec2/getInstanceType:getInstanceType", {
+            efaSupported: true,
+            maximumNetworkCards: 1,
+        });
+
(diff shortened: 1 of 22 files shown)


flostadler:

- #1591
    - #1592
        - #1593 ◀

tests/nodejs_test.go
```diff
@@ -1073,6 +1073,51 @@ func TestAccEksAutoModeUpgrade(t *testing.T) {
 	programTestWithExtraOptions(t, &test, nil)
 }
 
+func TestAccEfa(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
+	}
+
+	// this is one of the cheapest instances that support EFA
+	instanceType := "g6.8xlarge"
```
flostadler, line 1082: This test will cost roughly 1$ to run.
`g6.8xlarge`: $2.014 per hour, we run 2 of them for ~15 minutes

nodejs/eks/nodes/instances.ts
```diff
@@ -0,0 +1,209 @@
...
+                associatePublicIpAddress: i === 0 ? publicIpAssociation : undefined,
+                deleteOnTermination: "true",
+                networkCardIndex: i,
+                // EFA requires the first interface to have device index 0, and all the secondary interfaces to have device index 1.
+                // see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start.html#efa-start-tempinstance
+                deviceIndex: i === 0 ? 0 : 1,
+                // The primary interface must be efa, the secondary interfaces can be efa or efa-only.
+                // We choose efa-only because it prevents exhausting the private IP address space and circumvents
+                // possibly routing conflicts, like source IP mismatch, or host name mapping problems.
+                interfaceType: i === 0 ? "efa" : "efa-only",
```
flostadler, line 100: Making this configurable would yield a rather clunky API. A list of enums?
This would require users to know how many cards their instance type has.

In this case it's better if they pass a custom launch template instead.

nodejs/eks/nodes/instances.ts
```diff
@@ -0,0 +1,209 @@
...
+ *
+ * @param instanceTypes - list of EC2 instance types
+ * @returns The first instance type from the list, or "t3.medium" if no instance types are provided
+ */
+function getEfaInstanceType(
+    instanceTypes: pulumi.Input<pulumi.Input<string>[]> | undefined,
+): pulumi.Output<string> {
+    return instanceTypes
+        ? pulumi.output(instanceTypes).apply((types) => {
+              return types.length > 0 ? types[0] : DEFAULT_INSTANCE_TYPE;
```
flostadler, line 206: The other option here would be enforcing that EFA only supports homogenous node groups. IMO that's too limiting, users might wanna choose different types of similar instance types to ensure availability of those.

nodejs/eks/nodes/nodegroup.ts
```diff
@@ -1933,6 +1953,39 @@ export function createManagedNodeGroup(
...
+              name,
+              {
+                  strategy: "cluster",
+                  tags: args.tags,
+              },
+              { parent, provider },
+          )
+        : undefined;
+
+    // Require users to be explicit about which AZs they want to use when enabling EFA support or creating a placement group.
```
flostadler, line 1969: I'm curious what y'all think about this. I added my reasoning for this to the PR description as well. This prevents a major foot gun

nodejs/eks/nodes/nodegroup.ts
```diff
@@ -1933,6 +1953,39 @@ export function createManagedNodeGroup(
         });
     }
 
+    // Create a placement group if EFA support is enabled
+    const createPlacementGroup = args.enableEfaSupport === true;
+    const placementGroup = createPlacementGroup
```
corymhall, line 1958: Can this just be 

```ts
const placementGroup = args.enableEfaSupport ? new aws.ec2.PlacementGroup() : undefined;
```

nodejs/eks/nodes/nodegroup.ts
corymhall: I agree

nodejs/eks/nodes/instances.test.ts
```diff
@@ -0,0 +1,192 @@
...
+        );
+    });
+
+    let instances: typeof import("./instances");
+    beforeEach(async function () {
+        instances = await import("./instances");
+        callReturnValues.clear();
+    });
+
+    it("should return filtered subnet IDs when the availability zone is supported", async () => {
```
corymhall, line 53: I can't remember, does our mocking library allow for us to catch and assert on errors?

corymhall (approved):

nodejs/eks/nodes/nodegroup.ts
```diff
@@ -1789,6 +1790,16 @@ export type ManagedNodeGroupOptions = Omit<
      * - [Amazon EKS AMI - Nodeadm Configuration](https://awslabs.github.io/amazon-eks-ami/nodeadm/doc/api/).
      */
     nodeadmExtraOptions?: pulumi.Input<pulumi.Input<NodeadmOptions>[]>;
+
+    /**
+     * The availability zone for the placement group.
+     */
+    placementGroupAvailabilityZone?: pulumi.Input<string>;
```
t0yv0, line 1797: Is placementGroupAvailabilityZone useful separately from enableEfaSupport?

If not perhaps we could force it to be required at type level:

```
efaSupport: {
  enable: true,
  placementGroupAvailabilityZone: "us-west-1b"
}
```

nodejs/eks/nodes/nodegroup.ts
t0yv0: Have to trust you on this one, do not have a good enough mental picture of the implications here.

t0yv0 (approved):

nodejs/eks/nodes/nodegroup.ts
flostadler: I thought about that, but decided against it because we might add placement group support separately in the future. Tying it directly to EFA might maneuver us in a bad spot for the future

nodejs/eks/nodes/instances.test.ts
flostadler: Yes, exactly! I was almost going crazy until I remembered that @t0yv0 was having issues here recently as well (https://github.com/pulumi/pulumi/issues/16665).

Had to resort to testing those errors with an e2e test :(

nodejs/eks/nodes/nodegroup.ts
flostadler: It could, but I wanted to make the whole placementGroup logic more explicit (we're also using the `createPlacementGroup` const further down).

But I'm also happy to change it if it's more understandable that way!

nodejs/eks/nodes/instances.ts
```diff
@@ -0,0 +1,209 @@
...
+        .all([instanceTypeInfo.efaSupported, instanceType])
+        .apply(([efaSupported, instanceType]) => {
+            if (!efaSupported) {
+                throw new pulumi.InputPropertyError({
+                    propertyPath: args.instanceTypePropertyPath,
+                    reason: `The selected instance type '${instanceType}' does not support EFA.`,
+                });
+            }
+            return instanceTypeInfo.maximumNetworkCards.apply((cards) => {
+                return cards > 1 ? cards : 1;
```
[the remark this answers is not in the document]
flostadler: It's modeled as a `number`, but I couldn't find any guarantee on the AWS side that this always has a sensible value (i.e. non-zero).

Just wanted to safeguard this. Every EC2 instance should have at least a single network card

PR #1596, opened 2025-01-24
Onboard ci-mgmt

rquitales:
### Proposed changes

This PR revives #1474 by rebasing and modifying to fit the latest changes to pulumi-eks and ci-mgmt.

Changes made:

- Restructure the go modules to align with bridged providers for now
- Generated ci-mgmt workflows
- Added kubectl installation as a setup step
- Runs all tests in parallel across 25 shards (including upgrade tests)

Passing alpha release: https://github.com/pulumi/pulumi-eks/actions/runs/13015734092


tests/testdata/programs/ignore-scaling-changes/index.ts
```diff
@@ -61,5 +61,4 @@ const ngV2 = new eks.NodeGroupV2("ignore-scaling-config-ngv2", {
   operatingSystem: eks.OperatingSystem.AL2023,
   instanceProfile: new aws.iam.InstanceProfile("ignore-scaling-config-ngv2", {role: ngV2Role}),
   ignoreScalingChanges: true,
-  subnetIds: eksVpc.privateSubnetIds,
```
flostadler, line 64: Why was this changed? We're intentionally using private subnets in order to not exhaust the EIPs (we've had many test failures in the past caused by exhausting our EIP quota).

.pulumi/version
```diff
@@ -0,0 +1 @@
+3.147.0
```
t0yv0, line 1: I think this is possibly auto-computed and should not be checked in? WDYT?

provider/go.mod
```diff
@@ -0,0 +1,126 @@
+module github.com/pulumi/pulumi-eks/provider/v3
```
t0yv0, line 1: Curious your thoughts here, we were talking about how the preferred end-goal is one go.mod at the top level of the tree, but bridged template does not yet support it as it insists on the provider/go.mod location. Do you think this is a permanent or temporary addition?

provider/go.mod
```diff
@@ -0,0 +1,126 @@
...
+
+go 1.23
+
+toolchain go1.23.4
+
+replace github.com/pulumi/pulumi-eks/sdk/v3 => ../sdk
```
t0yv0, line 7: I'm curious why/how these Go files depend on the SDK for the EKS provider? Does it not create some circular dependency in the sense? Surprised. I thought all these files do is wrapping schema generation and schema-embed compression.

tests/testdata/programs/ignore-scaling-changes/index.ts
t0yv0: Indeed this is a bit of an odd one. Separate PR perhaps with explanation? Thanks!

provider/go.mod
rquitales: This would be a temporary reversal until we update all providers in unison. This scopes the work to onboard ci-mgmt without created additional logic branches.

provider/go.mod
rquitales: It actually doesn't. The replace directive isn't needed since we aren't actually using `github.com/pulumi/pulumi-eks/sdk/v3 `. It was included as I lifted the go.mod file as is. I can clean this up in a follow-up PR.

tests/testdata/programs/ignore-scaling-changes/index.ts
rquitales: The test fails with:
```
stderr: error: Running program '/tmp/TestIgnoringScalingChanges1893918203/001/ignore-scaling-changes/' failed with an unhandled exception:
        TSError: ⨯ Unable to compile TypeScript:
        index.ts(64,3): error TS2345: Argument of type '{ minSize: number; maxSize: number; desiredCapacity: number; cluster: eks.Cluster; instanceType: string; operatingSystem: "AL2023"; instanceProfile: InstanceProfile; ignoreScalingChanges: true; subnetIds: pulumi.Output<...>; }' is not assignable to parameter of type 'NodeGroupV2Args'.
          Object literal may only specify known properties, and 'subnetIds' does not exist in type 'NodeGroupV2Args'.
```

Looking at our current CI setup, this test does not actually run currently, which is probably why it wasn't surfaced. I can separate the test fix out in another PR to be merged first, to ensure that this test won't block ci-mgmt onboarding.

t0yv0: Yeah you can comment it with a t.Skip() and we'll take care of it later.

Issue #1601, opened 2025-01-27
`TestIgnoringScalingChanges` fails when run


rquitales:
### What happened?

The following error occurs when `TestIgnoringScalingChanges` is run.

```shell
stderr: error: Running program '/tmp/TestIgnoringScalingChanges1893918203/001/ignore-scaling-changes/' failed with an unhandled exception:
        TSError: ⨯ Unable to compile TypeScript:
        index.ts(64,3): error TS2345: Argument of type '{ minSize: number; maxSize: number; desiredCapacity: number; cluster: eks.Cluster; instanceType: string; operatingSystem: "AL2023"; instanceProfile: InstanceProfile; ignoreScalingChanges: true; subnetIds: pulumi.Output<...>; }' is not assignable to parameter of type 'NodeGroupV2Args'.
          Object literal may only specify known properties, and 'subnetIds' does not exist in type 'NodeGroupV2Args'.
```

From our documentation, `subnetIds` does not exist in `NodeGroupV2Args`, so it should be removed. Note that our current CI setup does not run upgrade tests, which is why this error was not observed in CI.

### Example

`TestIgnoringScalingChanges`

### Output of `pulumi about`

-

### Additional context

Surfaced in: #1596

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
Verified the test passes in CI: https://github.com/pulumi/pulumi-eks/actions/runs/12999614400/job/36255459840?pr=1596

PR #1602, opened 2025-01-27
Remove non-existant `subnetIds` field from `TestIgnoringScalingChanges`

rquitales:
### Proposed changes

Fixes the `TestIgnoringScalingChanges`. Previously, the following ` Object literal may only specify known properties, and 'subnetIds' does not exist in type 'NodeGroupV2Args'.` error occurs when running the upgrade test.

### Related issues (optional)

Fixes: #1601

### DIFF
--- a/tests/testdata/programs/ignore-scaling-changes/index.ts
+++ b/tests/testdata/programs/ignore-scaling-changes/index.ts
@@ -61,4 +61,5 @@ const ngV2 = new eks.NodeGroupV2("ignore-scaling-config-ngv2", {
   operatingSystem: eks.OperatingSystem.AL2023,
   instanceProfile: new aws.iam.InstanceProfile("ignore-scaling-config-ngv2", {role: ngV2Role}),
   ignoreScalingChanges: true,
+  nodeSubnetIds: eksVpc.privateSubnetIds,
 });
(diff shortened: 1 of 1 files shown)


tests/testdata/programs/ignore-scaling-changes/index.ts
```diff
@@ -61,5 +61,4 @@ const ngV2 = new eks.NodeGroupV2("ignore-scaling-config-ngv2", {
   operatingSystem: eks.OperatingSystem.AL2023,
   instanceProfile: new aws.iam.InstanceProfile("ignore-scaling-config-ngv2", {role: ngV2Role}),
   ignoreScalingChanges: true,
-  subnetIds: eksVpc.privateSubnetIds,
```
flostadler, line 64: It’d be great if we could change this to `nodeSubnetIds` instead. That way those nodes will not consume EIPs

flostadler (approved):

tests/testdata/programs/ignore-scaling-changes/index.ts
rquitales: I can update this PR. Wasn't sure what the best fallback option here was when updating the test.

Issue #1597, opened 2025-01-24
EKS auto mode with ComputeConfig set yields terraform error


baughj:
### What happened?

Attempting to set `ComputeConfig` within auto mode configuration to enable a custom node role yields (what appears to be) a Terraform error on `pulumi up`:

```
Diagnostics:
  aws:eks:Cluster (eks-cluster-eksCluster):
    error:   sdk-v2/provider2.go:515: sdk.helper_schema: compute_config.enabled, kubernetes_networking_config.elastic_load_balancing.enabled, and storage_config.block_storage.enabled must all be set to either true or false: provider=aws@6.66.1
    error: Preview failed: diffing urn:pulumi:dev::eks-repro::eks:index:Cluster$aws:eks/cluster:Cluster::eks-cluster-eksCluster: 1 error occurred:
        * compute_config.enabled, kubernetes_networking_config.elastic_load_balancing.enabled, and storage_config.block_storage.enabled must all be set to either true or false
```

### Example

This is the smallest repro that will yield the error:

```csharp

using Pulumi;
using Pulumi.Aws.Iam;
using Pulumi.Eks;
using System.Collections.Generic;
using Pulumi.Eks.Inputs;

return await Deployment.RunAsync(() =>
{

    var assumePolicy =
        """
        {
          "Version": "2012-10-17",
          "Statement": [ 
            { 
              "Effect": "Allow",
              "Principal": { 
                "Service": "ec2.amazonaws.com"
              },
              "Action": "sts:AssumeRole"
             }
             ]
        }
        """;

    var nodeRole = new Role("eks-node-role", new RoleArgs
    {
        Name = "eks-node-role",
        AssumeRolePolicy = assumePolicy,
        Description = "EKS node role",
        ManagedPolicyArns =
        [
            "arn:aws:iam::aws:policy/AmazonEKSWorkerNodeMinimalPolicy",
            "arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryPullOnly"
        ]
    });

    var cluster = new Cluster("eks-cluster", new ClusterArgs
    {
        AuthenticationMode = AuthenticationMode.Api,
        AutoMode = new AutoModeOptionsArgs
        {
            Enabled = true,
            CreateNodeRole = false,
            ComputeConfig = new ClusterComputeConfigArgs { NodeRoleArn = nodeRole.Arn }
        }
    });
});
```

On `pulumi up`, this gives the error above.

### Output of `pulumi about`

```
[...]

PR #1603, opened 2025-01-28
Fix EKS autoMode preview with unknown values

flostadler:
The diff customizer of the upstream provider is not taking possibly unknown values into account (see https://github.com/hashicorp/terraform-provider-aws/blob/ae93494f39ba70fe442e891caf05f8df21bde1ac/internal/service/eks/cluster.go#L1776-L1791), which causes failures like this one:
```
* compute_config.enabled, kubernetes_networking_config.elastic_load_balancing.enabled, and storage_config.block_storage.enabled must all be set to either true or false
```
This happens because `unknown` attributes just default to their empty values in a diff customizer if they're unknown .

This change acts a hot fix until we can properly fix upstream's behavior here. The current diff for auto mode is quite broken and needs some more involved fixes (see https://github.com/hashicorp/terraform-provider-aws/issues/40582).

Fixes https://github.com/pulumi/pulumi-eks/issues/1597

### DIFF
--- a/nodejs/eks/cluster/cluster.ts
+++ b/nodejs/eks/cluster/cluster.ts
@@ -642,19 +642,42 @@ export function createCore(
         );
     }
 
-    let computeConfig: pulumi.Input<aws.types.input.eks.ClusterComputeConfig> | undefined;
+    let computeConfig: pulumi.Output<aws.types.input.eks.ClusterComputeConfig> | undefined;
     if (args.autoMode?.enabled) {
-        computeConfig = pulumi.output(args.autoMode.computeConfig).apply((config) => {
-            return {
-                enabled: true,
-                nodeRoleArn: eksAutoNodeRole?.directRole.arn,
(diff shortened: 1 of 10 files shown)


nodejs/eks/cluster/cluster.ts
```diff
@@ -642,20 +642,43 @@ export function createCore(
...
     }
 
     const storageConfig = args.autoMode?.enabled ? { blockStorage: { enabled: true } } : undefined;
 
+    // The upstream EKS cluster resource requires that all auto mode settings
+    // are either enabled or disabled and does not handle unknown values.
+    // Make all eks auto mode options depend on the other options to ensure
+    // that they're all either known or unknown when auto mode is enabled.
+    const autoModeConfig =
+        args.autoMode?.enabled && computeConfig && storageConfig && kubernetesNetworkConfig
```
t0yv0, line 666: Oof.

t0yv0 (approved):

PR #1612, opened 2025-01-30
Revert "Update first-party Pulumi dependencies"

t0yv0:
Reverts pulumi/pulumi-eks#1576

This PR did not automatically update schema and re-generate SDKs but defeated CI checks. Reverting temporarily, will test out the fixes later on that should enable Renovate to build these PRs properly.


rquitales (approved):

rquitales:
Note: local build of this PR verified that there are no changes (just in case CI is broken).

PR #1632, opened 2025-02-07
Allow `pulumi.Output` in EKS Addon `configurationValues`

Makeshift:
### Proposed changes

Someone went through the effort of creating a stringify function and forgot to use it 😁 

**Edit by @guineveresaenger:** To preserve ordering I took the author's suggestion and used `pulumi.jsonStringify`. This should fix up the PR enough to be mergeable. If a reviewer would like ordering-specific testing, we can do so; however this is a minor enough refactor that I do not think explicit test is warranted.

### DIFF
--- a/nodejs/eks/addons/addon.ts
+++ b/nodejs/eks/addons/addon.ts
@@ -24,7 +24,7 @@ import { Cluster } from "../cluster";
 export interface AddonOptions
     extends Omit<aws.eks.AddonArgs, "resolveConflicts" | "clusterName" | "configurationValues"> {
     cluster: Cluster;
-    configurationValues?: object;
+    configurationValues?: pulumi.Input<object>;
 }
 
 /**
@@ -47,7 +47,7 @@ export class Addon extends pulumi.ComponentResource {
             {
                 ...args,
                 clusterName: cluster.core.cluster.name,
-                configurationValues: JSON.stringify(args.configurationValues),
+                configurationValues: pulumi.output(args.configurationValues).apply(JSON.stringify),
             },
             { parent: this, provider: opts?.provider },
         );
--- a/nodejs/eks/addons/addon.ts
+++ b/nodejs/eks/addons/addon.ts
@@ -47,7 +47,7 @@ export class Addon extends pulumi.ComponentResource {
             {
                 ...args,
                 clusterName: cluster.core.cluster.name,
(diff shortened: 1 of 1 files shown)


nodejs/eks/addons/addon.ts
```diff
@@ -24,7 +24,7 @@ import { Cluster } from "../cluster";
 export interface AddonOptions
     extends Omit<aws.eks.AddonArgs, "resolveConflicts" | "clusterName" | "configurationValues"> {
     cluster: Cluster;
-    configurationValues?: object;
+    configurationValues?: pulumi.Input<object>;
```
rquitales, line 27: Note: This should be a safe change as the API is currently modeled as an Input already ([ref](https://github.com/pulumi/pulumi-eks/blob/78eeb0b7a55be5ae05138c801b1b689d1f103003/provider/cmd/pulumi-resource-eks/schema.json#L618-L623)) so this isn't a breaking change.

nodejs/eks/addons/addon.ts
```diff
@@ -37,17 +37,16 @@ export class Addon extends pulumi.ComponentResource {
...
-        });
+        }));
 
         const addon = new aws.eks.Addon(
             name,
             {
                 ...args,
                 clusterName: cluster.core.cluster.name,
-                configurationValues: JSON.stringify(args.configurationValues),
+                configurationValues: stringifyAddonConfiguration(args.configurationValues),
```
rquitales, line 49: We might need to be more careful here. `stringifyAddonConfiguration` changes the ordering, and may cause replacements for existing users. We'll need to test this out.

rquitales (commented):
Thanks for submitting this PR and helping clean up our codebase! Overall the PR looks like a great addition. I'll test this change on the side and merge when possible.

nodejs/eks/addons/addon.ts
Makeshift: Indeed! The types were telling me that an `Input` should be fine but then it put `Calling [toJSON] on an [Output<T>] is not supported` in my JSON blob

Makeshift:
I was just doing some refactoring of my component that handles EKS addons and ran into this issue again. I don't think it's been fixed elsewhere. @rquitales could we have a look at this again?

guineveresaenger:
@pulumi-eon please review

guineveresaenger:
@pulumi-eon rereview

corymhall (approved):

Issue #1799, opened 2025-06-17
Pulumi Preview/Up fails with EKS provider in Deployment Pipelines


minato7:
### Describe what happened

Hi,

We are experiencing issue during pulumi preview/up with usage of EKS provider in Azure Deployment Pipeline. This is not reproducible in local setup but only in the pipeline. 

Any help to debug this issue would be appreciated. We tried different ways to preview but just fails with ToString error. Our usage is quite straightforward with a EKS cluster creation.

### Sample program

Dotnet program with EKS provider and EKS cluster creation in Azure Deployment Pipeline. Tested this with multiple programs in different pipelines, same error message in all of them.

### Log output

```
pulumi up --diff --non-interactive --yes --refresh=False 
Previewing update (sample):
  pulumi:pulumi:Stack: (sample)
error: eks:index:Cluster resource 'diagnostics-eks-cluster' has a problem: Calling [toString] on an [Output<T>] is not supported.

To get the value of an Output<T> as an Output<string> consider either:
1: o.apply(v => `prefix${v}suffix`)
2: pulumi.interpolate `prefix${v}suffix`

See https://www.pulumi.com/docs/concepts/inputs-outputs for more details.

Or use ESLint with https://github.com/pulumi/eslint-plugin-pulumi to warn or
error lint on using Output<T> in template literals.
error: Error: Calling [toString] on an [Output<T>] is not supported.

To get the value of an Output<T> as an Output<string> consider either:
1: o.apply(v => `prefix${v}suffix`)
2: pulumi.interpolate `prefix${v}suffix`

See https://www.pulumi.com/docs/concepts/inputs-outputs for more details.

Or use ESLint with https://github.com/pulumi/eslint-plugin-pulumi to warn or
error lint on using Output<T> in template literals.
    at Proxy.OutputImpl.toString (/snapshot/eks/node_modules/@pulumi/pulumi/output.js:85:23)
    at Array.join (<anonymous>)
    at Array.toString (<anonymous>)
    at Object.transferProperties (/snapshot/eks/node_modules/@pulumi/pulumi/runtime/rpc.js:95:89)
    at /snapshot/eks/node_modules/@pulumi/pulumi/runtime/resource.js:576:33
    at Generator.next (<anonymous>)
    at /snapshot/eks/node_modules/@pulumi/pulumi/runtime/resource.js:21:71
    at new Promise (<anonymous>)
[...]

PR #1803, opened 2025-06-20
Replace tslint with eslint and fix a missing interpolate

blampe:
In https://github.com/pulumi/pulumi-eks/issues/1799 a user reported an error message surfacing from TS which indicated a missing interpolation.

I wanted to use https://github.com/pulumi/eslint-plugin-pulumi to track down the source of the problem, but this repo is still using (deprecated) TSLint.

So I upgraded us to eslint and dropped in the pulumi plugin, which immediately identified a problem with our invalid subnets error.

Fixes https://github.com/pulumi/pulumi-eks/issues/1799.


flostadler (approved):
Good catch!

mjeffryes (approved):

PR #1960, opened 2025-11-03
add support for EKS deletion protection

schnipseljagd:
### Proposed changes

Add support for EKS deletion protection.
this requires an update of the pulumi AWS provider.

PR #1976, opened 2025-11-14
Add support for EKS deletion protection

### LINKED ISSUE #1961: add support for EKS deletion protection
PR: https://github.com/pulumi/pulumi-eks/pull/1960

pose:
Fixes #1961

This PR builds on top of PR #1960  authored by @schnipseljagd. I verified the changes by deploying an EKS cluster (test included as part of the PR), introduced deletion protection to the Pulumi EKS Cluster's schema and I re-generated all SDK. On top of that, I made sure that @schnipseljagd is properly attributed as the first commit author.

### DIFF
--- a/tests/nodejs_test.go
+++ b/tests/nodejs_test.go
@@ -1161,7 +1161,9 @@ func TestAccAutoModePreview(t *testing.T) {
 // However, since it creates a cluster that cannot be deleted, it is skipped
 // by default and only meant to be run manually.
 func TestAccDeletionPolicy(t *testing.T) {
-	t.Skip("skipping since it creates a cluster that cannot be deleted.")
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
+	}
 	test := getJSBaseOptions(t).
 		With(integration.ProgramTestOptions{
 			Dir: path.Join(getExamples(t), "deletion-policy"),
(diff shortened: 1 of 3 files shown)


tests/nodejs_test.go
```diff
@@ -1157,6 +1157,24 @@ func TestAccAutoModePreview(t *testing.T) {
 	t.Logf("Preview:\n%s", previewResult.StdOut)
 }
 
+// TestAccDeletionPolicy tests that the EKS cluster can be deleted.
+// However, since it creates a cluster that cannot be deleted, it is skipped
+// by default and only meant to be run manually.
+func TestAccDeletionPolicy(t *testing.T) {
+	t.Skip("skipping since it creates a cluster that cannot be deleted.")
```
corymhall: There are a couple of things we could do to test this one out. The one that might be the easiest is to use `EditDirs` to run multiple versions of the program. then we could do something like:

```go
test := getJSBaseOptions(t).
		With(integration.ProgramTestOptions{
			Dir: path.Join(getExamples(t), "deletion-policy"),
			ExtraRuntimeValidation: func(t *testing.T, info integration.RuntimeValidationStackInfo) {
				utils.ValidateClusters(t, info.Deployment.Resources, utils.WithKubeConfigs(
					info.Outputs["kubeconfig1"],
				))
			},
            EditDirs: []integration.EditDir{
				{
					// Run the new program which has the cluster removed from the code.
					Dir:             path.Join(getExamples(t), "deletion-policy", "step2"),
					ExpectFailure: true, // we expect the up to fail because we can't remove the cluster
					Additive:        true,
				},
				{
					// We run it again with the cluster back in the code and disable deletion protection
                    // so it can be deleted by the test.
					Dir:             path.Join(getExamples(t), "deletion-policy", "step3"),
					Additive:        true,
				},
			},
		})
```

We could also potentially use `ExtraRuntimeValidation` within the first `EditDir` to validate the error.

PR #2267, opened 2026-05-26
Reconcile ManagedNodeGroupOptions amiType variance with node group args

pose:
## Summary

The nodejs SDK build fails at `make build_nodejs` with TS2345 in `nodegroupMixins.ts`, where `createManagedNodeGroup` passes a `ManagedNodeGroupOptions` into `new ManagedNodeGroup(...)`. `ManagedNodeGroupOptions` derives its node group fields from `aws.eks.NodeGroupArgs`, whose optional scalars are now typed `pulumi.Input<T | undefined>` (the `undefined` lives inside the `Input`). The generated `ManagedNodeGroupArgs` types the same fields as `pulumi.Input<T> | undefined` (the `undefined` lives outside). The two forms are equivalent at runtime, since an `Input` that resolves to `undefined` is treated the same as an unset field, but they are not mutually assignable, so the type checker rejects the assignment. The error first surfaces on `amiType` and, once that is addressed in isolation, recurs on `capacityType` and the other optional scalars, so this is a single variance issue at the construction boundary rather than a per-field one.

## Fix

Reconcile the variance once at the construction boundary with an explicit assertion to `ManagedNodeGroupArgs`, documented inline, instead of narrowing each affected field with `Omit`. Narrowing per field would need to chase every optional scalar and would silently drop coverage whenever upstream adds another. The change is applied to both the codegen template (`provider/cmd/pulumi-gen-eks/nodejs-templates/nodegroupMixins.ts`) and the generated SDK copy (`sdk/nodejs/nodegroupMixins.ts`) so they stay in sync.

An alternative would be to reconcile this at the codegen/schema layer so the generated args adopt the `Input<T | undefined>` shape directly. The boundary assertion is the smaller, runtime-safe change; happy to redirect to the codegen layer if reviewers prefer that.

Verified by running `tsc --noEmit` on the nodejs SDK with `@pulumi/aws ^7.14.0` installed: it failed with the reported TS2345 before the change and passes after.

Part of pulumi/pulumi-eks#2266
Part of pulumi/pulumi-eks#2243

### DIFF
--- a/nodejs/eks/nodes/nodegroup.ts
+++ b/nodejs/eks/nodes/nodegroup.ts
@@ -836,6 +836,15 @@ function createNodeGroup(
         aws.ec2.getAmi(
             {
                 owners: ["self", "amazon"],
+                // Resolve deprecated AMIs too. AWS auto-deprecates every AMI two
+                // years after creation, and EKS-optimized AMIs are additionally
+                // deprecated as their Kubernetes version ages out, but a
+                // deprecated AMI is still available and launchable. When a user
+                // pins nodeAmiId to such an image the image-id filter must still
+                // find it; without includeDeprecated this lookup returns no
+                // results once the AMI is deprecated and node group creation
+                // fails even though the AMI is perfectly usable.
+                includeDeprecated: true,
                 filters: [
                     {
                         name: "image-id",
@@ -1317,6 +1326,15 @@ export function createNodeGroupV2(
         aws.ec2.getAmi(
             {
                 owners: ["self", "amazon"],
+                // Resolve deprecated AMIs too. AWS auto-deprecates every AMI two
+                // years after creation, and EKS-optimized AMIs are additionally
+                // deprecated as their Kubernetes version ages out, but a
+                // deprecated AMI is still available and launchable. When a user
+                // pins nodeAmiId to such an image the image-id filter must still
+                // find it; without includeDeprecated this lookup returns no
+                // results once the AMI is deprecated and node group creation
+                // fails even though the AMI is perfectly usable.
+                includeDeprecated: true,
                 filters: [
                     {
                         name: "image-id",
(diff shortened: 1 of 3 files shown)


corymhall (approved):

pose:
Added a second, standalone commit (`Resolve a pinned node AMI after AWS deprecates it`) for a separate root cause that was also red on this PR's `test` checks and is independently red on `master`.

**What was failing:** the `test (25, 6)` shard failed `TestEksClusterInputValidations/{instanceRoles,roleMappings,userMappings}_0` with `invoking aws:ec2/getAmi:getAmi: Your query returned no results`. Those `_0` subcases expect the preview to succeed, so the lookup failure failed them; the matching `_1` subcases kept passing because their input-validation error fires before AMI resolution is reached.

**Root cause (independent of this PR's type change):** the test program pins `nodeAmiId: ami-0384725f0d30527c7`, the EKS 1.29 Amazon Linux 2 image created 2024-05-23. AWS auto-deprecates every AMI two years after its creation date, so this one was deprecated on 2026-05-23, three days before the CI run. The provider resolves a pinned `nodeAmiId` via `aws.ec2.getAmi` with an `image-id` filter scoped to `owners: ["self", "amazon"]`, and AWS `DescribeImages` omits deprecated AMIs from filter-based queries unless `includeDeprecated` is set. So a still-available, launchable AMI returned no results and node group creation failed. `aws ec2 describe-images --image-ids ami-0384725f0d30527c7` confirms `State: available`, `DeprecationTime: 2026-05-23T05:59:51Z`.

**Fix:** set `includeDeprecated: true` on both `getAmi(image-id)` lookups (`createNodeGroup` and `createNodeGroupV2`). This is the regression it prevents: a pinned `nodeAmiId` otherwise stops resolving roughly two years after that AMI is published, once auto-deprecation removes it from filter-based results, even though the AMI is still available and launchable.

Kept as a standalone commit (no shared issue refs) because it is a distinct concern from the `Input<T | undefined>` variance fix, but bundled into this PR since both currently block `master`.

The other two red shards, `test (25, 2)` and `test (25, 7)`, are unrelated infrastructure flakes (resource plugins "exited prematurely / did not begin responding to RPC connections", and an OIDC token-exchange timeout reaching the JWKS endpoint). They should clear on re-run.

PR #2368, opened 2026-08-26
Support advanced Kubernetes control plane configuration

### LINKED ISSUE #2367: Support EKS advanced Kubernetes control plane configuration
## Hello!

## Issue details

AWS launched [advanced Kubernetes control plane configuration](https://docs.aws.amazon.com/eks/latest/userguide/control-plane-configuration.html) in August 2026. It lets you set kube-scheduler, kube-apiserver and kube-controller-manager parameters on a cluster — for example `nodeResourcesFit.scoringStrategy` to `MostAllocated`, so the scheduler packs pods onto fewer nodes and node pools that support consolidation can remove the ones left empty.

`aws.eks.Cluster` gained `kubeSchedulerConfig`, `kubeApiServerConfig` and `kubeControllerManagerConfig` in `@pulumi/aws` 7.42.0, but `eks.Cluster` has no passthrough for them. Clusters created through this component can only get these settings by dropping down to the raw `aws.eks.Cluster` resource or by applying them out of band with the AWS CLI, where Pulumi no longer describes the cluster's configuration.

I'd like to add the three inputs, following the same shape as `upgradePolicy` (#1776 / #1787): optional, no default, passed straight through.

I have this working locally and will open a PR.

### Affected area/feature
https://www.pulumi.com/registry/packages/eks/api-docs/cluster/#inputs

ranyhb:
### Proposed changes

Adds `kubeSchedulerConfig`, `kubeApiServerConfig` and `kubeControllerManagerConfig` to `eks.Cluster`, passed straight through to `aws.eks.Cluster`. This exposes AWS's [advanced Kubernetes control plane configuration](https://docs.aws.amazon.com/eks/latest/userguide/control-plane-configuration.html) to clusters created through this component.

All three are optional with no default, so leaving them unset sends nothing and AWS's own defaults apply — no change for existing users. I followed `upgradePolicy` (#1787) file-for-file, including exposing them on `ClusterResult`.

These types only exist in `@pulumi/aws` 7.42.0+, so this bumps `nodejs/eks` from 7.25.0 to 7.43.0 (and `sdk/go.mod` to match, or the Go SDK doesn't compile). That bump surfaces 7 pre-existing type errors in `nodegroup.ts`, because optional list and string inputs are now typed `Input<T | undefined>` rather than `Input<T> | undefined`. **This is the same breakage currently blocking #2209**, so this PR should unblock that too. I fixed them by widening the affected signatures and guarding the resolved values rather than casting, so runtime behaviour is unchanged — with one deliberate exception, called out in its own commit: in `getRecommendedAMI`, `version` now falls back with `version ?? clusterVersion`. Previously an `args.version` that *resolved* to `undefined` was passed straight into `ssmParameterName`, producing a malformed SSM parameter path.

Regenerating the schema also flips 57 `/aws/v7.14.0` refs to `v7.43.0`. That's pre-existing drift the regen picks up, purely mechanical.

The example sets `kubeSchedulerConfig` only. `kubeControllerManagerConfig` requires EKS Provisioned Control Plane, so it would fail on the test cluster.

Split into three commits — the `nodegroup.ts` type fixes, the dependency bump, then the feature plus regenerated schema and SDKs — in case you'd rather take them separately.

Verified locally: `tsc` clean, 417/417 unit tests with snapshots, `make build_nodejs`, `go build ./go/...`, `make provider` and the provider tests. dotnet, java and python SDKs are regenerated but I can't build those here. I did not run the integration tests, since they create real EKS clusters.

One thing to watch: regenerating the nodejs SDK on my machine rewrites the `sdk/nodejs/types/*.ts` imports from `../utilities` to `./utilities`, which doesn't compile. I reproduced that on an unmodified `master`, so it isn't coming from this change, and I reverted those files (along with `sdk/java/settings.gradle` and the version stamp in `build.gradle`) to keep the diff to what this change actually needs. If `build_sdk (nodejs)` disagrees, take CI's version of those files.

### Related issues

Fixes #2367

### DIFF
--- a/examples/cluster/index.ts
+++ b/examples/cluster/index.ts
@@ -55,6 +55,13 @@ const cluster3 = new eks.Cluster(`${projectName}-3`, {
     upgradePolicy: {
         supportType: "STANDARD",
     },
+    kubeSchedulerConfig: {
+        nodeResourcesFit: {
+            scoringStrategy: {
+                type: "MostAllocated",
+            },
+        },
+    },
 })
 
 // cluster4 is a graviton cluster to test the ARM64 architecture
--- a/nodejs/eks/cluster/cluster.ts
+++ b/nodejs/eks/cluster/cluster.ts
@@ -740,6 +740,9 @@ export function createCore(
                   }
                 : undefined,
             upgradePolicy: args.upgradePolicy,
+            kubeApiServerConfig: args.kubeApiServerConfig,
+            kubeControllerManagerConfig: args.kubeControllerManagerConfig,
+            kubeSchedulerConfig: args.kubeSchedulerConfig,
             deletionProtection: args.deletionProtection,
         },
         {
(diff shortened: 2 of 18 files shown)


ranyhb:
@pose @guineveresaenger — could you take a look when you get a chance? Thanks!

ranyhb:
@iwahbe — I know you're an active maintainer across the Pulumi repos, and I saw your recent contributions here too. Could you take a look as well? Thanks!

ranyhb:
@Frassle is there any chance that u can take a look ?

Frassle:
EKS falls under the @pulumi/providers team, and they’ll review this PR when they have capacity. Please don’t ping individual maintainers directly to request attention on a PR. That creates unnecessary pressure on contributors and maintainers, and is not an appropriate way to expedite review. The owning team will triage the PR through the normal process.

ranyhb:
Closing this in favour of splitting it up, as the three commits here were doing quite different things.

The dependency bump and the `nodegroup.ts` type fixes are now in #2371 on their own, which also unblocks #2209. That keeps the dependency change reviewable without the regenerated SDKs in the same diff.

